PR #1, opened 2016-08-26
Added Dockerfile and modified the README

darookee:
This uses the nodejs LTS docker image as a base and installs mstream from npm.


IrosTheBeggar:
Just tested this and it works.  I'll merge it!

PR #37, opened 2017-07-09
Update cli_arguments.md

robozman:
Corrected image-cache directory instructions to use -I flag instead of -m flag


IrosTheBeggar:
Good find.  Thanks for taking the time to fix this

PR #43, opened 2017-08-29
Add Volume slider

FoxxMD:
This is pretty much just a functioning hack at the moment to address #42. Basically no styling. But I wanted to initiate this PR so I could ask some questions and clean it up:

1. What is the difference between js `/webapp2` and `/public`?

 I see that public is what is being served but most of the mstream JS files are duplicated across both folders. Why?

2. How should I approach styling?

I see png icons are used someplace while svg in other...and then `icons.png` is just a big ol' file. How 2 use?

3. How can I develop this locally on my mac more easily? 

Following the ubuntu directions fails at `npm link` but I am able to get it working correctly if I skip that and do `node mstream-command-line-boot-wrapper.js ` directly.

PR #44, opened 2017-08-30
Fix npm link issue with sqlite3

FoxxMD:
An issue with 3.1.9 (mapbox/node-sqlite3#866) causes npm link to fail. Pin `sqlite3` to 3.1.8 as a workaround.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "musicmetadata": "^2.0.3",
     "nat-upnp": "^1.1.0",
     "public-ip": "^2.0.1",
-    "sqlite3": "^3.1.4",
+    "sqlite3": "3.1.8",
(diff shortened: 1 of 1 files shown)


FoxxMD:
The same issue I was having with npm in #43 on my mac occurred on Ubuntu as well. Some investigation led me to the sqlite issue. Pinning the version fixed the problem for me.

Also, I don't want to start pushing my standards on you so I didn't add it but you should consider checking in `package-lock.json` or `npm-shrinkwrap.json` so newer versions of npm install the same version of packages. Just a suggestion :)

IrosTheBeggar:
Good find.  The Sqlite3 plugin has been nothing but a headache.

PR #66, opened 2017-10-27
Changed JSON Config link to json_config.md

Zelaf:
It was linking to the cli config.


IrosTheBeggar:
Thanks for taking the time to fix this!

PR #104, opened 2018-05-24
Add tooltips to UI buttons

ghost:
AutoDJ and RemotePlay buttons were unclear so I removed any icon ambiguity by setting tooltips.


IrosTheBeggar:
Good idea, thanks for the making a pull request instead of an issue!

PR #164, opened 2018-12-11
Added UPNP functionality...

jugendhacker:
I tried to add the upnp Server functionality...!

### DIFF
--- a/mstream.js
+++ b/mstream.js
@@ -11,6 +11,7 @@ const dbModule = require('./modules/db-management/database-master.js');
 const jukebox = require('./modules/jukebox.js');
 const sharedModule = require('./modules/shared.js');
 const defaults = require('./modules/defaults.js');
+const UPNPServer = require('upnpserver');
 
 exports.logit = function (msg) { /* Nothing. This is for electron */ }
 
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
Interesting idea. I don't use uPNP myself but it would make interfacing with some other programs easier.  Unfortunately it's not that easy to add something to this project. A few things need to be done first.

* This needs to be optional, and disabled by default

Adding big features like this need to be opt-in for a while.  Current users aren't too fond of updates that start opening other ports

* This should be separated into a module

All non-critical functions are in a module under the `/modules` folder

* Bonus Round: Add controls to enable this to the JSON config and CLI flag

The user should be able to configure stuff like the port, server name, etc.  These controls should be in the CLI flags and the JSON config

technetos:
I saw mStream meantioned in a [comment on hackernews][1] and using upnp would be a further step towards a decentralized system.  Using upnp would mean you wouldnt have to host a hole punching server for people to hole punch with.  

+ `Current users aren't too fond of updates that start opening other ports`: video game consoles do this all the time.  Although I do agree that it would be a good opt-in feature.  You could market it as not having to purchase the hole punching software :smiley:.

This is just my 2 cents, sorry for posting on a closed issue, I didnt think it was important enough to open a new issue.  

[1]: https://news.ycombinator.com/item?id=20435617

PR #190, opened 2019-06-23
Using relative url paths whenever possible

lukaswozniak:
Hello,  first off thanks for the amazing work you made with this project.
I made this change because its much easier to use nginx reverse proxy while the URL paths are not absolute. I guess there are few more possible reasons to keep the paths relative: https://www.freecodecamp.org/news/never-use-an-absolute-path-for-your-apis-again-9ee9199563be/
I could miss some paths so please review if there is anything left to change. I hope to get my commit accepted, thanks! :)

### DIFF
--- a/public/favicon/manifest.json
+++ b/public/favicon/manifest.json
@@ -9,7 +9,7 @@
    "<all_urls>"
  ],
  "background": {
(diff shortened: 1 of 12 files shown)


IrosTheBeggar:
Looks good, I just need some time to test all those changes

IrosTheBeggar:
I've been doing some testing and this works well. One font is broken and the share page is broken too.  Need to fix that before merging into master.

Thanks again @lukaswozniak

PR #196, opened 2019-07-10
Retain folder structure of uploaded directories

lukaswozniak:
This commit introdues new dependency mkdirp in order to retain folder structure of uploaded directories on server side. On browser side I used an undocumented member `file.fullPath`, which I found working on Google Chrome. Its probably a bad idea to rely on undocumented solutions, so please check if you can find a better one.

### DIFF
--- a/modules/file-explorer.js
+++ b/modules/file-explorer.js
@@ -3,6 +3,7 @@ const fs = require("fs");
 const fe = require("path");
 const archiver = require('archiver');
 const winston = require('winston');
+const mkdirp = require('mkdirp');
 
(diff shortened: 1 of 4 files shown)


IrosTheBeggar:
Good idea.  The undocumented solution should be fixed before merging, but this is a great start

IrosTheBeggar:
@lukaswozniak 

I pushed a change which fixes his problem. I did use some of your code!  Thanks for the contribution.

You clearly put some good work into this.  Would you be interested in working on some other mstream projects?  You can get in contact with me on a new discord channel: https://discord.gg/AM896Rr

PR #216, opened 2019-10-27
subsort tracks by filename

Multimodcrafter:
sort tracks by track number and then by filename. This is good if the media files do not have metadata, but the track number is at the beginning of the file name.

### DIFF
--- a/modules/db-read/database-public-loki.js
+++ b/modules/db-read/database-public-loki.js
@@ -391,7 +391,7 @@ exports.setup = function (mstream, program) {
           {'album': { '$eq': album }},
           artistClause
         ]
-      }).simplesort('track').data();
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
Thanks for the pull request.  I always appreciate some help.  Are you familiar with LokiJS?  I didn't know compound sort was a feature it supported

If you have some time, stop by the new discord channel:  https://discord.gg/AM896Rr

Multimodcrafter:
Pleasure! I'm not familiar with lokiJS. I just had a look at the code which returned the songlist, saw that it was using the simplesort method from lokiJS and looked at the docs to see what it does. There I found the Compoundsort method which did what I was looking for ;-) I'll join the channel and see if I find the time to do some more contributions. I really appreciate your work!

PR #217, opened 2019-10-28
typo and improved clarity

edward-cro:
i'm a noob and i really think clarity is key entering the "world of everything." had a few inopportune attempts at building mstream because the readme said conf.json instead of its' file name package.json... of which there are two .json files, the package and package-lock.json. Maybe this is a config.json file and i couldn't find it yet this thing built and ran anyways and I'm about to find errors. lmk if there is some logic i'm missing -- i'm pretty sure this is more one to one for the build process

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,10 @@ mStream can be configured with a JSON file that is loaded on boot. You can use t
 
 ```shell
 # Brings up an interactive shell program to edit all things in the config
-mstream --wizard /path/to/config.json
+mstream --wizard /path/to/package.json
 
 # Boot mStream with the config file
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
package.json and package-lock.json are config files used by NPM to build to project.  T

he readme is correct when using `mstream --wizard /path/to/config.json`.  This command will create a config.json file that you can use to boot mstream with `mstream -j config.json`

IrosTheBeggar:
I fixed the typo in the latest commit

edward-cro:
ah, understood. thanks Paul you're awesome

PR #220, opened 2019-10-30
Retain scroll position and search value on back button press and betw…

lukaswozniak:
…een screen changes

### DIFF
--- a/public/js/mstream.js
+++ b/public/js/mstream.js
@@ -207,6 +207,7 @@ $(document).ready(function () {
         timeout: 3500
       });
       if (programState[0].state === 'fileExplorer') {
+        saveExplorerListState();
         senddir(fileExplorerArray);
       }
     } else if (successCount === 0) {
@@ -219,6 +220,7 @@ $(document).ready(function () {
       });
 
       if (programState[0].state === 'fileExplorer') {
+        saveExplorerListState();
         senddir(fileExplorerArray);
       }
     }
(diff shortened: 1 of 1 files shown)


public/js/mstream.js
```diff
@@ -383,6 +385,60 @@ $(document).ready(function () {
...
+    if (statesList) {
+      statesList.push({
+        filelistScrollTop: $('#filelist').scrollTop(),
+        searchFoldersVal: $('#search_folders').val()
+      });
+    }
+  }
+
+  function saveExplorerListStateBeforeScreenChange() {
+    if (programState.length == 1) {
```
lukaswozniak, line 414: maybe I should add comment here: // only for screen root

public/js/mstream.js
```diff
@@ -383,6 +385,60 @@ $(document).ready(function () {
...
+        restoreExplorerListState(previousState);
+      }
+    }
+  }
+
+  function restoreExplorerListAfterScreenChange() {
+    const statesList = currentScreenExplorerListStates();
+    if (statesList.length > 0) {
+      restoreExplorerListState(statesList[0]);
+      savedExplorerListStatesPerScreen[programState[0].state] = [];
```
lukaswozniak, line 433: we restore only state for root of a screen, maybe it would be better to clear the others before screen change, not here

public/js/mstream.js
```diff
@@ -483,7 +544,7 @@ $(document).ready(function () {
           newArray.push(fileExplorerArray[i]);
         }
 
-        senddir(newArray);
+        senddir(newArray, restorePreviousExplorerListState);
```
lukaswozniak, line 547: the passed function is executed when api call completes, that's why I cant just call it right here

public/js/mstream.js
```diff
@@ -383,6 +385,60 @@ $(document).ready(function () {
...
+    fileExplorer: [],
+    allPlaylists: [],
+    allAlbums: [],
+    allArtists: [],
+    allRated: [],
+    recentlyAdded: []
+  }
+
+  function currentScreenExplorerListStates() {
+    if (programState.length > 0) {
```
lukaswozniak, line 398: always true, remove?

public/js/mstream.js
```diff
@@ -383,6 +385,60 @@ $(document).ready(function () {
...
+      const previousState = statesList.pop();
+      if (previousState) {
+        restoreExplorerListState(previousState);
+      }
+    }
+  }
+
+  function restoreExplorerListAfterScreenChange() {
+    const statesList = currentScreenExplorerListStates();
+    if (statesList.length > 0) {
```
lukaswozniak, line 431: should be always true in this context, remove?

Issue #213, opened 2019-10-23
Transcoding Bugs


IrosTheBeggar:
An early versions of transcoding has been added.  It works well enough in most cases, but it's not bug free yet.  This ticket is to track all issues related to transcoding:

- Doesn't work in Safari
- Duration and Seek don't work in webapp

aroundmyroom:
will this feature, if enabled detect if you are on mobile yes/no ?
I see this feature in the menu now and its disabled, but it is not in the config to be enable/disable yet correct?

IrosTheBeggar:
There is no mobile detection.  You manually enable it.

You can enable it wit hthe `-t` flag or in the config with 
```
  "transcode": {
    "enabled": true,
    "defaultCodec": "opus",
    "defaultBitrate": "128k"
  },
```

laduke:
Hi! I was messing with something similar; I hadn't found this project yet. 

Here's what I know about making it work in safari:
- need 206 status code
- need content range header
- need to be able to send an arbitrary range (for example: 0-1 or 16384-32767)

unfortunately, the content range header needs the total size, so you have to guess (bitrate * track length). 
The guess will be a little off, you you have to pad some 0's onto the end of the stream. 
So, the browser will make 3 or 4 requests for the same song, and you'll have to transcode the file that many times, or figure out some caching of the stream. Have fun!

romjeans:
Hi,
Do you know a way to permanently enable transcoding ? After enabling transcoding, when I close the browser and open again mstream web page, the transcoding option becomes "disabled".
Thanks
Rom

jimthedj65:
I have checked this app in 

> Safari Version 14.1.1 (16611.2.7.1.4) Big sur 11.4 on MacBook Pro (15-inch, 2017)

> Google Chrome Mac Version 91.0.4472.114 (Official Build) (x86_64), 

> Vivaldi | 4.0.2312.33 (Stable channel) (x86_64), Revision | 75595d1359deaec4b68624c45e95be53e1e6ad13
OS | macOS Version 11.4 (Build 20F71), JavaScript | V8 9.1.269.36, User Agent | Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36

I enabled transcoding and all works great, Vivaldi plays the songs but doesn't show the seek and time for the track.

[...]

PR #222, opened 2019-11-02
Support range request combined with ffmpeg

Borewit:
In order to be able to seek in the media stream, I believe a proper handling [HTTP range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) is required,

This PR is a start of that work towards the range requests.

### Testing
To be able to test the HTTP-range-requests in a controlled environment I have put a simple client together which is using the HTTP-range-requests to download audio information: [mstream-http-range-client](https://github.com/Borewit/mstream-http-range-client)

Related issue: #213 _Transcoding bugs_

### DIFF
--- a/modules/ffmpeg.js
+++ b/modules/ffmpeg.js
@@ -4,10 +4,19 @@ const ffmpeg = require("fluent-ffmpeg");
 const winston = require('winston');
 
 const codecMap = {
-  'mp3': 'libmp3lame',
-  'opus': 'libopus',
-  'aac': 'aac'
+  'mp3': {codec: 'libmp3lame', contentType: 'audio/mpeg'},
+  'opus': {codec: 'libopus', contentType: 'audio/ogg'},
+  'aac': {codec: 'aac', contentType: 'audio/aac'}
 };
+
+function initHeaders(res, audioTypeId, audioPath) {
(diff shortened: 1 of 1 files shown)


Borewit:
Using the transcode flag:
```js
const transcode = false;
```
one can switch between transcoded and straight link, in [client.js](https://github.com/Borewit/mstream-http-range-client/blob/master/client.js).

The client currently is able to extract metadata-data without transcoding. This reflects the differences as described in issue #213. 

I suggest to look for an existing library supporting the range requests, such as:
* [ ] [express-range](https://www.npmjs.com/package/express-range)
* [ ] Investigate how we can read chunks from ffmpeg. Maybe it is required to cache it's output.

IrosTheBeggar:
Github says there are merge conflicts. I also noticed this included the same code style commit the other PR had.

Borewit:
Conflicts addressed. 

Since I force pushed, please delete old local branch, and pull a fresh one from GitHub.

Don't merge this one into the master as it is not suitable for production.

Borewit:
It was just a beginning...

IrosTheBeggar:
@Borewit for experimental features I like to put commits out early and often.  Some of the more active users find and report bugs and give feedback

I made sure to do a cross platform and cross-browser test first

PR #244, opened 2020-01-22
Fix uploading to folders with non-Ascii characters in name

lukaswozniak:


IrosTheBeggar:
Fixed in latest commit

PR #255, opened 2020-05-05
Support deploying in subdirectory (#254)

### LINKED ISSUE #254: Support mStream in subdirectory of server / reverse proxy
Currently, mStream only supports running at the root of the server, like `http://localhost:3000/`.

In order to make mStream deployable on a domain without affecting the rest of the site, mStream should be deployable at `http://localhost:3000/mStream`.

This pattern is supported by various other server software,
like BookStack https://www.bookstackapp.com/docs/admin/subdirectory-setup/
and Gitea https://docs.gitea.io/en-us/reverse-proxies/

d10n:
Fixes #254 

Set rootPath to mount mStream in an http subdirectory.
For example:
    rootPath: "/mstream"
makes mStream accessible at `http://localhost:3000/mstream/`


d10n:
It looks like I independently did some of the same work from #190, which is not merged into master, but I also fixed the font issue

d10n:
This is now mergeable. I only tested the web server, not electron, but everything I tested seemed to work.
HTML pages in /public/ are moved to /template/. This lets express inject the root path of the mstream application into the HTML and JS.

public/js/api2.js
```diff
@@ -1,3 +1,15 @@
+var ROOT_PATH = (function() {
+  if (window.ROOT_PATH == null) {
+    // Guess root path
+    var selfPath = new URL(document.currentScript.src).pathname;
```
d10n, line 4: document.currentScript is not supported in IE, but this is just backup code anyway; ROOT_PATH is overridden in <script> tags in the templated html

public/js/mstream.js
```diff
@@ -156,7 +156,7 @@ $(document).ready(function () {
   const myDropzone = new Dropzone(document.body, {
     previewsContainer: false,
     clickable: false,
-    url: '/upload',
+    url: '/upload', // TODO is ROOT_PATH needed?
```
d10n: Forgot about this TODO; can someone else test if this works

public/js/api2.js
d10n: I see other code in mStream that does not support IE so I guess this is not important

IrosTheBeggar:
Taking a look at this now

IrosTheBeggar:
@d10n you have some time to talk about this.  It's a a big update and I have some questions

d10n:
@IrosTheBeggar yes. I tried to join the discord channel mentioned in the readme, but I get shown this message. The link might need to be an invite link instead of a channel link.
image

buergi:
Have the open questions been sorted out? I really would love to see this merged. I'm using d10n's branch since some weeks now without any problems.

buergi:
May I kindly ask if there are any news here. I'm still trapped in d10n's branch as I require the subdirectory feature for my purpose. I'd really love to see this feature in mStream.

IrosTheBeggar:
@d10n @buergi 

I'm working on a solution to this problem that doesn't require server side rendering.  It's going to take some time since I'm working on fixing several long standing pieces of technical debt right now as well.

Btw, there's nothing wrong with this branch and using server side rendering, I'm just trying to avoid it for mStream because I want to keep the webapp decoupled from the server.  Decoupling will allow the webapp to be compiled as a desktop app via electron in the future.

If you want to help test, let me know in the Discord: https://discord.gg/AM896Rr

IrosTheBeggar:
Relative URLs are being worked on and are supported on the main page of mStream now.  Thanks for working on this and providing a path others could use

4volodin:
> Fixes #254
> 
> Set rootPath to mount mStream in an http subdirectory. For example: rootPath: "/mstream" makes mStream accessible at `http://localhost:3000/mstream/`

where should I set the "rootPath":"/mstream" ?

PR #256, opened 2020-05-07
Fix volume increase on track change or restart

d10n:
When clicking on the currently playing item in "Now Playing" a few
times:

In Chrome, an exception is thrown:

Failed to execute 'createMediaElementSource' on 'AudioContext': HTMLMediaElement already connected previously to a different MediaElementSourceNode.

but Firefox is happy to try connecting the same source multiple times,
leading to apparent volume increases despite gain and volume settings
remaining constant.

### DIFF
--- a/public/js/mstream.player.js
+++ b/public/js/mstream.player.js
@@ -441,12 +441,16 @@ var MSTREAMPLAYER = (function () {
     // connect to visualizer
     if (VIZ) {
       var audioCtx =  VIZ.get();
-      var analyser = audioCtx.createAnalyser();
       try {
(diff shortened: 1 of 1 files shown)


jerrycan619 (approved):
Works for me without any problems!
Now the Tracks don't get louder and louder by time (or restart)

Metasynic:
Is this getting merged soon? Having the same problem.

IrosTheBeggar:
Thanks for making a pull request

Sorry for taking so long to get around to this.  2020 has been a real pain of a year.

PR #260, opened 2020-05-23
Client-side logout

jerrycan619:
Added a Logout Button on the top bar to be able to perform a logout and switch users.

Since jwt doesn't support token delete or expire on demand there is no way to perform a logout on the server-side. I could probably be done using jwt-redis package instead of jwt but it would require a lot more changes.

So the simplest method is to destroy the local token saved on the browser.
This is a potential vulnerable because the token still exists on the server side and could be used by someone else. Never the less i think that's not gonna happen that easy.

Tested: Firefox,Chrome,Opera,Android Chrome


IrosTheBeggar:
@jerrycan619 

I'm getting back to work on mStream. This is a simple feature I'd like to get integrated.  Can you clean up the commits so there's no undos or rollbacks?  

If you have some time, hop in the Discord to discuss this further: https://discord.gg/AM896Rr

IrosTheBeggar:
The latest version has a logout button

Issue #262, opened 2020-06-02
3rd world problem: shuffle play should be working on first song not after first song


aroundmyroom:
If we add songs from a folder into a playlist and press shuffle and than play, the first song is always the first song from the list and not a random song from that playlist.
I would love to see that behaviour changed.

Now I have to remind myself to press next song to get a random other song.

jerrycan619:
So the trick would be if shuffle is on and play is pressed to just do a "track skip next" internally?
I don't know about other players.. is this a common behavior?
or maybe add another button for shuffle on first track?

1.
The first track on the playlist is played immediately after its added.
So after adding tracks to the playlist, pause the first one, enable shuffle and then it should not continue the same track and instead shuffle to another?
This is quite simple and just require some small code changes which I just did.

2.
The bigger problem is if you enable shuffle before anything is in the playlist, and then add a whole folder to the playlist... to shuffle without playing the first track
I don't know at the moment how to do this and have no time to figure it out. 
I maybe come back later to this.

So.. if you are interested in the 1. solution and have installed mStream from the CLI (via npm install)
I cloud send you the changed file and you could try and test it yourself.
And if it works like expected I'll do a pull request.

aroundmyroom:
Indeed in the settings there must be this:
enable auto play when loading playlist yes / no
when no you can say: when shuffle is pushed the first random track is being played

Now it is indeed loading first item, what is automatically played so random track will be always first track after initial track

When you insert a CD all tracks are normally loaded but being played only when pressing play
pressing play or pressing shuffle is doing the same: playing a song, where play is starting the first song or the selected song being loaded and shuffle is doing a random track of the loaded playlist

Option one is a good alternative

I am able to change the code on the place it is needed ;)

jerrycan619:
Ok.. so I spend some more minutes thinking about it and i've found a way to realize option 2.
So now if you turn shuffle on BEFORE adding a track list it will directly random play without playing the first one (except it is chosen by the shuffle :) ). I just tested it 5min and it seems to work.. Please test it further since I don't use shuffle that often (Auto DJ ftw).

The downside or Bug (let's call it feature) is, that if you turn shuffle off and on while a playlist is played it's shuffles two times on the next track change, but it's not noticeable.

For option 2. edit /public/js/mstream.player.js
Search for "mstreamModule.toggleShuffle" and edit:
```js
if (mstreamModule.playerStats.shuffle === true) {
      newShuffle();
    } else {
[...]

PR #263, opened 2020-06-02
Modify shuffle behaviour

aroundmyroom:
Changing behaviour of shuffle play: initially mStream is always playing the first track and than shuffles, Now with these changes it is loading first track, but before playing shuffles to another random track


jerrycan619:
Just for documentation:

Result of #262 
Further changes needed to be completely working in all cases

PR #265, opened 2020-06-09
Rating 0 Stars / Remove Rating

jerrycan619:
If you don't like a song anymore or rated by mistake there is no way to remove that rating.

With this simple code change it's now possible to remove the rating by selecting 0 Stars on the song
(Click slightly outside next to the first star, so that 0 stars are filled/selected).
It also disappears from the Rated list in the Explorer.

Without this change an error appears by selecting 0 stars.

I think the other value validation in this if-statement is enough to keep it safe.


IrosTheBeggar:
The latest version cleans up the rating API and adds this feature

PR #267, opened 2020-06-11
MediaSession API Support

jerrycan619:
Adds the ability to control mStream with media keys on the keyboard and on Chrome you have a nice little icon next to the address bar which shows on click the current playing track with cover, all its metadata and controls. Also a panel appears on smartphone lockscreen and smartwatches(not tested).

See: https://developer.mozilla.org/en-US/docs/Web/API/Media_Session_API
Example: https://googlechrome.github.io/samples/media-session/

Code information:
Seek commented out because I think its unnecessary in this place.
Seek is working fine on desktop chrome while I can't get it to work on my android lockscreen.

The check for `mstreamModule.playerStats.metadata.title === "undefined"` results of a behavior which I can't explain. 

If you add a first Track to a empty playlist
`console.log(mstreamModule.playerStats.metadata)` 
outputs it's metadata like expected with title and so on..

but `console.log(mstreamModule.playerStats.metadata.title)` outputs "undefined".. 
but it's available because you can see it in the first console.log..!?!?!?!?.

So it just calls itself again until the title is not undefined anymore (on not tagged files the title gets "null")
Try it yourself by uncommenting the two console.log

The place for  updateMediaSession() is the best I could find because if you place it  for example in "onplay" there could be a time gap if the track is loading and the media session disappears for like 1s and then comes back. 
To minimize / close this gap that's the best place I've figured out after a lot of testing.

Tested on: 
Android Chrome
Chrome (best choice because of the Icon next to address bar)
Opera (no Icon -> no controls, but media keys on keyboard work)
Firefox (need to be enabled in settings however it doesn't work for me but documentation says it is supported)


IrosTheBeggar:
I added media sessions support in the last commit.  Thanks for the advice

Issue #276, opened 2020-08-11
Log failed login attempts and bind to localhost only


AlphaJack:
Is it possible to log failed login attempts to mstream.log?
It may also be useful to update this file whenever the database gets updated with new tracks, by default every 250 new files.

---

I noticed that by default mStream binds to 0.0.0.0 and there is no option to restrict the host to 127.0.0.1, I had to modify `mstream.js`
from
```javascript
server.listen(program.port, () => {
```
to
```javascript
server.listen(program.port, '127.0.0.1', () => {
```
Is it possible to add an option in `mstream --wizard`?

IrosTheBeggar:
I merged a pull request from @jmacmahon that adds this functionality. I also updated the default address to 127.0.0.1

I didn't realize the default behavior of express was to bind to 0.0.0.0.  Thanks for pointing this out.

I'll look into better logging as well

PR #279, opened 2020-08-31
Allow user-configurable bind address

jmacmahon:
This affects issue IrosTheBeggar/mStream#276

Also default `folders` attribute in config to prevent an error on
startup when it is not set.

### DIFF
--- a/modules/config/config-inquirer.js
+++ b/modules/config/config-inquirer.js
@@ -6,6 +6,7 @@ const path = require('path');
 const Login = require('../login');
 const br = require('os').EOL;
 const defaults = require('../defaults').setup({});
+const Joi = require('@hapi/joi');
 
 exports.addKey = function(current, filepath, callback) {
   if (!filepath) {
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
I checked this out and everything looks good.  I'm going to update the default address from `0.0.0.0` to `127.0.0.1`.  See more here: https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0

aroundmyroom:
Actually, since you have modified it to 127.0.0.1 it is not working anymore to be allowed to access the HTML page from another machine than localhost.

today I updated and I noticed that I had to stash some changes I made to be able to update
than I noticed that the default from default.js was back.. this above change.

So I have to bind the address back to 0.0.0.0 to access it through my network 

this behaviour is seen many times with other applications as well

127.0.0.1 binding causes the application only to be working on the machine itself not from other machines in the network, but as mStream is running on a own VM through my server I never access it through localhost.

0.0.0.0 allows the webserver to be accessible through any other machine in the network.

I have modified default.js line 25 to 0.0.0.0 again and I can access it again.

the real normal behaviour would be:
in the config use 127.0.0.01 but allow it to become 0.0.0.0

PR #285, opened 2020-10-30
Add ReplayGain support.

dlbeswick:
I've added ReplayGain (track-level only) support in this pull request.

There's a new "RG" button next to the volume mute control. It can be clicked to activate ReplayGain, and further clicks cycle through a list of predefined pre-gain values before disabling ReplayGain.

The user's ReplayGain activation and pre-gain settings are saved.

The metadata db will need to be rebuilt so that the ReplayGain metadata for each track is made available.

Thanks for this project!


IrosTheBeggar:
This looks pretty great. Will get it merged soon

PR #336, opened 2021-10-19
added init script 

AlexanderGottschlich:
I've added a init script and updated README with instructions.


d10n:
@AlphaJack has a systemd unit file too at <https://aur.archlinux.org/cgit/aur.git/tree/?h=mstream-git>, but it has a different root path

PR #345, opened 2021-11-12
fix admin add folder error

tannineo:
Fix the error when adding a folder.

The form data is wrapped as `value` inside the return of `joiValidate()`.

image


IrosTheBeggar:
Thanks for this fix!

PR #346, opened 2021-11-13
Fix another add-directory error

ericpaulbishop:
There's another spot where input.vpath needs to be replaced by input.value.vpath in the code for adding a directory, otherwise it doesn't get scanned properly.


IrosTheBeggar:
Good catch, thanks for making a PR

PR #360, opened 2022-03-14
create configs.transcode.ffmpegDirectory on enable if it does not exist

### LINKED ISSUE #359: BUG: ffmpeg not extracted in docker
1. Click "enable transcoding" in admin panel
2. get notification ffmpeg was downloaded
3. attempt to use ffmpeg binary
4. get EACCESS error in logs from trying to execute ffmpeg binary
5. `docker exec -it $MSTREAM_CONTAINER_ID bash`
6. `# find /opt -type f -name ffmpeg`
7. No results.  Directory is empty.
8. Check /config/.ffbinaries_cache/, see ffmpeg and ffprobe zips

For whatever reason mstream is not unzipping the binaries when run under docker.  Manually unzipping the binaries worked, so the container does have unzip(1) installed.

EDIT: this appears to be a problem with `enableTranscode()` from `src/util/admin.js` not setting the default directory.  Pull request incoming.

tidux:
Running mStream from Docker without a pre-existing config.json file, enabling transcoding through the UI doesn't work.  The binaries are downloaded correctly, but without an output directory provided to `ffbinaries.downloadFiles()` the extract-and-chmod+x step fails silently.  Resolves https://github.com/IrosTheBeggar/mStream/issues/359.

### DIFF
--- a/src/util/admin.js
+++ b/src/util/admin.js
@@ -307,6 +307,10 @@ exports.enableTranscode = async (val) => {
   const loadConfig = await this.loadFile(config.configFile);
   if (!loadConfig.transcode) { loadConfig.transcode = {}; }
   loadConfig.transcode.enabled = val;
+  if (val == true && !loadConfig.transcode.ffmpegDirectory) {
+      // default value hard coded until UI enables changing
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
This fix has been added by the linux server group

PR #365, opened 2022-05-02
add albums on addAll()

fabi321:


ToxicFrog:
It looks like this makes it so you can click the little `[+]` in the top right of the browser and have it recursively add everything, rather than just files? I'm going to try this out on my server and see how it goes.

fabi321:
Yes, it recursively iterates all albums if you press the plus button, just like folders are already iterated if you press the plus in the file view

ToxicFrog:
File view doesn't do that, though. Clicking the `+` next to an individual folder will recurse _in that folder_ but clicking the larger `[+]` in the top right does nothing. And upon testing, this PR doesn't change that - so I'm guessing it only affects album view and not file view?

PR #368, opened 2022-05-02
more data escaping

fabi321:
If this is merged, #365 won't work anymore


IrosTheBeggar:
Thanks for contributing. Sorry it took so long to get back. I wrote another PR based on this which fixes the escaping throughout the app, so thanks for pointing this out

PR #371, opened 2022-05-13
added hotkeys

DySprozin:
added hotkeys: ctrl left/right (prev/next song) and just left/right for 5 seconds back and forward


IrosTheBeggar:
I have added a hotkeys panel to the webapp. Thanks for making a PR

PR #398, opened 2022-11-30
Escape question marks in filenames

### LINKED ISSUE #397: Playback fails on filenames containing question marks
Whenever I attempt to play a song that has a question mark in the title and filename, mStream pops up "Failed to play song".

A quick peek at the F12 console reveals that mStream does not encode question marks when making API requests:

scrot

This isn't just annoying, it could even pose a security risk - I haven't worked out a viable exploit, but it seems pretty obvious that anyone in control of filenames on the server can plant files whose filenames escape the URL path and leak query parameters into the URL.

tdammers:
This should fix #397.

Factoring out the URL-encoding, and checking it for other potentially problematic characters, might be worth doing though.


IrosTheBeggar:
Thanks for taking the time to fix this

PR #543, opened 2026-04-03
feat: replace ffbinaries with self-contained ffmpeg auto-download (BtbN static builds)

aroundmyroom:
## Problem

The `ffbinaries` package used by the transcode feature has two serious issues:

1. **Serves ffmpeg 4.x** — the last version indexed by ffbinaries. This is from 2022 and lacks modern codec improvements. ffmpeg 6+ is the current stable baseline.
2. **Broken on arm64 / Alpine / Docker** — `ffbinaries` has unresolved issues detecting the platform correctly on arm64 Linux and Alpine-based containers, causing transcoding setup to silently fail.

## Solution

New module `src/util/ffmpeg-bootstrap.js` that handles everything:

- Downloads static ffmpeg + ffprobe from **[BtbN/FFmpeg-Builds](https://github.com/BtbN/FFmpeg-Builds)** — the de-facto standard source for GPL static builds, actively maintained
- **Linux x64** and **Linux arm64** are both supported
- **macOS / Windows**: logs a warning and skips auto-download; user must supply binaries manually (same as today)
- Checks the installed binary's major version at startup — if below v6, re-downloads automatically (so existing old ffbinaries installs self-upgrade)
- Single cached promise prevents concurrent downloads
- Extracts using system `tar` with explicit paths (no `--wildcards`) — works with both GNU tar and BusyBox tar (Alpine/Docker)
- **No new npm dependencies** — only Node.js built-ins (`https`, `fs`, `child_process`)

`transcode.js` is updated to use `ffmpegBin()`/`ffprobeBin()` from the new module. The `ffbinaries` import is removed.

## Testing

Verified on Linux x64 and Linux arm64 (Docker). ffmpeg 7.x is downloaded and transcoding works correctly.

### DIFF
--- a/src/api/transcode.js
+++ b/src/api/transcode.js
@@ -1,12 +1,10 @@
 import path from 'path';
-import ffbinaries from 'ffbinaries';
 import ffmpeg from 'fluent-ffmpeg';
 import winston from 'winston';
 import * as vpath from '../util/vpath.js';
 import * as config from '../state/config.js';
 import { Readable } from 'stream';
-
-const platform = ffbinaries.detectPlatform();
+import { ensureFfmpeg, ffmpegBin, ffprobeBin } from '../util/ffmpeg-bootstrap.js';
 
 const codecMap = {
   'mp3': { codec: 'libmp3lame', contentType: 'audio/mpeg' },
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
Nice find. I didn't realize how out-of-date that library was.  I looked through your code and made some imrprovements on it.  I added checksum verification for security, some automated testing, and the ability to download binaries for windows and linux: You can see it in the new branch https://github.com/IrosTheBeggar/mStream/pull/544/changes/80212fec1c983d1abc2b6dca8564dbcbc0f5aecc

PR #568, opened 2026-04-24
add config option to set the expressjs "trust proxy" setting when running behind a reverse proxy

rigel314:
When running behind a reverse proxy, the logs always show the reverse proxy's IP when logins fail.  The logs should show the original client IP by looking at the `X-Forwarded-For` header set by the proxy.  This is something that expressjs can already do with a small application code change [More details](https://expressjs.com/en/guide/behind-proxies.html).

I intend to use this with fail2ban scanning the logs for failed logins for blocking IPs, but the logs need to show the real client IP, not my proxy's IP.

The new option should be left false when _**not**_ running behind a trusted reverse proxy to prevent a trivial client IP spoof.

### DIFF
--- a/docs/json_config.md
+++ b/docs/json_config.md
@@ -14,7 +14,7 @@ All config params have default values. A newly generated config will be empty js
 { }
 ```
 
-A heavily edited config would look like: 
+A heavily edited config would look like:
 
 ```json
 {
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
Looks good, thanks for contributing!

PR #595, opened 2026-05-16
feat(schema): V36 — track ytdl provenance via tracks.source + embedded file tag

IrosTheBeggar:
## Summary

Adds `tracks.source TEXT` (V36) — an open-enum provenance column —
backed by an embedded file tag (`MSTREAM_SOURCE`) so the signal survives
scanner rescans and follows the file across copies/moves. Today only
the ytdl handler populates it ('ytdl'); future inserters (upload API,
plugin importers) can add their own labels without a migration.

Bundled bugfix: the ytdl INSERT statements still referenced the
`tracks.genre` column that V34 dropped, so the handler errored at
runtime on master post-V34. Fixing it without also adding provenance
would leave the original problem (no DB-level signal that a track came
from ytdl) untouched — bundling keeps "make ytdl work post-V34" and
"track ytdl provenance" in one atomic change.

## How it works

1. **ytdl handler** (`src/api/ytdl.js`) — writes `-metadata MSTREAM_SOURCE=ytdl`
   via ffmpeg during the post-download tag pass (alongside any
   user-supplied title/artist/album). ffmpeg encodes the key naturally
   per container:
   - **MP3 / WAV (ID3v2)** → TXXX frame, description=`MSTREAM_SOURCE`
   - **FLAC / OGG / Opus (Vorbis comments)** → `MSTREAM_SOURCE=ytdl`
   - **M4A / M4B (MP4 atoms)** → `----:com.apple.iTunes:MSTREAM_SOURCE`
2. **DB insert** — both ytdl handler INSERTs now pass `source = 'ytdl'`
   directly (and drop the dead `genre` column; and stop overloading
   `scan_id` with the literal `'ytdl'`, which was never effective
   provenance because every subsequent scan overwrites it).
3. **Scanners** — both `src/db/scanner.mjs` (JS fallback) and
   `rust-parser/src/main.rs` (default) extract `tracks.source` from the
   file's native tag namespace. Priority order:
   1. Explicit `MSTREAM_SOURCE` tag — used directly.
   2. yt-dlp's `purl` field (embedded by `--embed-metadata`) containing
      youtube.com / youtu.be — infer `'ytdl'`. Catches files downloaded
      via plain `yt-dlp` outside mStream.
   3. NULL — pre-existing rows, manual imports, providers we don't
      recognise.

Round-trip safety: when the scanner takes the re-extract branch (mtime
drift) it re-reads the marker tag, so `INSERT OR REPLACE` preserves
provenance instead of clobbering it.

## Other small fixes bundled in

- `SCHEMA_VERSION` was 34 on master but `MIGRATIONS` already included
  V35; bumped to 36 so the runner stops early-returning on a V35+ DB.
- `test/db-v34-drop-genre.test.mjs`'s "column drop is the only schema
  delta V34 introduces" test was unbounded on its post-V34 snapshot and
  would otherwise pick up V36's `source` column as a spurious delta;
  bounded at `upToVersion: 34`.

## Rollback

`SCHEMA_V36_DOWN` ships behind `scripts/rollback-v36.js` (same pattern
as V31 — operator runbook in `docs/migration-rollback.md` applies).
Boomerang caveat: a V36-aware build re-applies V36 on next boot; pair
the script with a code revert.

## Files touched

- `src/db/schema.js` — V36 migration + V36_DOWN + SCHEMA_VERSION bump.
- `src/api/ytdl.js` — write MSTREAM_SOURCE tag, fix V34 genre-column
  bug in both INSERTs, set `source='ytdl'`, stop writing `scan_id='ytdl'`.
- `src/db/source-detect.js` — `detectSource(parsed)` helper, extracted
  to its own module so tests can import without triggering scanner.mjs's
  CLI-arg parser.
- `src/db/scanner.mjs` — call `detectSource()` in `parseMyFile`, bind
  `source` in the prepared INSERT.
- `rust-parser/src/main.rs` — same shape on the Rust side via
  `detect_source_from_tag()` (lofty's `tag.items()` iteration with the
  same priority order).
- `scripts/rollback-v36.js` — standalone rollback CLI, mirrors
  `rollback-v31.js`.
- `test/db-v36-tracks-source.test.mjs` — schema-shape, INSERT
  round-trip, V36_DOWN tests (12 cases).
- `test/ytdl-provenance.test.mjs` — end-to-end: ffmpeg writes the tag,
  scanner reads it back, mtime-drift round-trip, NULL case, purl
  fallback. Skips when bundled ffmpeg is missing (matches the existing
  `scanner-parity` pattern).
- `test/db-v34-drop-genre.test.mjs` — bound the post-V34 snapshot at
  `upToVersion: 34`.
- `package.json` — add new test files to `npm test`.

## Test plan

- [x] V36 migration tests (12 cases) — schema shape, INSERT round-trip,
      rollback. All passing on `:memory:` DBs.
- [x] V36 `detectSource()` unit tests — 7 cases covering MSTREAM_SOURCE
      precedence, purl fallback, youtu.be short form, non-YouTube purl,
      no-marker NULL, ID3v2 TXXX (prefix + object forms), MP4 freeform.
- [x] Existing V34 test passes again after bounding fix.
- [x] Rust binary builds cleanly.
- [ ] Full suite on a machine with bundled ffmpeg available — local
      env doesn't have `bin/ffmpeg/ffmpeg.exe`, so 408 ffmpeg-dependent
      tests cancel-cascade. All those cancellations trace back to
      `Error: bundled ffmpeg missing at .../bin/ffmpeg/ffmpeg.exe` —
      not introduced by this PR. Without ffmpeg, the totals are
      `262 pass / 0 fail / 408 cancelled / 34 skipped`. The
      ffmpeg-driven `test/ytdl-provenance.test.mjs` scanner-readback
      and round-trip cases need CI / a local box with the bundled
      binary to actually exercise.


IrosTheBeggar:
## Smoke test results (post-merge of 9e5e8301)

Ran a 35-case end-to-end smoke test against the actual scanners + DB.
Fixture matrix: FLAC / MP3 / Opus / M4A with `MSTREAM_SOURCE`, plus
purl-only fallback (youtube + non-youtube), no-marker control, and the
conflict case (`MSTREAM_SOURCE` vs. non-youtube `purl`).

```
PHASE 1 — JS scanner                              8/8 OK
PHASE 2 — Rust scanner                            8/8 OK   (parity with JS)
PHASE 3 — ytdl handler tag-write path             3/3 OK
PHASE 4 — round-trip (mtime drift)                4/4 OK   (JS + Rust)
PHASE 5 — V34 ytdl genre-column fix INSERT        3/3 OK
PHASE 6 — rollback + re-migrate                   9/9 OK
SUMMARY: 35 passed, 0 failed, 0 skipped
```

### What the smoke test caught

**M4A is the documented "fragile container" the original task brief warned about.** ffmpeg's MP4 muxer silently drops non-standard `-metadata` keys on write — `MSTREAM_SOURCE`, `purl`, and `----:com.apple.iTunes:KEY` all vanish (verified with ffprobe). yt-dlp itself faces the same limitation.

Honest scope of the limitation:
- **Scanners READ freeform iTunes atoms fine** (lofty + music-metadata both handle them). Files tagged externally via mutagen / AtomicParsley still get attributed correctly.
- **Handler's INSERT still sets `source='ytdl'`** for M4A downloads.
- **Scanner's mtime fast-path preserves the column** across normal rescans.
- **Only break case: re-extract after mtime drift on an M4A** — the scanner can't re-derive 'ytdl' from the file because the tag never landed.

Follow-up commit (9e5e8301) updates `src/api/ytdl.js`, `src/db/source-detect.js`, and `test/ytdl-provenance.test.mjs` to document this honestly. The end-to-end M4A test now asserts `source = null` (the real outcome); the unit test for the freeform-atom branch stays in place because lofty/music-metadata reading is correct — the limitation is purely on the ffmpeg-write side.

### What worked beyond the unit-test coverage

- Both scanners parity: identical `source` values across all 8 fixtures, including the MSTREAM_SOURCE-vs-purl conflict case (priority-1 wins on both).
- Rust scanner's lofty-based readback handles all three working containers (FLAC Vorbis, MP3 TXXX, Opus Vorbis) — the gap I'd flagged as a possible Rust-side risk didn't materialise.
- Round-trip: after mtime bump → forced re-extract → `source` re-populated from the tag on both scanners.
- V34 genre fix verified: post-fix 16-column INSERT runs cleanly, pre-fix 16-column INSERT with `genre` errors out as expected (negative control).
- Rollback round-trip: `source` column dropped → `user_version=35` → forward re-migration restores column at 36 with track rows intact.

### Shipping verdict

PR is **shippable as-is** with the M4A limitation documented. M4A users get DB-level attribution at download time; only the mtime-drift recovery is lost. Future improvement: write the M4A marker via mutagen or AtomicParsley in a follow-up if anyone reports the limitation hitting them.

PR #603, opened 2026-05-29
fix(scanner): repair and re-align the JS fallback scanner

### LINKED ISSUE #3: UI obstructed by playback bar
The file explorer extends behind the playback bar making parts hidden.  Using Firefox 45.3.0 

mstream-screenshot

IrosTheBeggar:
## Summary

Auditing the scanner surfaced a cluster of bugs in the **JS fallback scanner** (`src/db/scanner.mjs`). It had quietly diverged from the Rust scanner because the parity suite only ever exercises the Rust binary. Fixes, highest-severity first:

1. **Config rejection — the fallback was dead on arrival.** `task-queue.js` sends `waveformCacheDir` in every `jsonLoad` and no longer sends `generateWaveforms`, but the JS Joi schema listed only the latter and rejected unknown keys — so the fallback exited with `Invalid JSON Input` before doing any work. Now accepts `waveformCacheDir` and tolerates unknown keys (`.unknown(true)`) so future Rust-only config fields can't silently break it again.

2. **Data loss on a vanished mount.** With no accessibility guard, a transient CIFS/NFS outage made the walk find nothing and then `DELETE FROM tracks WHERE scan_id != ?` wipe **every** track for the library (cascading to albums / artists / `user_album_stars`). Added the upfront + post-walk guards the Rust scanner already has in `run_scan`.

3. **Album-art hash corruption + JS/Rust parity.** Embedded and directory art were hashed via `buffer.toString('utf-8')` — a lossy UTF-8 round-trip, so the digest was neither the real MD5 of the image nor what Rust produces (it hashes raw bytes). Distinct covers could collide onto one filename. Now hashes the raw bytes and maps the embedded MIME type to an extension the way Rust does (`image/jpeg` → `.jpeg`, not mime-types' `.jpg`).

4. **Sidecar stat storm.** The fast-path issued ~22 `statSync` calls per file on every scan (even unchanged files and sidecar-free libraries). Added a per-directory listing cache (`sidecarMtimeCached`) mirroring the Rust scanner's `DirListing` — one `readdir` per directory.

5. **Double tree walk + symlink-cycle hang.** Replaced the separate count-then-scan passes with a single `collectFiles` walk, and added realpath-based cycle detection when `followSymlinks` is enabled (walkdir gives the Rust scanner this for free).

Also fixes a stale comment line-reference in `rust-parser/src/main.rs` (comment-only).

## Rebased onto master + ported the #612 Rust scanner optimizations

This branch is rebased onto current master, which now includes #612's scanner write-path overhaul. Most of #612 reaches the JS fallback automatically through the rebase — `synchronous=NORMAL` / `cache_size` / `temp_store` PRAGMAs and the `NOT IN`→`NOT EXISTS` orphan cleanup. Two wins needed a **JS-side port**, added in a separate commit (`perf(scanner): port Rust write-path optimizations to the JS fallback`):

6. **Keep the write lock free during extraction.** The scan loop wrapped the entire walk — including `parseMyFile` (read + tag parse + hash + album-art I/O) — in one batched transaction, so a concurrent API write (playlist save, scrobble) could block for the length of a decode and blow past `busy_timeout`. Now the cheap unchanged-file `scan_id` bumps batch under one transaction, but `processFile` **flushes that batch — releasing the write lock — before parsing each changed file**, parses with no transaction open, then writes that one track in its own tight `BEGIN…COMMIT`. A mid-write failure rolls back just that track (the JS analogue of the Rust writer's per-song savepoint) instead of leaving a partial row for the next batch flush. Mirrors the Rust serial-path / greedy-drain restructure.

7. **Guard the per-album `updateAlbumTags` UPDATE** so it's a no-op (0 rows, no row rewrite) when `album_artist`/`compilation` are unchanged — otherwise every track of a shared album rewrites the album row identically. Same guard ported from the Rust `find_or_create_album`.

## Tests

New `test/scanner-js-guard.test.mjs` — the first test to drive `scanner.mjs` directly:
- **happy path**: indexes a nested library, then rescans as a fast-path no-op (exercises the refactored walk + cached sidecar probe + the new per-file write transaction)
- **accessible-but-empty** library: stale tracks are correctly swept
- **inaccessible directory**: scan aborts (exit 1) and leaves tracks intact
- every case uses a real task-queue-shaped config (incl. `waveformCacheDir`), so they regression-guard fix #1

After the rebase + port:
- Full project suite ✅ **1145 tests, 0 fail, 0 skipped** (incl. `scanner-js-guard`, `scanner-parity`, `audio-hash-parity`, `lyrics-parity`, `waveform`, `task-queue`).
- Direct JS-scanner smoke run (initial scan / no-op rescan / force-rescan): row counts **identical to the Rust scanner**, album rows byte-identical across a no-op rescan (guard no-ops correctly), no scanner warnings, `scan_progress` cleaned up.
- No new lint vs. baseline.

## Notes / out of scope
- The Rust scanner's parallel album-art redundancy was addressed in #612 (per-directory `OnceLock` dedup), which this branch is now rebased on top of.
- These fixes only touch the **JS fallback** path; installs running the Rust binary are unaffected. On a JS-fallback host, the first whole-library scan after upgrading will rewrite embedded-art filenames once (old files orphaned, not broken) as a consequence of fix #3.


IrosTheBeggar:
## Rebased onto current master (force-pushed `621ab59b` → `339a2b0a`)

**Shape change:** the branch's two scanner commits were squashed and replayed as one (`0fa7b231`) — they reworked the same file, and replaying them separately would have meant resolving the same conflicts twice. The stray `6.10.1` release bump commit was dropped (master is the source of truth for versions). A second commit (`339a2b0a`) carries post-merge hardening.

**The merge itself** was the intricate part: this PR's restructure of `scanner.mjs` (single-pass `collectFiles` walk; `ensureBatch`/`flushBatch` batching with **flush-before-parse** so extraction never holds the write lock; per-track write transactions; the `isAccessibleDir` + `countLibraryTracks` data-loss guards; art-hash/genre/lyrics parity fixes) had to be combined with everything master gained since this branch's base: #623's UPSERT `insertTrack` + chunked, yielding, schema-guard-aware stale sweep; #624/#626's `expectedSchemaVersion` guard + exit-code mapping; `BEGIN IMMEDIATE` at every transaction open. The merged file carries **both** property sets — the known conflict (this PR's `countLibraryTracks` vs #623's removal of `deleteOldTracks`) resolved as planned: guard statement kept, legacy unchunked delete gone.

**Post-merge adversarial review** (two lenses vs both parents + cold semantic read) confirmed 8 findings; fixed in `339a2b0a`:
- *(medium)* `flushBatch` COMMIT failure wedged `batchOpen=true` forever — auto-rollback-class errors (disk full, I/O) would bury the root cause under hundreds of per-file warnings. Now resets state and fails the scan loudly once.
- *(medium)* a walk that **found files but processed zero** (systemic failure) bypassed the vanished-mount guard and would have let the stale sweep delete the entire library. New guard: skip cleanup, exit 1.
- *(low ×5)* `filesScanned` now counts per-file errors (matching the Rust emitter's contract); the fast-path batch gets the Rust writer's 50ms wall-clock lock cap; `cleanupOrphans` re-verifies `user_version` per chunk like the stale sweep (Rust-side parity queued for the hardening PR); wrapped `db.close()` + rejection backstop so nothing can override a guard's exit 3; UTF-8 BOM stripped.

**Verification:**
- Full suite: 1177/1178 — the one failure (`subsonic-lyrics-lrclib`) is a parallel-load flake: passes 3/3 in repeated isolated runs, and the code this PR touches (`lyrics-extraction.js`) only loads in the JS scanner process, which that test never runs.
- Behavioral smoke of the merged scanner, forked exactly as task-queue does — all six scenarios pass: initial scan (counts exact, FTS parity), UPSERT retag (same rowids + `created_at`, old genre links gone, `MATCH 'RETAG'` = 10), 40-file stale sweep (zero FTS ghosts, zero dead M2M refs, orphans exact), **mid-sweep schema bump → exit 3 exactly one 500-row chunk after the bump, clean convergence after restore**, vanished-root bail (exit 1, zero writes), wrong `expectedSchemaVersion` (exit 3, zero writes).

With this merged, the JS fallback finally stops holding the write lock across whole 25-file extraction batches — the last big contention source the scanner audit found.

IrosTheBeggar:
## Gap-closing smoke results (+1 bug found and fixed: `0879caa5`)

Three scenarios closing the gaps in the earlier smoke pass — all against the final tree:

**G1 — all six behavioral checks re-run post-hardening, plus the new `filesScanned` contract.** All pass. Zero-write proofs now use `PRAGMA data_version` + full before/after snapshots. The error-counting contract was exercised non-trivially: with 2 files exclusively locked mid-scan, `filesScanned: 20 = 18 successes + 2 errors` with correct per-file warnings. (Honest nuance: byte-corrupted "mp3s" don't error — music-metadata parses them as tagless files and they index with null tags, hence the lock-based variant.)

**G2 — the JS fallback under a REAL server, finally measured.** Disposable copy of the tree with the rust scanner broken (`findRustParser` → cargo fails → JS fallback under task-queue, proven via the pidfile: `kind: 'js'`, marker = scanner.mjs). The headline claim of this PR, with numbers:

> 400-track boot scan, 13.5s locked window, concurrent `BEGIN IMMEDIATE` writer probing every 25ms: **375 samples, 0 busy timeouts, p50 7.5ms / p90 33ms / p99 79ms / max 109ms.**

The pre-PR JS scanner held the write lock across whole 25-file decode batches — tens of seconds per batch on this workload. Extraction now provably runs lock-free. The admin-API rescan after deleting 250 files swept exactly 250 (chunked) with the probe still at 0 timeouts.

**G3 — album-art pipeline (`skipImg: false` for the first time) — found a real bug.** The raw-byte-MD5 naming verified exactly against independently computed hashes (`cover.jpg`, `random-name.png`, embedded art with `image/jpeg → .jpeg` mapping and dedup, `zl-`/`zs-` variants, idempotent rescan with zero rewrites). But **`Folder.JPG` still got no art**: the candidate-*name* match was case-insensitive, while the extension filter two lines earlier compared `getFileType()` verbatim against `['png','jpg']` — uppercase extensions were dropped before reaching the candidate list. Fixed in `0879caa5` and re-verified end-to-end (`Folder.JPG` → `albums.album_art_file` = raw-MD5 name, variants written).

Parity nit for the backlog (not this PR): the stored art filename keeps the source extension's case (`.JPG`), so a JS-scanned and rust-scanned cache could name the same cover differently — worst case one duplicate cached image per album when switching scanners.

PR #604, opened 2026-05-29
Fix garbled non-latin filename of uploaded files

donmor:
Filenames with non-latin characters gets garbled when uploaded to mStream. This patch fixes it by recreating the string with utf-8.


IrosTheBeggar:
Good find. I dug into this a bit a found a different approach by having this handled by the busboy library:

```javascript
const bb = busboy({ headers: req.headers, defParamCharset: 'utf8' });
```

This fixes some unusual edge cases where the user sends a custom form fields formatted like: filename*=UTF-8''%E6%AD%8C.mp3. 

Feel free to try this out and update your PR. I'll merge it in later so you get the credit for it

IrosTheBeggar:
Looks good, thanks for helping!

PR #605, opened 2026-05-30
fix(backup): audit findings — follow_symlinks, validation codes, data-safety

IrosTheBeggar:
Audit of the file-backup API + worker/service surfaced a handful of bugs. This PR fixes them, ordered by impact below. All changes are backend-only (DB queries, API validation, worker file-copy logic); each fix carries a comment explaining the failure mode.

### 1. `follow_symlinks` never reached the worker — symlinked tracks silently dropped
`getBackupDestination*` selected `d.*, l.name, l.root_path` but **not** `l.follow_symlinks`, and `backup_destinations` has no such column — so `runBackupTask`'s `dest.follow_symlinks === 1` evaluated `undefined === 1` → **every backup ran with `followSymlinks=false`**, regardless of the library's flag (contradicting the worker's own doc comment). The scan path works because it reads a full `SELECT * FROM libraries` row. Fixed by JOINing `l.follow_symlinks` into all four destination queries.

### 2. Missing `dailyAtHour` returned `500 Server Error` instead of `400`
`requireDailyHour()` throws a plain `Error` and was called **outside** any try/catch (unlike the sibling `validateExcludeGlobs` / `checkPathContainment`). Under Express 5 the throw reaches the global error middleware, which only maps `Joi.ValidationError` / `WebError` → falls through to `500 { error: 'Server Error' }`, swallowing the explanatory message. So `POST/PATCH {triggerType:'daily'}` with no hour (valid per Joi) 500'd. Now caught → `400` with the real message.

### 3. Source dir vanishing mid-walk trashed the dest subtree (data-safety)
`readSortedDir` returns `{ existed:false }` on `ENOENT`, but `syncDir` ignored it and walked dest against an empty source list — turning every dest entry into an "orphan" and trashing it. A source that unmounts (or races) between the parent's stat and the child `readdir` would soft-delete the whole corresponding dest subtree, or with `retentionDays=0` **permanently delete it**. The `existed` flag was already computed but never consumed. Now `syncDir` bails when `existed:false`; a genuine source-side deletion is still handled safely at the parent level (where source legitimately reads `existed:true`). Zero behavior change on the happy path.

### 4. `finished_at` written in two incompatible timestamp formats
`createBackupRunRow` stamped pre-finished rows (skipped / disabled-before-start) with `new Date().toISOString()` (`...T...Z` + ms), while `started_at` and `finishBackupRunRow` use `datetime('now')` (`YYYY-MM-DD HH:MM:SS`). The whole codebase parses these via `s.replace(' ','T')+'Z'` (e.g. the admin UI's `formatTime`), which turns the ISO form into `...ZZ` → `Invalid Date`. Unified on the SQLite datetime format.

### 5. Failing daily destination flooded history with `failed` rows
The scheduler keyed on the last **successful** run, so a fast-failing dest (unplugged drive) was re-triggered on every 5-minute tick — up to ~288 `failed` rows/day, the same flooding the skip-row policy avoids. Now keys on the last run of **any** status: one scheduled attempt per local day. **Behavior note:** a failed daily run no longer auto-retries same-day; manual + after-scan triggers still fire, and an operator can retry manually. Flagging in case a different trade-off (rate-limited retry) is preferred.

### Minor
- **check-path**: an existing **file** destPath left `parentExists=false` (spurious Windows "drive not mounted" warning) and gave no hint the path was a file. Now sets `parentExists` on any successful stat and warns when it's not a directory.
- **history `?limit`**: a negative value reached SQLite as `LIMIT -1` = "no limit". Clamped to `[1, 500]`.
- Dropped a dead argument on the `triggerForLibrary` call site.
- Finding from the audit that turned out **benign** (no change): a non-numeric `:id` → `Number(...)=NaN`; `node:sqlite` binds `NaN` to a no-match, so the existing `!dest` check already returns a clean 404 (covered by a regression test).

## Tests
- `test/backup-db-helpers.test.mjs` — `follow_symlinks` propagation through all destination queries; `createBackupRunRow` timestamp format for running + pre-finished rows.
- `test/backup-routes.test.mjs` — daily-without-hour → 400 (+ message), daily-with-hour clears the gate, malformed `:id` → 404.
- All five are genuine regression guards (fail on the pre-fix code). Both files registered in the `npm test` script.

Notes: existing `test/task-queue.test.mjs` (which spawns real `worker.mjs` processes) still passes, exercising the `existed`-guard change on the happy path. Lint delta is **0** (project baseline 139 unchanged). Fixes 3 and 5 are verified by reasoning/review — deterministic tests would require injecting a filesystem race / time travel.


IrosTheBeggar:
Closing in favor of the 2026-06 full backup audit's fix series, which landed (or queues) everything here with deeper test coverage. Mapping of this PR's fixes:

| #605 fix | Status |
|---|---|
| 1. `follow_symlinks` never reached the worker | **Merged in #724** (batch 1) â€” same four-getter JOIN fix, plus end-to-end wiring tests through the real task queue |
| 2. `dailyAtHour` â†’ 500 instead of 400 | **In #726** (batch 3) â€” via `WebError(400)` + async handlers instead of per-route try/catch |
| 3. `existed`-flag guard (source vanish mid-walk) | **Merged in #725** (batch 2) â€” with a deterministic `vanish-readdir` fault-injection test for the race this PR could only verify by reasoning |
| 4. `finished_at` dual timestamp formats | **Folded into batch 4** (upcoming) â€” including this PR's sharper observation that the `replace(' ','T')+'Z'` consumers render the ISO rows as Invalid Date in the admin UI |
| 5. Daily scheduler failure flooding | **Folded into batch 4** â€” adopting exactly this PR's design (key on last run of any status, one attempt per local day), reconciled with the midnight-straddle dedup issue in the same change |
| check-path file-as-dest handling | **Folded into batch 4** â€” genuinely new find, wasn't in the full audit's 53 findings |
| history `?limit` clamp | **In #726** â€” with `Math.trunc` added (fractional limits also bypassed) |
| `triggerForLibrary` dead arg | Queued in the audit's cleanup batch |

The branch is 6 weeks stale and conflicts with all three landed batches, so closing rather than rebasing. Thanks past-us â€” fixes 4, 5, and the check-path find survive into batch 4.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

PR #613, opened 2026-06-03
Expand track-metadata fields (fidelity + more columns + track/disc totals, schema V43)

IrosTheBeggar:
## Summary

Expands the track metadata mStream's native API returns, and fixes the scanner gaps behind it. Everything is **additive** — no breaking changes.

## 1. Audio fidelity fields (native API)
`renderMetadataObj` (`src/api/db.js`) — the single mapper behind every native metadata endpoint — now emits `bitrate, format, duration, sample-rate, channels, bit-depth, file-size`. Subsonic already exposed these.

## 2. Scanner population (bitrate + file_size)
Neither scanner wrote `tracks.bitrate` / `file_size` (always NULL). Now both do — Rust via lofty `audio_bitrate()` + the existing stat; JS fallback via `music-metadata` + `statSync`. The **JS bitrate is rounded to whole kbps** so it matches the Rust scanner's quantised value (header-reported bitrates match exactly; computed estimates may differ ~1 kbps, inherent to the two libraries).

## 3. More existing columns exposed ("Tier 1")
Pure column maps already on the row: `audio-hash` (preferred stable identity), `created-at`, `modified`, `source`, `bpm-source`, `has-lyrics` / `has-synced-lyrics` / `lyrics-lang`.

## 4. Track/disc totals (schema V43)
- **Schema**: `SCHEMA_V43` adds `tracks.track_total` / `disc_total` (`rescanRequired: true`). `SCHEMA_VERSION` → 43.
- **Scanners**: Rust `track_total()` / `disk_total()`; JS `track.of` / `disk.of`.
- **API**: `track-total`, `disc-total` (pair with the existing `track` / `disk`).

## Verification
- `bitrate` 52/53, `file_size` 53/53 populated on rescan (were 0/53); track/disc totals confirmed via a tagged-file scan.
- Fresh migration yields `track_total` + `disc_total` (no composer); both scanner INSERTs are **31 cols / 31 params**; Rust compiles.
- 350+ tests green across Subsonic / DB / migration / API; OpenAPI validates.

**Binaries:** source-only — both `build-rust-parser.yml` and `-musl.yml` rebuild + commit the platform binaries on merge (`rust-parser/**` trigger). Don't tag the release until those land.

## Deferred to follow-ups
- **Composer** — was prototyped here as a `tracks.composer` column, then **removed**: it belongs in the `track_artists` M2M as `role='composer'` (the documented intent of that table's role enum, and the model every other server uses — Navidrome/Lyrion/Kodi/Jellyfin). Doing it right needs a role-aware query refactor, shared with…
- **Multi-artist lists** — surfacing `track_artists`/`album_artists` with role filtering.
- **Remaining Tier 2** — album ReplayGain + peaks, MusicBrainz IDs (artist/album MBID columns already exist, unpopulated), etc.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -291,6 +291,91 @@ components:
                 `play-count`, `last-played`, `replaygain-track`).
                 The DB column name is `musical_key` (snake_case,
                 SQL convention).
+            bitrate:
+              type: integer
+              nullable: true
+              description: |
+                Audio bitrate in bits per second. (The Subsonic API reports
+                this as kbps.) Populated by both scanners; null on rows
+                scanned before that change until a force-rescan.
+            format:
+              type: string
+              nullable: true
+              description: Container / codec, e.g. `flac`, `mp3`, `opus`.
+            duration:
+              type: number
+              nullable: true
+              description: Track length in seconds.
+            sample-rate:
+              type: integer
+              nullable: true
+              description: |
+                Sample rate in Hz (e.g. 44100, 96000). NULL on rows scanned
+                before schema V16 until a force-rescan repopulates it. DB
+                column `sample_rate`.
+            channels:
+              type: integer
+              nullable: true
+              description: |
+                Channel count (e.g. 2 for stereo). NULL before schema V16
+                until a force-rescan.
+            bit-depth:
+              type: integer
+              nullable: true
+              description: |
+                Bits per sample (e.g. 16, 24). NULL before schema V16 until a
+                force-rescan. DB column `bit_depth`.
+            file-size:
+              type: integer
+              nullable: true
+              description: |
+                File size in bytes. DB column `file_size`. Populated by both
+                scanners; null on rows scanned before that change until a
+                force-rescan.
+            audio-hash:
+              type: string
+              nullable: true
+              description: |
+                V14 audio-payload hash — the preferred stable track identity
+                (survives tag edits, album-art changes, ReplayGain rewrites),
+                unlike `hash` (whole-file MD5). DB column `audio_hash`.
+            created-at:
+              type: string
+              nullable: true
+              description: When the track row was first scanned (≈ date added to the library). DB column `created_at`.
+            modified:
+              type: integer
+              nullable: true
+              description: File modification time, epoch milliseconds. DB column `modified`.
+            source:
+              type: string
+              nullable: true
+              description: Provenance label from embedded tags (V36), e.g. `ytdl`. Null when no recognised marker is present.
+            bpm-source:
+              type: string
+              nullable: true
+              description: Where `bpm` came from (`tag` vs scanner analysis). DB column `bpm_source`.
+            has-lyrics:
+              type: boolean
+              description: |
+                True when the track has embedded or synced lyrics. The lyrics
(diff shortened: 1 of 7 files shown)


IrosTheBeggar:
## Rebased onto current master + smoke-tested (force-pushed `ac9cfd9e` → `9132d2ad`)

**Shape:** the six original commits squashed into one (they layered onto the same statements; original messages preserved in the commit body) and replayed over the UPSERT-era scanners (post #623/#624/#626/#603).

**The two traps this rebase had to avoid, and how:**

1. **Schema-number collision** — this PR was authored as "V43" on a pre-V43 base; master has since shipped a *different* V43 and V44. Keeping `version: 43` would make every production DB (already at `user_version ≥ 44`) silently skip the ALTERs forever while the scanners bind the new columns — every scan dead, unrecoverable by rerun. **Renumbered to V45**, `SCHEMA_VERSION = 45`, `rescanRequired: true` kept. (Also: `bitrate`/`file_size` turn out to be V1 columns that were never populated — the migration only adds `track_total`/`disc_total`.)
2. **Silent UPSERT staleness** — the PR predates the UPSERT and extended a plain `INSERT OR REPLACE`. The four new columns are merged into **both** scanners' `DO UPDATE SET` lists, not just the INSERT lists; a column missing from SET prepares fine but never refreshes on rescans. **New CI test**: parses both scanners' UPSERT statements from source and asserts every inserted column (minus the conflict target) appears in `DO UPDATE SET`, placeholder counts match, the JS and Rust lists are identical, and `created_at` stays unwritten — the next column addition can't repeat this near-miss.

**Smoke tests (all PASS):**

- **M1 — the production upgrade path.** Fabricated a V44 DB with 60 existing tracks (NULL bitrate/file_size, no totals columns), booted a real server on it: `Applying migration v45` exactly once → `rescanRequired` epoch rescan → **all 60 rows backfilled** (bitrate ×1000-quantised, `file_size` byte-exact vs disk) → **row ids preserved through the forced re-parse** (UPSERT identity, 0 mismatches) → `.rescan-pending` cleared → `POST /api/v1/db/metadata` serves `bitrate`/`file-size`/`track-total`/`disc-total` correctly, NULLs where tags have no totals.
- **M2 — cross-scanner parity.** Same 20 files (10 tagged `track=N/12 disc=1/2`) scanned by rust and JS into fresh DBs: all four fields **identical on 20/20 filepaths, zero ±1 kbps divergences** — the quantisation claim holds exactly. UPSERT refresh: retag `3/12 → 3/14`, rescan → `track_total` 12→14, same row id.
- **M3 — fresh installs + guard interplay.** Fresh V45 scans populate everything (both scanners); an old server (`expectedSchemaVersion: 44`) against a V45 DB **refuses with exit 3** in both scanners; legacy payloads (no version field) still scan; mid-scan SIGKILL of the JS scanner converges on rescan with zero duplicate rows.

**Tests:** full suite 1180/1181 (the one failure is the known dlna transcode parallel-load flake — 70/70 in isolation), including the new UPSERT-parity tests.

**Release notes:** `rescanRequired` means everyone gets a one-time boot force-rescan after this merges. The Rust half ships once CI rebuilds `bin/rust-parser/`.

PR #620, opened 2026-06-06
perf: tune main DB connection + brotli/gzip response compression

IrosTheBeggar:
## Summary

Two connection/transport-level performance wins that touch ~every request, plus a configurable write-durability knob. Follows up #619 (query-level work).

## 1. SQLite connection tuning — `manager.js`

The main server connection (which serves **every** HTTP request) only set `WAL` + `busy_timeout`. Added the scanner's read/sort tuning, which are pure wins with no durability trade-off:
- `cache_size = -65536` — 64 MB page cache (vs the 2 MB default) keeps more of the DB/indexes hot.
- `temp_store = MEMORY` — sort/`GROUP BY` temp B-trees (recently-added sort, search, stats) build in RAM instead of on disk.

## 2. Configurable `synchronous` (default **FULL**) + Admin API/UI

`scanner.mjs` documents that the main connection *deliberately* stays on `synchronous=FULL` because user data (scrobbles, ratings, playlists) isn't re-derivable from files. So instead of silently switching to `NORMAL`, this exposes it as a setting:

- **`config.db.synchronous`** — `FULL | NORMAL`, default `FULL`.
- **`manager.setSynchronous()`** — allowlisted, applies the PRAGMA to the **live** connection (per-connection + live-settable → **no restart needed**).
- **`util/admin.editDbSynchronous()`** — persists to config + applies live.
- **Admin API** — `GET /api/v1/admin/config` exposes `dbSynchronous`; `POST /api/v1/admin/config/db-synchronous`.
- **Admin UI** — a new "Database" card with a FULL/NORMAL toggle (default UI).

`NORMAL` skips the per-commit fsync (faster scrobbles/ratings/playlist saves) and is crash-safe under WAL (never corrupts), but a hard power loss can lose transactions committed since the last WAL checkpoint — hence opt-in, default off.

## 3. Brotli/gzip response compression — `util/compression.js` + `server.js`

New middleware (first in the chain) on Node's built-in `zlib` — **no new dependency** for the Electron build. Prefers brotli, falls back to gzip.

- **Content-Type gated**: only text payloads (JSON, HTML, JS, CSS, XML, SVG) are compressed. `audio/*`, `image/*` (except SVG), video and `octet-stream` pass through untouched — so **audio streaming and HTTP range/seek requests are unaffected** (the one thing that must not break).
- **Buffers** the (bounded) text body and compresses once with an accurate `Content-Length`. Streaming through a zlib transform deadlocks piped responses (`express.static`) on the socket-drain/Content-Length mismatch — buffering sidesteps that.
- Skips `204/304/206`, `no-transform`, already-encoded, HEAD and tiny bodies.

## Testing

- **Server smoke test** (real boot): br/gzip round-trip on the static `GET /` (express.static stream) with matching `Content-Length`, image/png passthrough (gate), and the `db-synchronous` admin round-trip — FULL→NORMAL applied **live** (no reboot), persisted to `config.json`, invalid values rejected.
- **Middleware in isolation**: br/gzip round-trips, `audio/mpeg` left uncompressed with bytes intact, below-threshold skipped.
- **Regression** (compression active, wraps all responses): `subsonic` 67/67, `dlna` 70/70, `playlist-load` 6/6, `file-explorer` 4/4 — XML/SOAP, JSON, and audio-passthrough all transparent.
- Lint clean (no new problems); `compression.js` clean.

## Reviewer notes
- `PRAGMA synchronous` is live-settable per connection, so no `reboot()` is needed for that toggle (confirmed empirically). The existing `server.js reboot()` is unchanged.
- Only the default admin UI got the toggle; the velvet admin (a near-clone) reads the same `GET /admin/config` and can mirror it later if wanted.


IrosTheBeggar:
Rebased onto master `08607759` (post-#630). One conflict, mechanical: the package.json test list (master had added `scanner-js-guard` / `scanner-schema-guard` / `transcode-route` since this branch forked) — resolved as the union with `compression-config.test.mjs` appended. All other files auto-merged; verified the `initDB()` ordering survived intact (PRAGMAs → `setCacheSize`/`temp_store`/`setSynchronous` → `runMigrations`).

Re-verified on the rebased head:
- **1199/1199** full suite (master's 1188 + this PR's 11 compression/cache tests), 0 skipped — including `subsonic-lyrics-lrclib` with this branch's non-flaky miss-TTL fix.
- Lint at exactly the master baseline (136 problems, none new).

No semantic interaction with the scanner-campaign merges — the main-connection `cache_size`/`temp_store` tuning mirrors what the scanner connection already does, `synchronous` stays FULL by default, and compression defaults to `none` until an operator opts in.

PR #623, opened 2026-06-09
perf(scanner): cut DB-writer-lock contention + speed up decode

IrosTheBeggar:
## Why

Audit of the Rust scanner (and its JS twin `src/db/scanner.mjs`) targeting the intermittent **"server DB writes hang during a scan"** symptom — the scanner holds the single WAL writer lock long enough that concurrent API writes (scrobble / star / playlist save / play-queue) exhaust the 5 s `busy_timeout` — plus raw scan throughput. A multi-agent review surfaced the findings; each was adversarially verified against the code before landing here.

### Contention (the hang)
- **Time-bounded commit + writer yield** — the parallel writer now ends a batch on `scanCommitInterval` rows **or** ~50 ms wall-clock (a heavy `commit_track` row costs far more under the lock than a cheap `scan_id` bump), and sleeps ~3 ms after a *capped* batch so a waiting API write can win the lock (SQLite's busy handler isn't fair). Skipped when the channel drains naturally → quiet rescans pay nothing.
- **Chunked stale-track DELETE** — `DELETE FROM tracks WHERE scan_id != ?` was a single unchunked statement that held the lock through the whole cascade + per-row FTS delete trigger. Now 500-row autocommit batches in both scanners (new `deleteStaleTracks` in `orphan-cleanup.js`), matching the orphan cleanup that was already chunked.
- **`INSERT OR REPLACE` → `ON CONFLICT … DO UPDATE … RETURNING id`** — one column-scoped FTS update per changed track instead of delete+insert (2 FTS ops + cascade churn + rowid churn).

### Speed
- **`opt-simd` + thin-LTO + `codegen-units=1`** in `Cargo.toml` — enables rustfft's SSE/AVX/NEON inverse-MDCT for the MP3/Vorbis/AAC decode that is ~90 % of scan wall-time, plus cross-crate inlining of the downmix/hash loops. `panic` stays `unwind` (`abort` would break the parallel writer's `std::thread::scope` worker-panic handling). rustfft does runtime CPU detection, so the generic prebuilt binaries stay safe on older CPUs.

### Hygiene
- **Schema V44** drops the redundant `idx_tracks_filepath` (duplicate of the `UNIQUE(filepath, library_id)` auto-index) — pure write-amplification on the hottest INSERT/UPSERT path.
- Hoisted `fs::create_dir_all` out of the per-track waveform write to once at scan start.

## Behavior improvements from the UPSERT
- **`created_at` is now preserved across rescans.** The old `INSERT OR REPLACE` silently reset it on every rescan of a *changed* file, so the V43 "recently added" sort was effectively "recently modified". Fixed.
- **Track ids are now stable across rescans.**

The UPSERT made `DELETE FROM track_artists` load-bearing (kept) and required an explicit `DELETE FROM track_genres` before repopulating, since the row no longer gets cascade-cleared.

## Testing
- **Scanner parity / determinism** (`scanner-parity.test.mjs`): two-scan determinism, `scanThreads=1` vs `=4` parity, parallel rescan with mixed changed/unchanged files, resume/no-op fast paths — all green.
- **audio-hash / lyrics / waveform** parity suites green (waveform determinism confirms `opt-simd` is reproducible within a build).
- **FTS5 / orphan-cleanup / schema** suites green, incl. "SCHEMA_VERSION is the latest migration" and "applying all migrations leaves user_version = SCHEMA_VERSION" (V44 applies cleanly).

## Note for release
The DB-contention fixes and decode speedups live in the **Rust binary**, so production installs get them only once CI (`build-rust-parser.yml`) rebuilds and re-commits the prebuilt binaries in `bin/rust-parser/`; local dev builds (`npm run build-rust`) pick them up immediately. The JS-fallback half (chunked delete, UPSERT) is live on merge.

## Not in this PR (audited backlog)
Defer waveform/BPM decode out of the DB-holding scan (~10× faster initial-scan library availability), raise the half-cores worker cap on decode-bound initial scans, WAL-checkpoint management (`wal_autocheckpoint` + end-of-scan `wal_checkpoint(TRUNCATE)`), merge the two `read_dir`-per-directory caches.

### DIFF
--- a/rust-parser/Cargo.lock
+++ b/rust-parser/Cargo.lock
@@ -691,6 +691,7 @@ dependencies = [
  "bytemuck",
  "lazy_static",
  "log",
+ "rustfft",
 ]
 
 [[package]]
--- a/rust-parser/Cargo.toml
+++ b/rust-parser/Cargo.toml
@@ -37,6 +37,16 @@ symphonia = { version = "0.5", default-features = false, features = [
     "isomp4",
     "ogg",
     "wav",
+    # opt-simd gates the rustfft-backed (SSE/AVX/NEON) inverse-MDCT path
+    # used by the transform codecs (MP3 hybrid filterbank, Vorbis, AAC) —
+    # the per-packet hot loop that dominates waveform/BPM decode, which is
+    # ~90% of scan wall-time. Without it those decode on rustfft's scalar
+    # fallback. rustfft does runtime CPU detection, so AVX is used only
+    # when present and otherwise falls back — safe for the generic
+    # prebuilt binaries. (Do NOT pair this with `-C target-cpu=native`:
+    # the committed binaries must run on older CPUs.) FLAC/WAV are integer
+    # codecs and don't touch the MDCT, so the win is codec-mix-dependent.
+    "opt-simd",
 ] }
 # File-level parallelism for the scan loop. Pure Rust, no C deps —
 # fits the same philosophy as md-5 / image-rs above. Workers run
@@ -61,3 +71,12 @@ stratum-dsp = "=1.0.0"
 
 [profile.release]
 strip = true
+# thin-LTO inlines across the symphonia/rustfft/md-5 crate boundaries into
+# the per-sample downmix + hashing hot loops (full LTO buys little more here
+# but costs much longer CI builds); codegen-units = 1 lets the optimizer see
+# each crate whole. Together ~5-15% across the decode + hash pass on top of
+# the opt-simd MDCT win above. panic stays at the default 'unwind' — NOT
+# 'abort' — because the parallel writer relies on std::thread::scope unwinding
+# a worker panic rather than killing the process mid-transaction.
+lto = "thin"
+codegen-units = 1
--- a/rust-parser/src/main.rs
+++ b/rust-parser/src/main.rs
@@ -203,7 +203,9 @@ struct ExtractedTrack {
 
     // Captured from the prior tracks row (when one existed) so the
     // writer can run user_*-row + album-stars migrations after the
-    // INSERT OR REPLACE swaps the canonical identity.
+    // UPSERT in commit_track refreshes the canonical identity (the row
+    // is updated in place — same rowid — so these pre-images are the
+    // only record of what the identity used to be).
     old_hash: Option<String>,
     old_audio_hash: Option<String>,
     old_album_id: Option<i64>,
@@ -509,8 +511,30 @@ fn chunked_orphan_delete(
     loop {
         let changes = stmt.execute([])?;
         if changes == 0 { break; }
+        // Back-to-back chunks free the writer lock for only microseconds
+        // — useless to the server's busy handler (one poll per 100ms in
+        // its tail). A full chunk means more work likely remains, so
+        // open a real window. Same rationale + width as WRITER_YIELD.
+        if changes == ORPHAN_CHUNK_SIZE {
+            chunk_yield();
+        }
     }
     Ok(())
+}
+
+// Inter-chunk writer yield shared by the chunked delete loops: 10-20ms,
+// jittered so the cadence can't phase-lock with the server's 100ms
+// busy-retry tail. See the WRITER_YIELD comment block for the model.
+// Wall-clock subsecond nanos as the entropy source — Instant is an
+// opaque monotonic point (elapsed-since-now is always ~0), and pulling
+// in a rand dependency for one modulus would be overkill.
+fn chunk_yield() {
+    let nanos = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.subsec_nanos())
+        .unwrap_or(0);
+    let jitter = u64::from(nanos) % WRITER_YIELD_JITTER_MS;
+    std::thread::sleep(Duration::from_millis(WRITER_YIELD_MIN_MS + jitter));
 }
 
 // Per-chunk row cap for the end-of-scan stale-track sweep. Same value as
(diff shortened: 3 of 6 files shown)


IrosTheBeggar:
## Rebased + hardened (force-pushed `63acc1a6` → `ac5cae70`)

**Rebase:** now on current master (post #624/#625/#626/#627). Auto-merge was clean; the #626 seams (schema guard at open + pre-sweep re-check in both scanners) were verified hunk-by-hunk — ordering holds: guard → chunked sweep, sweeps in autocommit.

**New commit on top** — from an adversarial multi-agent review of the rebase + an A/B contention benchmark:

1. **Writer yield 3ms → 10-20ms jittered.** Modeled against the server's actual busy-handler retry schedule (dense for ~330ms, then one poll per 100ms; no `SQLITE_ENABLE_SETLK_TIMEOUT` in node's bundled SQLite): a fixed 3ms window still left p90 waits in the seconds, a few-percent chance of exhausting the full 5s timeout, and could phase-lock with the 100ms tail. Still gated on `hit_cap` — quiet/decode-bound scans pay nothing.
2. **The same yield now runs between chunks** of the chunked stale-track and orphan deletes (both scanners) — back-to-back autocommit chunks freed the lock for only microseconds, a ~0% duty cycle for the whole sweep. JS-side it's opt-in (`yieldBetweenChunks`) because `admin.js removeDirectory` runs the same helpers in the *server* process, where a synchronous sleep would block the event loop with nobody to yield to.
3. **`BEGIN IMMEDIATE` in the scanners' own batch transactions** (all four sites, rust + JS). Batches often open with a read (cache-miss artist/album lookup, `getTrack`), so a deferred `BEGIN` pins a snapshot that a server commit invalidates → `SQLITE_BUSY_SNAPSHOT`, bypassing `busy_timeout` — the class #624 fixed server-side. The kicker: this PR's yield makes server commits land in scanner gaps *more* often, so the fix that helps the server raised the odds of poisoning a scanner batch (whose swallowed per-file errors feed the stale sweep). IMMEDIATE closes it on both sides.
4. **Schema guard re-armed per chunk.** #626's mid-scan `user_version` re-check guarded what used to be one atomic DELETE; chunking made the destructive phase an unbounded sequence of transactions with deliberate gaps. The stale sweeps now re-verify before every chunk and abort with the guard sentinel (exit 3) if a migration lands mid-sweep.
5. **`scan_progress` UPDATE moved inside the batch txn** (one fewer lock acquire per interval, atomic with the rows it describes), and its errors now abort the batch like a COMMIT failure instead of being swallowed inside the transaction.
6. Three comments still describing the removed `INSERT OR REPLACE` cascade semantics reworded for the UPSERT reality.

**Measured** (800-track metadata-only force-rescan ×4, concurrent writer probing `BEGIN IMMEDIATE` + INSERT every 25ms, `busy_timeout=5000`):

| | master writer | this branch |
|---|---|---|
| concurrent write p50 | 6.9ms | 7.4ms |
| concurrent write p90 | 122ms | 44ms |
| concurrent write p99 / max | **5503ms** | **82ms** |
| hard `SQLITE_BUSY` (5s exhausted) | **1** | **0** |
| rescan wall time | 6.9s | 1.9s |

The baseline reproduced the historical stall (a full 5s `busy_timeout` exhaustion) inside a 7-second test window.

**Tests:** full suite green on the rebased branch (1175 tests; one parallel-run flake in `subsonic-polish` passes 18/18 in isolation — the known transcode-readiness race, unrelated). `scanner-parity` ran against the rebuilt binary. Reminder from the original description: the Rust half reaches production only after CI rebuilds `bin/rust-parser/`.

IrosTheBeggar:
## Smoke-test results (real binary, real deletions, real concurrent writes)

Five end-to-end scenarios against `ac5cae70`'s freshly built binary — all **PASS**:

| # | Scenario | Key evidence |
|---|----------|--------------|
| S1 | Chunked stale sweep: delete 600 of 800 real tracks, rescan under a concurrent `BEGIN IMMEDIATE` writer | `staleEntriesRemoved: 600` exact; a 2ms-resolution reader caught the sweep mid-flight at 800→**300**→200 (one full 500-row chunk, then the 100-row chunk **16ms later** — the jittered yield, visible in the wild); one probe write committed *inside* the inter-chunk window (25.9ms wait); **0 timeouts**; FTS ghosts 0, all dead-ref/orphan probes 0 |
| S2 | UPSERT claims: retag 20 of 60 files in place, rescan | All 20: **same rowid, same `created_at`**, new title in tracks *and* FTS (`MATCH 'RETAG'` = exactly 20), old genre association gone (the explicit-DELETE path), no duplicate M2M rows; 40 untouched files byte-identical rows; third no-change rescan: `filesUnchanged: 60, filesProcessed: 0` |
| S3 | Mid-sweep schema-guard abort (rust): 25k synthetic stale rows → ~50-chunk sweep; `user_version` bumped mid-sweep | **Exit 3**, stderr `schema-version guard … mid-sweep`, no `scanComplete`; sweep halted after exactly 1,000 of 25,000 deletions (≤ one in-flight chunk of abort latency — the per-chunk granularity working as designed); 50 real tracks untouched; recovery rescan swept the remaining 24,000 cleanly |
| S4 | JS fallback parity | Same four behaviors verified against `scanner.mjs`: exact stale accounting + 0 probe timeouts, mid-sweep guard exit 3, stable ids/`created_at`/genre-swap on retag |
| S5 | Full server e2e (port boot, 800-track scan, API search, delete 500, admin-API rescan under write hammer, quiet rescan) | `totalFileCount` 800→300 exact, deleted titles unsearchable via `/api/v1/db/search`, scan exit 0 with no error-level log lines, pidfile cleared, quiet-rescan fast path clean |

One non-bug worth recording: S3's synthetic rows initially had NULL `modified`, which the rust scanner's `load_existing_tracks` reads strictly as `i64` (exit 1). Audited all production write paths (rust insert, scanner.mjs, both ytdl.js inserts) — every one always sets `modified`, so the strict read is safe against real data; only hand-seeded rows can trip it. Same strict-typed-read family as the `lyrics_sidecar_mtime` item already on the audit backlog.

PR #629, opened 2026-06-10
fix(scanner): data-loss hardening — verify-absence sweep, rollback hygiene, tolerant reads, died-on-arrival fallback

IrosTheBeggar:
The scanner audit's data-loss cluster, fixed: every swallowed error used to feed the stale sweep, and the sweep deleted on circumstantial evidence (an unstamped `scan_id`). The theme of this PR: **a row is deleted only on proof its file is gone, and every ambiguity resolves toward keeping data.**

## What changed

**1. Verify-absence stale sweep (both scanners).** Presence is decided from the candidate's **parent-directory listing**, not a per-file stat: exact-case name present as a regular file (or a symlink resolving to one under `followSymlinks`) → kept and re-stamped `"<scanId>-kept"`; name absent or directory ENOENT → deleted; listing unreadable for any other reason → left untouched (**fail closed** — EACCES/EIO on a degraded mount is not proof of deletion). The listing approach simultaneously fixes the case-insensitive-stat trap (a case-only rename now converges instead of resurrecting the old-casing row forever) and amortizes mass deletions to one `readdir` per directory instead of a network round-trip per gone file. Keyset pagination guarantees termination; the library root is re-verified every chunk so a mount vanishing mid-sweep aborts instead of erasing the library.

**2. Walk errors shield, not veto.** Enumeration failures are classified (dangling symlinks/loops are content, not lost subtrees) and recorded as per-path shield prefixes: rows under a failed subtree are untouched this scan while the rest of the library cleans normally — so one permanently unreadable `#recycle`-style dir can't freeze cleanup forever. `walkErrors` rides the `scanComplete` event and surfaces at warn level in the server log.

**3. Rollback hygiene (rust).** The serial path's write failure used to fall through to **COMMIT**, persisting half-written tracks — now it rolls back. Both rollback paths evict all four memo caches (artists, albums, genres, `various_artists_id`): a rolled-back insert leaves cached ids dangling, or reassigned to a *different name* (`sqlite_sequence` rolls back too), silently misattributing every later track.

**4. Tolerant reads + V46 repair.** One REAL-poisoned `lyrics_sidecar_mtime`/`modified` row (fractional `stat.mtimeMs` from older JS scanners) used to abort the **entire** rust scan. The rust reads are now type-tolerant (NULL `modified` maps to a forced re-parse), the JS writers truncate, and migration **V46** (no rescan) CASTs already-poisoned rows clean.

**5. Died-on-arrival fallback (task-queue).** A rust binary that execs but dies instantly (linker abort, stale arg format, instant panic) used to log "completed with code N" at INFO and be re-spawned forever, leaving the library unindexed. Now: `signal === null && exit ∉ {0, 3} && zero stdout && never completed` → error log, rust disabled for the process lifetime, JS fallback runs the scan. The signature is precise, not inferential — the binary emits a `{"event":"scanStart"}` banner as its *first* statement, so a downed mount, busy DB, or poisoned row (all stderr-only, transient) can never be mistaken for a dead binary, and signal-kills are excluded.

**6. Parity items:** rust's orphan-delete loops gain the per-chunk schema guard; directory-art extensions lowercase in both scanners.

## Review + smoke

The first draft of verify-absence went through a 4-lens adversarial review that confirmed 23 findings — most importantly that it treated *every* stat error as proof of absence, which would have defeated the protection under exactly the degraded-mount conditions it exists for. The listing-based fail-closed design above is the result.

Fault-injection smoke (all PASS, exact evidence in the scenario logs):

| Fault injected | Expected & observed |
|---|---|
| 3 files exclusively locked (FileShare.None) during a rescan, 5 others deleted | Locked rows survive as `"<scanId>-kept"` with warnings; exactly the 5 deleted rows swept; after unlock, next scan re-parses the 3 back to normal. Same behavior in the JS scanner (EBUSY path) |
| `icacls`-denied artist directory | `walkErrors: 1`, its 7 rows completely untouched (original stamps), rest of library cleans; after ACL restore, full convergence. **Zero shielded rows were ever deleted** |
| REAL-poisoned `modified`/`lyrics_sidecar_mtime` + NULL `modified` on a V45 DB | Rust rescan exits 0 (old binary died with InvalidColumnType), poisoned rows re-parse to clean integers, nothing deleted; server boot applies V46 exactly once, `1781126455012.5 → 1781126455012` exact, no rescan marker |
| `whoami.exe` masquerading as the rust binary | `Rust parser died on arrival (exit 1, no output)` at ERROR, JS fallback completes the 60-file scan, second scan goes straight to JS with no second DOA |
| `taskkill /F` on the rust child mid-scan (healthy binary) | No fallback, no disable — banner-based `sawOutput` carries the exclusion on Windows (where kills surface as `code=1, signal=null`); next scan uses Rust again |

The smoke also caught the rust emitter absorbing errored files into `filesUnchanged` (contract skew vs the JS scanner) — fixed in the second commit.

**Tests:** 1186/1186 full suite; 5 new unit tests (verify-absence keep/delete/sentinel, prefix shielding, case-rename convergence, symlink walk-fidelity, V46 exact repair).

## Notes

- The `-kept` sentinel intentionally keeps rescued rows in the next scan's candidate set, and a resumable boot-rescan epoch does **not** mistake them for already-re-parsed.
- Known minor follow-up (benign, one-time): switching scanners can re-parse files whose stored `modified` differs by ~1ms between rust and JS (truncation-vs-rounding of NTFS fractional timestamps); the scan after that is stable.
- The rust half ships once CI rebuilds `bin/rust-parser/`.

### DIFF
--- a/rust-parser/src/main.rs
+++ b/rust-parser/src/main.rs
@@ -1035,6 +1035,12 @@ fn run_scan(config: &ScanConfig) -> Result<(), Box<dyn std::error::Error>> {
 
     let mut file_count = 0u64;      // new/modified files parsed
     let mut total_processed = 0u64; // all files touched (including unchanged — for progress)
+    let mut error_count = 0u64;     // per-file failures — kept separate so
+                                    // filesUnchanged doesn't absorb errored
+                                    // files (the scanComplete contract is
+                                    // visited = processed + unchanged +
+                                    // errors; mirrors errorCount in
+                                    // src/db/scanner.mjs)
     // Commit cadence: doubles as progress-update cadence and write-lock release.
     // Lower = more responsive API writes during scans but more COMMIT/BEGIN overhead.
     // Admin-configurable via scanCommitInterval; default (25) is a balanced starting point.
@@ -1078,6 +1084,7 @@ fn run_scan(config: &ScanConfig) -> Result<(), Box<dyn std::error::Error>> {
                 Err(e) => {
                     eprintln!("Warning: failed to process {}: {}", entry.path().display(), e);
                     total_processed += 1;
+                    error_count += 1;
                     continue;
                 }
             };
@@ -1116,6 +1123,7 @@ fn run_scan(config: &ScanConfig) -> Result<(), Box<dyn std::error::Error>> {
                     // a compilation track can CREATE the "Various Artists"
                     // row inside the rolled-back transaction.
                     *various_artists_id.lock().unwrap() = None;
+                    error_count += 1;
                     eprintln!("Warning: failed to commit {}: {}", entry.path().display(), e);
                 }
             }
@@ -1282,9 +1290,12 @@ fn run_scan(config: &ScanConfig) -> Result<(), Box<dyn std::error::Error>> {
                                 .and_then(|mut stmt| stmt.execute(rusqlite::params![config.scan_id, existing_id]))
                             {
                                 Ok(_) => {}
-                                Err(e) => eprintln!(
-                                    "Warning: scan_id update failed for {}: {}", rel, e
-                                ),
+                                Err(e) => {
+                                    error_count += 1;
+                                    eprintln!(
+                                        "Warning: scan_id update failed for {}: {}", rel, e
+                                    );
+                                }
                             }
                         }
                         Ok(ExtractResult::Extracted(et)) => {
@@ -1322,15 +1333,19 @@ fn run_scan(config: &ScanConfig) -> Result<(), Box<dyn std::error::Error>> {
                                     // "Various Artists" row inside the
                                     // rolled-back savepoint.
                                     *various_artists_id.lock().unwrap() = None;
+                                    error_count += 1;
                                     eprintln!(
                                         "Warning: failed to commit {}: {}", rel, e
                                     );
                                 }
                             }
                         }
-                        Err(e) => eprintln!(
-                            "Warning: failed to process {}: {}", rel, e
-                        ),
+                        Err(e) => {
+                            error_count += 1;
+                            eprintln!(
+                                "Warning: failed to process {}: {}", rel, e
+                            );
+                        }
                     }
                     total_processed += 1;
                     batch += 1;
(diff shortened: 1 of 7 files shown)


IrosTheBeggar:
## Second verification round: re-review of the redesigned code + gap-closing smoke (`6aaecf10`)

The first review judged the *draft*; the listing-based redesign deserved its own hostile pass. Two cold-read reviewers + three smoke scenarios closed every coverage gap:

**Smoke (all PASS):**

| Gap closed | Result |
|---|---|
| Keyset pagination at scale | 2,600 stale rows swept in observed 500-row chunks (a 2ms poller caught all five boundaries: 2650→2150→1650→1150→650→550), with an icacls-denied dir shielding exactly its 500 rows in the same sweep |
| Mid-sweep root vanish | Library root renamed away during the sweep (2ms reaction inside a yield window): **halted after exactly one chunk** (500 of 1500), "aborting stale-track cleanup … to avoid wiping the library", full convergence after restore |
| JS shields end-to-end | Real denied dir through the JS scanner: its 10 rows untouched, 5 genuine deletions swept, convergence after ACL restore |
| Rust serial path (`scanThreads:1`) | Locked-file survivors `-kept` + recovery through the serial ROLLBACK arm |
| `-kept` × resumable epoch | Same scanId reused three times: kept rows re-examined (not mistaken for re-parsed) and recovered with correct metadata |
| Dangling symlink (followSymlinks) | `walkErrors: 0` (benign — logged, never shields), sweep precision unaffected |
| Banner compatibility | `--waveform` probe contract intact, garbage-input path prints no banner, raw stdout contract exact, server boot clean |

**Re-review confirmed 8 findings, all fixed in `6aaecf10`:**
- *(medium)* The rust listing builder failed **open** twice — per-entry `read_dir` errors silently dropped (unlisted live files deleted) and `file_type()` failures classified as "not a file". Entry errors now poison the listing to unverifiable; type failures map to an Unknown kind → row skipped.
- *(medium)* **The cross-scanner churn root cause**: Node's `stat.mtime.getTime()` *rounds* the fractional ms while rust *truncates* — every scanner alternation re-parsed ~half the library off a phantom 1ms drift. `collectFiles` now stores `Math.trunc(stat.mtimeMs)`. (This deletes the "known minor follow-up" from the PR description.)
- *(low ×6)* Walk-faithful extension filter (rows for de-configured extensions converge again instead of becoming immortal `-kept` rows); rust `NotADirectory` ≙ JS `ENOTDIR`; two "provably gone" look-alikes now detected before deleting (mount vanishing inside a chunk; a down symlinked mountpoint under followSymlinks); `launchJsScanner` gets a latched error-or-close handler so a failed fork can't crash the server and wedge the task queue.

**Final state:** 1186 tests (1185 in the parallel run + the known lrclib load-flake, 17/17 in isolation), rust build clean, lint at baseline.

PR #633, opened 2026-06-11
fix(db): V18 albums rebuild cascade-deletes album stars and nulls track album links

IrosTheBeggar:
## What changed

`SCHEMA_V18` rebuilds the albums table with the `DROP TABLE` + rename pattern. The migration runner (`src/db/manager.js`) sets `PRAGMA foreign_keys = ON` before migrating, and per SQLite semantics a `DROP TABLE` under `foreign_keys=ON` performs an implicit `DELETE FROM` first — which **fires foreign-key actions** on child tables (it skips triggers, but not FK actions).

At V18 time, `albums` has two FK children:

- `tracks.album_id` (`ON DELETE SET NULL`, V1)
- `user_album_stars.album_id` (`ON DELETE CASCADE`, V11)

So every upgrade from `user_version <= 17` nulled all track→album links and **permanently deleted all album stars**. The album-link loss was masked by V18's forced rescan; the star loss was silent — and contradicted V18's own comment claiming the album-migration helper (`src/db/album-migration.js`) remaps stars during the rescan. The helper can never see rows that were already cascade-deleted at migration time.

The fix applies the same TEMP-table dance `SCHEMA_V24` already uses for its tracks rebuild: snapshot `user_album_stars` and the `tracks(id, album_id)` mapping into TEMP tables, let the drop fire, restore after the rename. Album ids are copied verbatim into `albums_new`, so the restored rows pass FK checks.

## Why rewriting the shipped migration in place is safe

The runner only applies migrations with `version > user_version`. DBs that already ran the lossy V18 never see the new text; only DBs still at `<= 17` — which would otherwise hit the data loss fresh on their next upgrade — get the fixed version. Fresh DBs run the dance against empty tables (no-op).

## Also in this PR

- Corrected V24's comment claiming V18 "didn't need this dance" (it did), and V18's comment claiming FKs survive the rebuild unaided.
- New regression test `test/db-v18-albums-rebuild.test.mjs`: builds a populated V17 fixture (albums, tracks, album stars across two users, an artist star as control) and upgrades it under `foreign_keys=ON`, both V17→V18 and V17→full chain. It fails 5/9 against the old migration text and passes 9/9 with the fix. Also asserts no `_v18_*` TEMP tables leak and that CASCADE still works normally post-rebuild.
- `test/helpers/apply-migrations.mjs` gains an exclusive `fromVersion` lower bound, mirroring the runner's `version > currentVersion` gate, so tests can upgrade populated fixtures.

## Reviewer notes

- I audited the rest of the chain for the same hazard: `user_artist_stars` references `artists`, which V18 never drops, and V39's `managed_torrents` rebuild has no FK children — V18 was the only lossy drop.
- `PRAGMA foreign_keys = OFF` inside the migration is not an alternative: the runner wraps each migration in `BEGIN IMMEDIATE`, and the pragma is a no-op inside a transaction.
- Full suite: 1175/1175 pass; eslint clean on changed files.


IrosTheBeggar:
Reviewed with a 4-lens adversarial panel + empirical verification; pushed two fixes and rebased onto current master (747df288, post-V51).

## The fix itself is correct — verified beyond the PR's own tests

- Reproduced the bug on master (regression test fails 5/9 there, 9/9 here), under the exact production wrap: file DB + WAL + `foreign_keys=ON` + `BEGIN IMMEDIATE` per migration.
- Full V17→V51 chain on a populated fixture: stars verbatim, links intact, clean `foreign_key_check`/`integrity_check`, no TEMP leakage, rollback-safe mid-migration, `sqlite_sequence` carried correctly across the drop+rename.
- The rewritten chain produces a **byte-identical schema** to master's on every upgrade path (fresh 0→51, upgrade-from-17, V1–17 prefix) — only the V18/V24 sql text differs, V24's diff being comment-only.
- In-place rewrite safety confirmed: nothing applies migrations outside `runMigrations`' `version > user_version` gate (loki import runs after; rust-parser and the art worker only *compare* `user_version`).

## Fix 1 (pushed): the restore UPDATE was O(n²) — multi-minute boot stall

The track-link restore's correlated subquery scanned the unindexed TEMP backup once per track (`EXPLAIN QUERY PLAN`: `CORRELATED SCALAR SUBQUERY → SCAN b`). Measured inside the boot transaction: 6.2 s @ 10k tracks, 83 s @ 40k, **7.7 min @ 100k** — clean quadratic ratios. And this fires on a *real* path: v6.0–v6.4 shipped schema 8, so those installs run V18 today with a fully populated tracks table. One `CREATE INDEX` on the TEMP table → **402 ms @ 40k** (re-measured post-fix); the index drops with its table.

## Fix 2 (pushed): the regression test was dead code

`test/db-v18-albums-rebuild.test.mjs` was never added to package.json's explicit `npm test` file list — and no GitHub workflow runs the node test suite at all, so that list is the only gate. Wired in.

## Worth knowing (no action taken)

- **Blast radius of the star loss is ~zero for released users**: release tags jump from schema 8 (v6.4.4) to schema 23 (v6.5.0), so a release-path upgrader creates `user_album_stars` (V11, empty) and runs V18 in the same boot — nothing to lose. Permanent star loss only ever hit git-master snapshots from the Apr 20–22 dev window. The fix's value is making the still-live legacy path correct + fixing the false comments the next table rebuild would copy (post-V48, with art junctions CASCADE-ing off tracks/albums/artists, that copy would be expensive).
- A DB containing FK-orphan rows (written out-of-band under `foreign_keys=OFF`) now hard-fails V18 in a boot loop where the old text silently scrubbed them. Narrow reachability (every in-repo writer enforces FK); a `WHERE EXISTS` filter on the two restores would degrade gracefully if we ever care.
- Pre-existing, not introduced here: the rebuild regresses `sqlite_sequence` to max(copied id) (same in master's V18 and V24); mitigated by the forced rescan's album remap.

## Verification on the pushed head (fbbc5c98)

Full suite **1265/1265** (1256 master + 9 new), lint at exact master baseline (136 = 89/47), 40k-track timing check `{"v18Ms":402,"links":40000,"stars":2,"leftovers":0,"pass":true}`.

PR #641, opened 2026-06-14
Advertise the API over mDNS/DNS-SD for LAN discovery

IrosTheBeggar:
## What

Advertises the mStream API as a `_mstream._tcp` **mDNS/DNS-SD** service so
zero-config LAN clients can find the server without anyone typing an IP. This is
the discovery half of the upcoming portable-player pairing story, but it's
generic LAN service discovery — any DNS-SD browser (`avahi-browse`, `dns-sd`,
mobile apps) will see it too.

Advertise-only: it publishes metadata via TXT records and touches **no routes
and no auth**.

## How

Hand-rolled with raw `dgram` and **zero new dependencies**, deliberately
mirroring the existing DLNA SSDP module (`src/dlna/ssdp.js`):

- Same multi-NIC pattern — joins the multicast group on every non-internal IPv4
  interface and fans announcements out across all of them.
- Same `winston` `[mdns]` log tags and **non-fatal failure handling**: a
  bind/socket error disables discovery and logs a warning; it never takes the
  server down on the boot path.
- Coexists with an OS mDNS responder (Bonjour/Avahi) already on UDP 5353 via
  `SO_REUSEADDR` (plus `SO_REUSEPORT` off Windows).

### TXT records published
`name` (friendly name, defaults to hostname), `id` (stable instance id),
`v` (version), `scheme`, `port`, `path`, `api`, `auth` (advertised auth modes),
and `pub` (optional public URL — reuses the configured relay URL if present, so
a roaming client can reach home).

## Changes

- **`src/discovery/mdns.js`** — the advertiser. Wire-format builders and the
  query parser are factored as pure exports so they're unit-testable without
  binding a socket.
- **`src/state/config.js`** — new `discovery.mdns { enabled, name, instanceId }`.
  `instanceId` is auto-generated and persisted on first boot using the exact
  same generate-and-persist pattern as `dlna.uuid`.
- **`src/server.js`** — start the advertiser after `listen` (next to the DLNA
  startup), stop it in `reboot()`.
- **`test/mdns.test.mjs`** — pure wire-format + query-matching unit tests
  (8 cases), wired into `npm test` and a new `npm run test:discovery`.

## Config

Opt-out via config; on by default:

```json
{ "discovery": { "mdns": { "enabled": true, "name": "" } } }
```

Empty `name` → derives from `os.hostname()`.

## Verification

- `npm run test:discovery` → 8/8 pass.
- `npx eslint` clean on all touched/new files. (Unrelated pre-existing
  `no-empty` in `reboot()`'s `closeAllConnections` catch left untouched.)
- Live boot smoke test on a real host: bound 5353, advertised on two interfaces
  (LAN + Docker bridge), answered a real multicast `_mstream._tcp` query, and
  persisted the instance id across a config re-read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "bundle:bun": "bun scripts/build-bun.mjs --bundle",
     "docs:api": "redocly build-docs docs/openapi.yaml -o docs/api.html",
     "docs:api:validate": "redocly lint docs/openapi.yaml",
+    "test:discovery": "node --test --test-reporter=spec test/mdns.test.mjs",
     "pretest": "node test/helpers/ensure-prereqs.mjs",
     "test": "node --test --test-reporter=spec \"test/**/*.test.mjs\"",
     "test:unit": "node --test --test-reporter=spec \"test/unit/**/*.test.mjs\"",
--- a/src/discovery/mdns.js
+++ b/src/discovery/mdns.js
@@ -1,11 +1,11 @@
 import dgram from 'node:dgram';
 import os from 'node:os';
-import { createRequire } from 'node:module';
 import winston from 'winston';
 import * as config from '../state/config.js';
 
-const require = createRequire(import.meta.url);
-const packageJson = require('../../package.json');
+// Static JSON import (not createRequire) so the Bun single-binary bundle can
+// resolve it — a runtime require() isn't bundled and dies inside bunfs.
+import packageJson from '../../package.json' with { type: 'json' };
 
 // ── mDNS / DNS-SD advertiser ──────────────────────────────────────────────────
 //
--- a/src/discovery/mdns.js
+++ b/src/discovery/mdns.js
@@ -1,0 +1,354 @@
+import dgram from 'node:dgram';
+import os from 'node:os';
+import { createRequire } from 'node:module';
+import winston from 'winston';
+import * as config from '../state/config.js';
+
+const require = createRequire(import.meta.url);
+const packageJson = require('../../package.json');
+
+// ── mDNS / DNS-SD advertiser ──────────────────────────────────────────────────
+//
+// Advertises the mStream API as a `_mstream._tcp` service over multicast DNS so
+// LAN clients (e.g. the portable mStream player) can find the server with zero
+// configuration — no IP typing. This is the discovery half of the player's
+// pairing story; it carries metadata only and touches no routes or auth.
+//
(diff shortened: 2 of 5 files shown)


IrosTheBeggar:
Rebased onto current master (`237a083b`, post-#722). Three conflict splices, all keep-both: `src/server.js` (iroh/discoveryP2p boot blocks + mdns start), `src/state/config.js` (storage-dir mkdir loop + mdns instanceId persist — reshaped by #722's config changes), `package.json` (master's test script is now glob-based and already picks up `test/mdns.test.mjs`, so only the `test:discovery` alias was added). A diff-of-diffs against the original branch confirms every code file carries byte-identical changes. Full test suite passes locally (exit 0), including #722's `discovery-export` tests 10/10, mdns 10/10. The stacked #721 branch was replayed onto this head in the same pass. Reminder from the stack rules: merge this first, then retarget #721 to master before merging it.

IrosTheBeggar:
CI caught a real bundling bug: `mdns.js` read the version via `createRequire('../../package.json')`, which Bun's static bundler can't see — all three natively-runnable bundle jobs (linux-x64, win-x64, darwin-arm64) died at import inside bunfs with `Cannot find module '../../package.json'` (the two green jobs are cross-arch bundles that skip the boot smoke). Fixed by switching to the same static JSON import the rest of the codebase uses (`server.js`, `dlna/ssdp.js`). Verified locally: `bundle:bun` builds, the .exe boots, serves `/api/v1/ping` 200, and advertises `_mstream._tcp` with the correct version TXT. #721's branch replayed on top again.

IrosTheBeggar:
Second CI round caught another stack-placement bug, reproduced and fixed: darwin-arm64's smoke test crashed the server at boot with `mdns: label exceeds 63 bytes` — that runner's hostname (the default mDNS instance name) is longer than a DNS label allows, and the encoder's throw escaped through the bind callback and killed the process. The fix already existed (`fix(mdns): clamp instance name…`) but was stranded in #721, one PR too late — the parent crashed without it. Moved that commit into this PR where it belongs. Reproduced locally both ways: pre-clamp boot with an 80-char name dies (ping 000), post-clamp the same config boots, serves 200, and advertises with the clamped label. The clamp commit also wraps `announce()` so no future encoding error can escape into the boot path or the re-announce timer — making the call-site comment ('never throws on the boot path') actually true. This PR is now 3 commits: mdns feature + bun-bundle import fix + clamp. #721 is down to its single advertise commit.

PR #679, opened 2026-06-27
ci: run tests on the `full-ci` label + fix iroh handshake rejection

IrosTheBeggar:
## Why

The previous trigger (`merge_group`, from #659) can never fire: **GitHub's merge queue is organization-only** — it's unavailable for repositories owned by a personal account, which is why the "Require merge queue" option is absent in the UI and the REST API rejected the `merge_queue` ruleset rule. ([docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue))

## What

Run the 9-job matrix only when a PR is **labelled `full-ci`** (plus a manual `workflow_dispatch`), not on every commit:

```yaml
on:
  pull_request:
    types: [labeled]
  workflow_dispatch:
# jobs gated: if: github.event_name == 'workflow_dispatch' || github.event.label.name == 'full-ci'
```

- Add the `full-ci` label when a PR is ready to merge → one full run. Re-add it (or dispatch) to re-run.
- Other label events are no-ops (jobs skip → ~no minutes).
- The `tests passed` aggregator still collapses the 9 legs into one stable required check.

This keeps the resource savings you were after (one run per PR near merge, not per commit) without the merge queue.

## To enforce (one-time, in the UI — required checks *are* available on your account)

Settings → Branches (or Rules) → protect `master` → **Require status checks to pass** → add **`tests passed`**. A PR then can't merge until it's been labelled `full-ci` and the run is green.

## Verified

Dispatched this branch via `workflow_dispatch` → all 9 jobs run (none skipped), confirming the gate is correct. (The label path can't be exercised until this is on master; it's the standard `github.event.label.name` check.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,17 +1,19 @@
 name: Test
 
 on:
-  # The heavy 9-job matrix runs only in the merge queue — i.e. once on the
-  # exact commit about to land on master, not on every PR push. Enable the
-  # merge queue on master and mark the "tests passed" check required (branch
-  # protection); merging a PR then enqueues it, which is what triggers this.
-  merge_group:
+  # The heavy 9-job matrix runs only when a PR is labelled `full-ci` — i.e. once
+  # you mark it ready to merge — not on every push. (Merge queue would be the
+  # native "before merge" trigger, but it's org-only and this repo is on a
+  # personal account.) Require the "tests passed" check in branch protection so
+  # a PR can't merge until it's been labelled `full-ci` and the run is green.
+  pull_request:
+    types: [labeled]
   # Manual escape hatch — run on demand from the Actions tab or
   # `gh workflow run test.yml --ref <branch>`.
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
Bundled a second commit: **fix(iroh): surface a rejected handshake as a clean error on all platforms**.

The `full-ci` run here surfaced a pre-existing failure on **ubuntu shard 3/3** — `test/integration/iroh-handshake.test.mjs` "wrong secret is rejected": the client's `connectTunnel` read threw the raw QUIC `ConnectionLost(ApplicationClosed(... reason: "unauthorized"))` instead of the clean `tunnel handshake rejected` error (a timing race — the server rejects by *closing* the connection; on Linux the close beat the non-"OK" check). Wrapped the handshake read so an `unauthorized` close maps to the same clean rejection.

It's bundled here (not a standalone PR) deliberately: with the required `tests passed` check enabled but the label trigger not yet on master, only a PR whose head carries the trigger — this one — can produce that check, so a separate iroh PR would be blocked by the gate. Merging this lands the trigger **and** the fix together.

PR #687, opened 2026-07-01
feat(discovery): separate discovery.db + admin export snapshot (P0)

IrosTheBeggar:
## What

P0 of the music-discovery dataset ([discovery-network plan](https://github.com/IrosTheBeggar/mStream)): a **second SQLite file, `discovery.db`**, deliberately isolated from `mstream.db`, plus an **admin export surface** that produces a cleaned, self-contained, shareable snapshot. No scanner/worker yet — this is the store + export scaffolding the CLAP embedding worker (next phase) will write into.

## Why a separate DB

- The whole point of the dataset is to be shared later — a standalone file is the exportable unit.
- Quarantines opt-in, privacy-sensitive data behind one file the operator can delete wholesale.
- Its schema evolves on its own version line, independent of mstream.db''s migration chain.
- Only coupling to mstream.db is the canonical `audio_hash` — and that never leaves the machine.

## Pieces

- **`src/db/discovery-db.js`** — open/migrate (WAL, busy_timeout, own SCHEMA_VERSION=1). Tables: `discovery_tracks` (internal: audio_hash PK / source_mtime / monotonic `updated_at` rowversion; exported: `export_id`, MBIDs, artist/title/duration, model pin, embedding BLOB, tier-1 filter cols), `discovery_lookups` (worker negative-cache, never exported), `discovery_meta` (self-description + row_seq counter + secret export salt). `upsertDiscoveryTrack()` is the single write path the future worker and tests share.
- **`src/db/discovery-export.js`** — snapshot builder: ATTACH fresh file + `INSERT…SELECT` **explicit allowlist** (`VACUUM INTO` can''t filter), one consistent transaction, deterministic ordering, self-describing `meta` (embedding model/dim/dtype/endianness/normalization), `manifest.json` (sha256, row count, sizes) + README. `export_id` is `mbid:<recording-mbid>` when known, else a **salted** opaque id — the raw local hash and the salt never travel.
- **Admin API** — `POST /db/params/collect-discovery-data` (toggle; ON creates the DB, OFF keeps data), `POST /db/discovery-export` (build, returns manifest; 404 before first enable, 409 concurrent), `GET …/manifest`, `GET …/download` (Range-capable → resumable; works through the iroh pairing tunnel as-is).
- **Config** — `scanOptions.collectDiscoveryData`, default **OFF** (opt-in by design). Boot initializes the DB only when enabled — non-fatally, unlike initDB().
- **Admin UI** — Database-panel row: live toggle (confirm dialog) + Export link (builds then downloads).

## Bug found during browser verification

`res.download`''s default `dotfiles:''ignore''` 404s any file whose **path contains a dot-segment** — e.g. a dbDirectory under `~/.config/mstream`. Fixed with `dotfiles:''allow''` on the download route (verified in a live server whose dbDirectory sits under a dot-directory).

## Testing

- 12 in-process tests (`test/db/discovery-db.test.mjs`): schema bootstrap, meta seeding, salt stability across reopen, export-id semantics, upsert/rowversion, snapshot allowlist + sha256 integrity + rebuild.
- 9 integration tests (`test/integration/discovery-export.test.mjs`): opt-in flow, 404s before enable, config persistence, empty + seeded round-trip (byte-for-byte sha256), ordering, Range/206.
- Full suite: **1451/1452** — the 1 failure (`public-mode-privacy` wrapped play-start) **reproduces on clean master**, unrelated to this change.
- Zero new lint issues (flagged ones pre-exist on master).
- Live-verified in the browser: toggle → discovery.db created + config.json persisted → export → download (attachment, valid SQLite bytes, Range 206).

## Explicitly deferred (per plan)

Embedding worker (CLAP, P1), AcoustID identity fill (P2), network-peer/public distribution (per-server keys, aggregation), incremental export (the `updated_at` cursor groundwork is in).

### DIFF
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -11,6 +11,8 @@ import * as dbQueue from '../db/task-queue.js';
 import * as imageCompress from '../db/image-compress-manager.js';
 import * as transcode from './transcode.js';
 import * as db from '../db/manager.js';
+import * as discoveryDb from '../db/discovery-db.js';
+import * as discoveryExport from '../db/discovery-export.js';
 import * as logger from '../logger.js';
 import { joiValidate } from '../util/validation.js';
 import { isAdminAllowed } from '../util/admin-network.js';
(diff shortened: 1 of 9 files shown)


IrosTheBeggar:
The failing check (ubuntu 3/3: public-mode-privacy wrapped play-start) is a pre-existing master bug that only manifests on the first day of a period — see #688 for the diagnosis + fix. Once #688 merges, update this branch / rerun checks and it goes green. Nothing in this PR touches that code path.

PR #689, opened 2026-07-02
feat(discovery): embedding worker with swappable model registry (P1)

IrosTheBeggar:
## What

P1 of the discovery dataset (follows [#687](https://github.com/IrosTheBeggar/mStream/pull/687)): the **embedding worker** — a 5th post-scan enrichment pass that fills the separate `discovery.db` with one audio embedding per canonical track, plus the share-safe metadata snapshot (artist/title/duration, and BPM/key when the library already knows them). Same operational contract as the other enrichment workers (serial task slot, `maxPerRun` + wall-clock budget, hitCap re-enqueue, Bun worker rules, orphan sweep, schema guard).

## Swappable models (a design requirement, not an afterthought)

The embedding engine is a **registry** ([discovery-features-lib.js](src/db/discovery-features-lib.js)): stable key → weights repo, version, dim, sample rate, windowing. Selected by `scanOptions.discoveryModel`, validated against the registry everywhere. The project may yet trade the current Apache-licensed CLAP for a better-quality non-commercial model (MTG Discogs-EffNet / MERT) if the commercial angle is dropped — with this design that's a config change:

- every row is pinned `(model_id, model_version)`;
- the worker re-embeds any row whose pin differs from the active model → **changing the config migrates the dataset in place** over the next passes;
- success writes **no** cooldown ledger row (success is visible in `discovery_tracks` itself; `discovery_lookups` records failures only) — which is exactly what lets a swap start re-embedding immediately instead of fighting stale success-cooldowns;
- a deterministic, dependency-free **`test-fake`** registry entry is both the no-download test path and living proof the swap mechanism works.

## Default engine

`clap-music-and-speech` → `Xenova/larger_clap_music_and_speech` (base weights `laion/larger_clap_music_and_speech`, **Apache-2.0**, verified) via `@huggingface/transformers` on onnxruntime-node. 512-d, 48 kHz, 3×10 s windows mean-pooled, L2-normalized — validated in a spike on real music (strong artist clustering, sine-tone control ≈ 0.0, ~4–6.5 s/track single-threaded fp32). Note: the originally planned `Xenova/larger_clap_music` **no longer exists on HF** (401) — a reminder that pinned weights should eventually be mirrored under project control; the registry treats the repo as a download source, not the identity.

- Weights download on first use into **`storage.modelCacheDirectory`** (new, default `appRoot/model-cache`) — deliberately outside `node_modules`.
- `@huggingface/transformers` is an **optionalDependency** (like `@number0/iroh`): a failed native install degrades the discovery feature with a clean worker error, never the music server.

## Wiring

- task-queue: `'discovery'` enrichment kind, chained **last** on scan drain (most expensive per track). The enqueue pre-check briefly ATTACHes the library DB to the discovery handle so quiet drains don't fork no-op children.
- Admin: the collect-discovery-data toggle now also enqueues a pass on enable; new `discovery-per-run` and `discovery-model` routes; Database-panel rows (model display, per-run modal).
- Config: `scanOptions.discoveryModel`, `scanOptions.discoveryPerRun` (default 50 — each track costs seconds of CPU; the worker also self-limits to a 300 s budget per pass and re-enqueues).

## Testing

- **11 forked-worker tests** (`test/integration/discovery-backfill.test.mjs`) on the `test-fake` model — the worker is spawned exactly as task-queue forks it, real ffmpeg decode, no network: embed + pin + snapshot fields, canonical-hash dedupe, idempotence, **model-swap re-embed**, duration/genre gating, error cooldown (backdated rows — no cooldown-0 clock races), success-clears-error, per-run cap + hitCap resume, library schema guard (exit 3, no writes), orphan sweep, invalid-model rejection.
- All P0 discovery suites + admin-params suite still green (38 tests).
- Full `npm test` run locally (see PR checks for the matrix).
- Live browser verification: boot with the new wiring, correct *no-op* on the fixture library (2 s tones are outside the duration window — the pre-check declined to fork), UI rows render, per-run modal round-trips to config.json, model route rejects unknown keys (400) and swaps live.
- Zero new lint issues (task-queue's 6 `no-empty` flags pre-exist on master at shifted lines).

## Deliberately deferred

Text-side CLAP (query embeddings) until the similarity/query surface lands; AcoustID identity fill (P2); genre/mood tag columns (P3); weights mirroring infra; admin UI model *picker* (API + config exist; UI shows the active model read-only until there's a second real model to pick).

## Heads-up for review context

The commercial aspect of the discovery feature has since been dropped, which unlocks MTG's **Discogs-EffNet** (CC-BY-NC-SA) — a follow-up PR will add it as a registry engine and flip the default (spiked successfully: official 18 MB ONNX via onnxruntime-node + the already-shipped essentia mel front-end; excellent style/similarity results on real music). The registry + per-row pin + re-embed mechanism in THIS PR is exactly the machinery that makes that a config-default change. CLAP stays selectable.

Note on checks: the local full `npm test` run was still in progress at PR time (targeted suites — the 11 new worker tests, all P0 discovery suites, admin params — are green); add the `full-ci` label to run the 9-job matrix here.


IrosTheBeggar:
Local full test suite result (was noted as in-progress in the PR body): **1475/1475 pass, 0 failures** on the combined branch state (this PR + #691). Bundle workflow is green on this branch's exact HEAD via workflow_dispatch (run 28564279879) after the --external fix; #691's PR checks are green on all 7 platform legs. Merge order: this PR first, then #691.

PR #694, opened 2026-07-02
feat(discovery): auto-fetch + cross-library similarity search (N4a)

IrosTheBeggar:
## What

N4a of the discovery network (follows #690/#692): **the payoff layer**. Servers now automatically download the most useful peers' snapshot DBs, and a user-facing API answers the question this whole effort exists for:

> *"What's on the network that sounds like this track — that I don't already have?"*

`POST /api/v1/discovery/similar {filepath, newArtistsOnly?}` → ranked, metadata-only results with peer attribution. With `newArtistsOnly`, every artist the local library already knows is dropped — the introduce-me-to-someone-new mode.

A privacy property worth calling out: **queries never leave the machine.** Similarity runs against local copies of peer snapshots, so peers only ever observe that you fetched their DB once — never what you search for.

## Pieces

**Peer-DB shelf** (`src/state/discovery-peer-dbs.js`) — the local collection of fetched snapshots:
- Auto-fetch reconcile loop (announcement-debounced + periodic): pulls the top-N catalog peers — **online-now first, then library size**. That's an honest proxy; real popularity ranking (seeder counts) needs N3's provider tracking and slots into the same sort later.
- Refresh-on-staleness: a peer's announced monotonic `snapshotSeq` moving past our copy triggers a re-fetch — the P0 counter doing exactly what it was designed for.
- Guardrails: `autoFetchCount` (default 3), `maxPeerDbStorageMb` (default 500), and `blockedPeers` (config) which also hides a peer from the catalog entirely — the v1 abuse lever.
- Every snapshot is **format-validated before first query** (it's an untrusted peer's SQLite file) and opened read-only.

**Similarity search** (`src/api/discovery.js`):
- Cosine = plain dot product (vectors are L2-normalized at write time), brute-force over every model-compatible peer DB — 50k peer tracks ≈ milliseconds.
- **Novelty filter chain** (no single identity signal covers every library): owned recording-MBID → normalized artist+title → near-duplicate embedding (≥0.99 vs the query catches untagged re-encodes). `newArtistsOnly` adds the artist-level drop.
- **Per-row `model_id` gating**: only vectors in the query track's model space are compared, so a network mid model-migration degrades to fewer results, never to garbage rankings.

**Admin surface**: catalog route now carries online/fetched/stale/compatible flags + most-useful-first sort + storage usage; manual fetch/remove routes; and a **Discovery Network card** in the admin UI — endpoint id, the shareable bootstrap ticket, a peer table with download/update/remove, and a storage gauge. Verified live in a browser: renders with real data, zero console errors.

## A real protocol bug the new tests caught

Gossip (Plumtree) **deduplicates identical message bytes** — so the sidecar's byte-identical 15-second re-announcements were being silently dropped after the first delivery. Consequences: late joiners would never hear existing peers, and last-heard timestamps (the "online" indicator) would never refresh. Announcements now carry an **unsigned heartbeat counter** so every re-broadcast is unique bytes. Replaying one buys an attacker nothing: the payload is still signed, `snapshotSeq` still can't roll back, and the per-origin rate limit caps processing. Relatedly, a **higher** `snapshotSeq` now bypasses the same-origin flood-guard gap — a seq bump right after a heartbeat is news, not a flood.

## Tests

18/18 in the discovery suite. The similarity engine is tested **fully synthetically** — hand-built snapshot files + directly-seeded embeddings, exact cosine assertions, no sidecar/network needed, so it runs everywhere including CI. The auto-fetch loop runs live against a raw peer sidecar (download-on-announce + refresh-on-seq-bump; needs a sidecar binary, auto-skips otherwise). Full suite: **1474/1474**.

## Reviewer notes

- `MSTREAM_TEST_DISCOVERY_DEBOUNCE_MS` is a test-only env knob (waiting 30 real seconds per assertion is unkind); production never sets it.
- Meaningful results need embeddings — #689 (the embedding worker) populates them; this PR works against whatever model the registry pins (per-row gating), so the EffNet default flip composes cleanly.
- Deliberate scope cuts: player-UI discovery panel is N4b; seeder-count ranking + multi-source swarming are N3; runtime blocklist editing (vs config-file) can ride along with N4b.

### DIFF
--- a/p2p-sidecar/src/main.rs
+++ b/p2p-sidecar/src/main.rs
@@ -122,12 +122,22 @@ struct AnnouncePayload {
     name: String,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Clone)]
 struct Announcement {
     v: u32,
     from: String,
     payload: AnnouncePayload,
     sig: String,
+    // Heartbeat counter, bumped on every re-broadcast and deliberately NOT
+    // part of the signature. Gossip (Plumtree) deduplicates by message
+    // content, so byte-identical re-broadcasts are silently dropped — late
+    // joiners would never hear us and receivers' last-heard timestamps (the
+    // "online" indicator) would never refresh. The counter makes each
+    // heartbeat unique bytes. Replaying with a fresh counter buys an
+    // attacker nothing: the payload is still signed, snapshotSeq still
+    // can't roll back, and the per-origin rate limit caps processing.
+    #[serde(default)]
+    n: u64,
 }
 
 struct Node {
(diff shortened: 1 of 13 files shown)


IrosTheBeggar:
Rebased onto master (through #698) and reworked the API namespace per discussion:

- **`/api/v1/discovery/local/*`** — the merged #695 library-similarity routes (`similar`, `similar-artists`), moved from `/api/v1/discovery/similar*`. Webapp Discover panel + tests updated; the Auto-DJ sonic picker consumes exported functions, not HTTP, so it's untouched.
- **`/api/v1/discovery/p2p/*`** — this PR's routes (`similar`, `peer-dbs`), now in their own `src/api/discovery-p2p.js`. Also adopted the local routes' conventions: `filePath` param and the shared `resolveSeedTrack()` (uniform 404s + vpath probe logging).

The two features had collided on `POST /discovery/similar` (parallel development). Nothing is released yet, so the local-route rename ships without back-compat shims.

Verification after rebase + rework: both discovery suites 33/33, full suite **1546/1546**.

PR #702, opened 2026-07-04
feat(discovery): first community seed — the network opens (PR-2)

IrosTheBeggar:
## What

Adds **seed-au-1** (DigitalOcean, Australia) to `seeds/discovery-seeds.json` and the baked `DEFAULT_SEEDS` fallback. **The moment this merges, any mStream server enabling `discoveryP2p` joins the public discovery network** — no ticket exchange, no introductions.

## Why Australia

Deliberate worst-case placement: if bootstrap works antipodally, it works anywhere. Measured live from the other hemisphere:

- Mesh join through the seed: **1.7 s** per node
- After introduction, the two joined peers connected **directly** (`neighbors=2`, announcement propagation 0.2 s) — the "seeds are training wheels, not hubs" design promise, observed in the wild

Also verified on the **exact production path**: a server on pure defaults (real GitHub list URL — which 404s until this merges) fell back to the baked list, joined via the AU seed, and heard a stranger's announcement end to end.

## The important half of the diff: test hermeticity

With `DEFAULT_SEEDS` no longer empty, the test helper's dead-URL guard alone would have let **every CI suite fall back to the baked seed and join the real network** — broadcasting fake test announcements into real users' catalogs. The helper now also forces `useCommunitySeeds: false`; the seed-mechanics suite re-enables it explicitly against its local stub server. Discovery suite 25/25 with zero real-network contact.

## Ops notes

- The seed's `identity.key` on the droplet **is** this entry — please back it up (restoring it restores the seed on any host).
- Rotating/adding seeds from here on = editing `seeds/discovery-seeds.json` on master (servers refresh ~daily); the baked list just needs to stay roughly in sync at release time.
- A second seed in another region remains the honest next step for redundancy — same one-liner deploy, one more entry here.

### DIFF
--- a/seeds/discovery-seeds.json
+++ b/seeds/discovery-seeds.json
@@ -7,6 +7,11 @@
       "name": "seed-au-1",
       "endpointId": "c961437a8ff60617d7b36b5bca0e866e9521b5194e8068de08a731631418b00b",
       "ticket": "endpointadewcq32r73amf6xwnvvxsqoqzxjkinvdfhia2g6bcttcyyudcyawayaenuhi5dqom5c6l3bobztcljrfzzgk3dbpexg4mbonfzg62bonruw42zof4aqblaraabjbwqdaeancjs65kinuay"
+    },
+    {
+      "name": "seed-eu-1",
+      "endpointId": "d7d3f4501a5a70e17ea93106f303b77714d5625e9849aa4ebaff6d0c5d4b2260",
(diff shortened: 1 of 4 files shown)


IrosTheBeggar:
Second seed added: **seed-eu-1** (DigitalOcean EU Central — region decoded from the ticket's `euc1` relay). Live-verified: the EU seed relays announcements between fresh nodes (0.8–0.9 s joins), and a node bootstrapping with **both** tickets meshed with both seeds (`neighbors=2` in 5.2 s). With two independent regions in the list + N3's swarm failover on deck, bootstrap has no single point of failure. Don't forget to back up `/data/identity.key` on the new droplet as well.

PR #703, opened 2026-07-04
feat(discovery): seeder beacons + swarm fetch + blob GC (N3)

IrosTheBeggar:
## What

The distribution layer of the discovery network grows up:

1. **Snapshots survive their authors.** Every node broadcasts a **signed "holds" beacon** — the snapshot hashes it currently holds (its own + fetched shelf). Fetch now offers the sidecar's downloader *every* known provider with shuffled failover, so a snapshot stays available while **any** holder is online. The integration test kills the author and pulls the bytes from a surviving holder.
2. **Popularity becomes observable.** Aggregated beacons yield live **seeder counts** per snapshot — the honest signal a decentralized network can actually measure (download counts were dropped in the original design as unverifiable/gameable). The catalog route + admin UI show Seeders and sort by it first, replacing the online+size proxy from N4a.
3. **The blob store stops leaking.** The sidecar store now runs GC (15 min interval) and a new `forget` command unpins superseded blobs — ours on re-publish, peers' on refresh/remove. This leak has existed since N1: every snapshot update pinned the old bytes forever.

## Why beacons are signed

Who-has-what is *exactly* the claim an attacker inflates to fake popularity. Beacons get the same origin-signature treatment as announcements, and the verification boundary got real teeth this round: `verify_wire()` extracted as a pure function with **7 cargo adversarial tests** — forged signatures, cross-identity impersonation, tampered payloads, padded hold-sets, oversize, separator injection — now running in the binaries CI workflow's native leg. (This closes the "forged-signature adversarial test" gap flagged since N2.)

An honest limit, documented in code: signatures prove a beacon's *origin*, not that the origin truly has the bytes. A liar inflates counts only for its own claimed holds, gets caught the first time anyone fetches from it (BLAKE3 verification fails → downloader fails over), and the blocklist handles repeat offenders.

## Tests

Discovery suite **26/26** (new: beacon → seeder-count aggregation on the server; the server *becoming* a holder as observed by a third node hearing its beacon; author-offline swarm fetch with byte-exact verification). Cargo **7/7**. Full JS suite green locally.

## Merge order

**#702 (seed list) first, please** — both PRs touch the discovery test file; I'll rebase this one after it lands if GitHub reports conflicts.

## After merge

The binaries workflow rebuilds the sidecar (protocol additions are backward-tolerant: old sidecars ignore unknown `kind` messages with a log line, new ones interop with old announcements). Remaining N-phase backlog after this: runtime blocklist editing UI, and the deliberately-deferred seed-list signing.

### DIFF
--- a/.github/workflows/build-p2p-sidecar.yml
+++ b/.github/workflows/build-p2p-sidecar.yml
@@ -85,6 +85,14 @@ jobs:
           curl -fsSL https://github.com/cross-rs/cross/releases/latest/download/cross-x86_64-unknown-linux-gnu.tar.gz \
             | tar xz -C "$HOME/.cargo/bin"
 
+      # The wire-verification unit tests (forged signatures, impersonation,
+      # tampered holds) run on the native leg — one platform suffices for
+      # pure logic, and the cross legs can't execute their test binaries.
+      - name: Unit tests
+        if: matrix.self_test
+        run: cargo test --manifest-path p2p-sidecar/Cargo.toml
+        shell: bash
+
       - name: Build
         run: |
           if [ "${{ matrix.use_cross }}" = "true" ]; then
--- a/p2p-sidecar/src/main.rs
+++ b/p2p-sidecar/src/main.rs
@@ -60,7 +60,11 @@ use iroh::{
     address_lookup::memory::MemoryLookup, endpoint::presets, protocol::Router, Endpoint,
     EndpointId, PublicKey, SecretKey, Signature,
 };
-use iroh_blobs::{store::fs::FsStore, ticket::BlobTicket, BlobsProtocol, Hash};
+use iroh_blobs::{
+    store::fs::{options::Options as StoreOptions, FsStore},
+    store::GcConfig,
+    ticket::BlobTicket, BlobsProtocol, Hash,
+};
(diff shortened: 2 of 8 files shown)


IrosTheBeggar:
CI failures diagnosed and fixed (commit just pushed):

**Root cause** — the protocol-PR chicken-and-egg: full-ci tests against the prebuilt sidecar from *master* (`bin/p2p-sidecar/`), which predates this PR's `setHolds`/`providers` protocol → `unknown command: setHolds` on all three platforms' shard 2. Local runs never hit it because a fresh cargo build shadows the prebuilt. Notably, this is the *first* PR where CI has executed the sidecar suites at all — until yesterday's binaries commit, they skipped for lack of a binary.

**Fix** — capability probe in the N3 suite's `before()`; old binary → `t.skip()` with an explanatory reason. Verified both ways locally: stale-prebuilt simulation = 25 pass / 1 skip / 0 fail (what CI will now do); fresh build = 26/26, no skips. Once this merges and the binaries workflow re-commits, CI runs the N3 suite for real on every subsequent PR.

This pattern will apply to any future sidecar-protocol PR — worth remembering rather than re-diagnosing.

PR #707, opened 2026-07-06
feat(scanner): ingest MusicBrainz/AcoustID external IDs from file tags (external-ID Phase 1)

IrosTheBeggar:
## What

External-ID tagging **Phase 1**: mStream now reads the identifiers that Picard/beets-tagged files already carry, instead of dropping them.

- **Schema V55**: `tracks.mbz_recording_id` / `mbz_release_track_id` / `acoustid_id` / `isrc` / `mbz_id_source`, and `albums.mbz_release_group_id`. `rescanRequired` — upgrades force-rescan to populate already-scanned libraries.
- **Both scanners** (rust-parser + JS fallback) ingest the MusicBrainz recording / release-track / release-group MBIDs, AcoustID id, and ISRC from embedded tags — including the historical naming quirk where `MUSICBRAINZ_TRACKID` actually carries the *recording* MBID. Album rows fill-NULL their release / release-group MBIDs first-writer-wins (same convention as album art). The long-existing `albums.mbz_album_id` column finally gets populated too.
- **Provenance**: `mbz_id_source = 'tag'` (mirrors `bpm_source`); `'acoustid'` is reserved for the Phase-2 fingerprint worker that will *derive* these ids for untagged files (Chromaprint → AcoustID → MBID).
- **ListenBrainz** submissions now carry `recording_mbid` when known — pins the exact recording instead of relying on artist+title matching.
- **Metadata API** exposes the new fields on track metadata.

## Why now

This is the foundation for the AcoustID fingerprint worker (external-ID Phase 2) and for upgrading discovery-network `export_id`s from `anon:` salted hashes to portable `mbid:` identifiers — identity-grade cross-library matching.

## CI note (capability gate)

The scanner-parity test's rust leg probes the binary in `before()` and **skips with a reason on a rust-parser that predates the ingestion** — full-ci tests against master's prebuilt binaries until the post-merge rebuild (the #703 rule). Verified both ways locally: fresh build → 2/2; build hidden (prebuilt simulation) → 1 pass + 1 clean skip. Source-only for `bin/` — the binaries workflow rebuilds on merge.

## Tests

- New: `test/scanner/scanner-external-ids.test.mjs` — both-engine parity over a generated tagged library (full MusicBrainz set + ISRC; recording-id-only track proving provenance and album fill-NULL convergence; untagged track proving all columns stay NULL).
- db suite 155/155 (schema-version assertions are all relative — V55 needed no renumbering, master was still V54 and none of the touched files drifted), scanner suite 110/110, full suite 1563 pass / 0 fail (4 cancellations in an unrelated subsonic suite are the known Windows parallel-boot flake; 6/6 in isolation).

### DIFF
--- a/rust-parser/src/main.rs
+++ b/rust-parser/src/main.rs
@@ -227,6 +227,17 @@ struct ExtractedTrack {
     // marker is present. See detect_source_from_tag().
     source: Option<String>,
 
+    // V55: external-service IDs from embedded tags. mbz_recording_id is the
+    // stable per-file RECORDING MBID; the rest ride alongside. mbz_album_id /
+    // mbz_release_group_id are written to the ALBUM row (fill-NULL) by
+    // commit_track, not the tracks row. Mirrors src/db/scanner.mjs.
+    mbz_recording_id: Option<String>,
+    mbz_release_track_id: Option<String>,
+    isrc: Option<String>,
+    mbz_id_source: Option<&'static str>,
+    mbz_album_id: Option<String>,
+    mbz_release_group_id: Option<String>,
+
     aa_file: Option<String>,
     // V48: where aa_file came from — "embedded" or "folder" (or the
     // preserved pre-image value on a skip_img re-parse, which can be any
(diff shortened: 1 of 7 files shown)


IrosTheBeggar:
**Rebased onto master post-#708** (symphonia 0.6) — clean, no hunk overlap; the composed rust-parser (0.6 + external-ID ingestion) builds warning-free and passes the parity test 2/2.

**Live smoke before merge (two scenarios):**

1. *Fresh server, real music with real+synthetic tags* — three FLACs served through the full stack (scan → DB → metadata API): full MusicBrainz set ingested exactly (recording/release-track MBIDs, ISRC, provenance `tag`), recording-only file gets just the recording id, and — nice surprise — the untouched Bandcamp originals carry **real embedded ISRCs** (`QZDA…`), which the scanner picked up in the wild (verified against ffprobe on the source files; per-file values, no cross-row leakage).

2. *Upgrade path on a live V54 server* — `Applying migration v55` + forced rescan on first boot (resumable epoch), columns repopulated from tags (5/6 real tracks gained ISRCs + provenance; 0 MBIDs as expected for Bandcamp files), and the next boot is fully idempotent (`0 processed, 6 unchanged`, no rescan loop). Discovery/p2p state on the same server unaffected.

Ready to merge from my side. Post-merge: the binaries workflow rebuild un-skips the rust test leg from the next PR onward.

PR #710, opened 2026-07-06
feat(db): AcoustID identification worker (external-ID Phase 2)

IrosTheBeggar:
## What

The 7th enrichment worker, completing the AcoustID identity chain built across #707/#709: tracks with no MusicBrainz recording MBID get fingerprinted (`rust-parser --fingerprint`) and resolved via api.acoustid.org into `mbz_recording_id` + `acoustid_id` with `mbz_id_source='acoustid'`.

**The network payoff**: matched identities propagate into discovery.db — `export_id` upgrades from `anon:<salted>` to `mbid:<recording>` with a rowversion bump, the embedding untouched — and the post-run auto-publish (#704) re-announces the snapshot. Cross-library matching and the "new artists only" filter become identity-grade as libraries get identified. The embedding worker now also carries identity when it *creates* rows, so either pass ordering converges on the same state.

## Shape

- **Worker** (`src/db/acoustid-backfill.mjs`) on the established enrichment contract: serial slot, `maxPerRun`/`runBudgetSec`, per-outcome cooldowns (`acoustid_lookups`, **V56** — success writes no ledger row), hitCap re-enqueue, schema guard, orphan sweep. Network-bound specifics: 350 ms throttle (3 req/s limit), **≥0.9 score acceptance** (live spike scored real music 98%+) with the most-submissions recording winning, early abort after 3 consecutive API failures, and a startup capability probe that hard-fails on a pre-`--fingerprint` binary instead of cooldown-poisoning the library.
- **Fill-NULL only, hash-fanned**: tag-sourced IDs are never overwritten; byte-identical copies all get the ID from one lookup.
- **Config**: `analyzeAcoustid` (default **OFF** — it sends acoustic fingerprints to an external service), `acoustidPerRun`, `acoustidApiKey` (project app key as default, Picard-convention; forks can register their own), `acoustidApiUrl` (test override). Admin toggle + per-run row/modal cloned from the analyze-bpm recipe; enabling enqueues an immediate pass.

## Tests

7 integration tests: real rust fingerprinting against a **scripted local AcoustID stub** (outcome keyed on the request's `duration` field) — match fanout + discovery `export_id` upgrade with embedding untouched, tag-provenance immunity, nomatch/lowconf/undecodable ledger outcomes + cooldown gating, error retry via backdated ledger rows (never `cooldown=0`), hitCap, schema-guard exit 3. Capability-gated, though CI's prebuilt already has `--fingerprint` post-#709, so everything runs for real. db suite 155/155 (V56 trips no assertions), full suite **1580/1580**.

With this merged, external-ID Phase 2 is complete. Phase 3 (MusicBrainz ws/2 enrichment + "fix bad tags" UI) remains future work.

### DIFF
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -329,6 +329,33 @@ export function setup(mstream) {
     res.json({});
   });
 
+  // Toggle the AcoustID identification pass (chromaprint fingerprint →
+  // MusicBrainz recording MBID for tracks whose tags carry none). Default
+  // OFF — it sends acoustic fingerprints to an external service. Flipping
+  // ON enqueues an immediate pass through the same guarded path as the
+  // scan-drain trigger (config, API key, rust-parser, eligibility).
+  mstream.post("/api/v1/admin/db/params/analyze-acoustid", async (req, res) => {
+    const schema = Joi.object({
+      analyzeAcoustid: Joi.boolean().required()
+    });
+    joiValidate(schema, req.body);
+
+    await admin.editAnalyzeAcoustid(req.body.analyzeAcoustid);
+    if (req.body.analyzeAcoustid === true) { dbQueue.maybeEnqueueAcoustid(); }
+    res.json({});
+  });
+
+  // Tracks identified per AcoustID pass. Mirrors analyze-bpm-per-run.
+  mstream.post("/api/v1/admin/db/params/acoustid-per-run", async (req, res) => {
+    const schema = Joi.object({
+      acoustidPerRun: Joi.number().integer().min(1).max(10000).required()
+    });
+    joiValidate(schema, req.body);
+
+    await admin.editAcoustidPerRun(req.body.acoustidPerRun);
+    res.json({});
+  });
+
   // Toggle collection of music-discovery data (embeddings + external IDs)
   // into the separate discovery.db. Flipping it ON creates/migrates the DB
   // immediately AND enqueues an embedding pass (through the SAME guarded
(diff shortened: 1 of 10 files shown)


IrosTheBeggar:
**Live smoke complete — the worker ran against the real AcoustID API on a real server, and the whole chain fired end to end:**

1. Admin toggle → task-queue enqueue → worker forked (V56 had migrated on boot).
2. **5 of 6 library tracks identified** at the real service (all five Icarus tracks — MBIDs match the spike's known-good values exactly); `glass_house.mp3` correctly recorded as `nomatch`.
3. discovery.db: 5 `export_id`s upgraded `anon:` → `mbid:`, embeddings untouched.
4. Auto-publish rebuilt the export (row_seq 6→11) and **announced the identity-upgraded snapshot to the live seed network**.
5. The peer server auto-refreshed its held copy to seq 11 without intervention.

The `meta=recordings sources` response shape I'd only seen stubbed parses correctly live (the most-backed recording won on every match).

**Two findings, fixed in 77091552:**
- `mbz_id_source` stayed `'tag'` when the track's ISRC was tag-read but the MBID came from AcoustID — the fill now sets `'acoustid'` unconditionally (the column names the *identity* provider; the WHERE already protects tag-MBID'd rows). The stub tests couldn't catch this — it needed real Bandcamp files with embedded ISRCs.
- The worker's discovery-db import produced a `[winston]` no-transport stderr banner that got error-labeled; demoted with the other runtime chatter.

Worker suite 7/7 post-fix.

PR #712, opened 2026-07-07
feat(admin): P2P discovery UI — nav reorg, Discovery page, one-click enable

IrosTheBeggar:
> **Stacked on #711** — base is `feat/discovery-db-description`. Merge #711 first.

The P2P UI update: the discovery network becomes a first-class admin surface — its own nav entry, its own page, and a one-click runtime enable with no config-file edits and no restarts.

## 1. Sidebar reorg

| Group | Items |
|---|---|
| **Config** | Users, Directories, Quick Connect, **Discovery** (new page) |
| **Features** (new) | Subsonic API, DLNA, Torrent, Federation |
| **Server** | unchanged |

The Discovery Network card moves off the Database page onto the new deep-linkable Discovery page (`/admin/#discovery-view`). The local-analysis scan settings stay with the scan settings on Database. Fixed along the way: a latent TDZ crash where hash deep-links mounted views synchronously before tail-of-file consts initialized (previously reachable via `/admin/#db-view`).

## 2. One-click enable (`POST /api/v1/admin/discovery/p2p/enabled`)

- **Enabling forces discovery data collection on** (the network is useless without embeddings) via the exact steps the collect toggle uses — DB init + immediate embedding pass — then persists both flags and starts the runtime stack. Start failure rolls the flag back so the config file never claims a state the process didn't reach.
- **Disabling stops the stack but keeps collection on** — local Discover/similar features don't depend on the network.
- The consent moment moves from "you edited the config file" to an explicit dialog: metadata-only, never audio files, name+description public, published to the community network by default.
- `src/state/discovery-p2p-stack.js`: the server.js boot IIFE extracted into an idempotent start/stop pair shared by boot and the route — one code path, never two drifting copies (the `announceCurrentSnapshot` precedent). New stop halves: `stopAutoFetch` (listener now named so it detaches) and `stopMeshHealthWatch`.

## 3. The name slice (`POST /api/v1/admin/discovery/p2p/name`)

Same live save + re-announce contract as the description route. The UI gets a combined **identity modal** (name + description, saves only changed fields) that **auto-opens right after enabling** — so nobody sits in the public catalog as the 47th "mStream".

## Verification

- New integration suite (runtime enable/disable): force-on semantics + `collectForced` flag, both flags persisted to the config **file**, discovery.db initialized, live gates flip (catalog 403→200, ping flags) with zero restarts, name validation (trim/pipe/blank/oversize don't clobber), disable keeps collection, full stop/start cycle + double-enable idempotence. Whole p2p file: **35/35**.
- Hermeticity: the runtime path resolves seeds through the same config keys the test helper neutralizes — CI can never touch the real network.
- **Live end-to-end through the browser**: a factory-defaults server (no discovery config at all) → Enable button → consent dialog → identity modal ("Lab C (Runtime Enable) — Born in the admin UI — zero config-file edits") → forced embedding pass → auto-publish → **appeared in a peer's catalog with name + description and was auto-fetched**, alongside rum.st on the public mesh. Then verified the boot suites still pass through the extracted stack.

Also: `model-cache/` added to .gitignore (the default model download location was un-ignored — an accident waiting to happen).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,4 +79,4 @@ save/waveform-cache-smoke/
 
 # Generated API docs (built from docs/openapi.yaml via `npm run docs:api`;
 # also built+deployed by .github/workflows/publish-api-docs.yml)
-docs/api.html
+docs/api.htmlmodel-cache/
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -575,6 +575,81 @@ export function setup(mstream) {
     }
   });
 
+  // The p2p master switch, runtime edition — what the config-file flag +
+  // reboot used to be. Enabling FORCES discovery data collection on (the
+  // network is useless without embeddings; the UI's confirm dialog says so)
+  // through the exact same steps as the collect-discovery-data route, then
+  // starts the same runtime stack boot runs. Disabling stops the stack but
+  // deliberately leaves collection on — local Discover/similar features
+  // don't depend on the network. NOT behind requireP2pEnabled, obviously.
+  mstream.post("/api/v1/admin/discovery/p2p/enabled", async (req, res) => {
+    const schema = Joi.object({ enabled: Joi.boolean().required() });
+    joiValidate(schema, req.body);
+    const stack = await import('../state/discovery-p2p-stack.js');
+
+    if (req.body.enabled === false) {
+      await admin.editDiscoveryP2pEnabled(false);
+      try {
+        await stack.stopDiscoveryP2pStack();
+      } catch (err) {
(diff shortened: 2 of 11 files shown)


IrosTheBeggar:
Added per review: the Discovery page now auto-polls status+catalog every 10s while open (paused when hidden / mid-toggle, quiet on transient errors) and shows an indeterminate progress bar with a 'searching for peers' caption from Enable until the first mesh neighbor appears. Verified live: poll fires and refreshes the catalog; bar appears at 0 neighbors and retires when the mesh connects.

PR #740, opened 2026-07-11
fix(backup): low-severity cleanup + deferred decisions (audit batch 6)

IrosTheBeggar:
Batch 6 — the final code batch of the backup audit: the low-severity cleanup backlog plus the three deferred decisions (made 2026-07-10). Adversarially reviewed (20 raw findings → 12 refuted, 3 confirmed — all fixed in the follow-up commit).

## The three decisions, implemented

**FAT/DST skew: handled rsync-style.** FAT-family filesystems store mtimes in *local* time, so after every DST transition each stored stamp reads back shifted by exactly ±1h — previously the first run after a transition saw every file on a FAT/exFAT USB destination as changed and trash+recopied the entire library, twice a year, demanding library-size trash headroom. `mtimesAgree()` now also accepts deltas of exactly ±3600s within the 2s slack (the same trade `rsync --modify-window` and `robocopy /DST` have made for decades), in both the trusted branch (FAT passes the fidelity probe — that's exactly where the skew bit) and the untrusted fallback's dest-older direction.

**Retention 0: stays true hard-delete.** No behavior change — the helper text in both forms now spells out that deletions are immediate and unrecoverable, so nobody mistakes it for "trash with 0-day retention."

**Missed daily window: anacron-style catch-up.** The per-destination decision is now a pure, exported `shouldTriggerScheduled()`. A destination whose last attempt *ended before yesterday's window moment* (yesterday's local date at `daily_at_hour`) fires promptly after boot instead of silently skipping the day — so a NAS that's off overnight still gets its daily backup each morning. A freshly-created destination still waits for its first regular window. Documented, accepted consequence: on a catch-up day the regular window fires too (the second run walks an already-synced mirror, which is cheap).

## Cleanup sweep

- **Orphan-dir bookkeeping** (`trashDestEntry`): leftover `.mstream-tmp-*`/`.mstream-partial-*` *files* in a dir removed from the source are now unlinked — previously tmps were skipped (ghost dir whose `rmdir` failed silently forever) and partials were trashed as user files (never-finalised garbage in the deletion log, inflated `filesTrashed`).
- **Empty-dir flip-flop**: the matched-dir "prune if now empty" rmdir is gone. The source dir exists (it's the matched pair), so its mirror should too — the prune put every empty source dir (placeholder albums, all-excluded dirs) into a permanent create/delete oscillation. Dest dirs whose source is gone are still removed via the dest-only path.
- **Source-root `.mstream-trash`**: filtered from the walk (once-per-run stderr warning), so a tree that previously *was* a backup destination doesn't get its old trash mirrored into the new dest's bucket — where the retention sweep would silently age-and-delete data the user believed was backed up. `hasAnyFiles` mirrors the filter in lockstep (pinned by test: a trash-only source *refuses* with dest untouched, rather than passing the guard while the walk sees nothing and sweeps the dest). Deeper `.mstream-trash` dirs still mirror as ordinary content.
- **Windows MAX_PATH trash paths**: the trash target adds ~26 chars over the live path, so a file mirrorable at <260 chars could never be *trashed* — a changed file in a deep tree was never refreshed, every run. Trash renames near the limit now use a `\\?\`-prefixed target, and the manager's sweep prefixes bucket roots on win32 so it can prune what the worker writes. (Defense-in-depth: Node's internal path namespacing may already cover part of this — the prefix is correct under either reading.)
- **Reboot hygiene**: `runAfterBoot` spares the live scan's `scan_progress` row on reboot-without-exit (an admin settings change mid-scan previously blanked the progress UI for the rest of the scan); `backupManager.shutdown()` now clears the previously-untracked post-boot timeouts and is called from `reboot()` (which re-inits); all four scheduler timers are unref'd.
- **`triggerForLibrary`'s ignored argument** is now real (the reason doubles as the trigger-type filter — same behavior, now honest); **queued-runs-lost-on-restart** documented as an accepted limitation with an enqueue log breadcrumb (minimal option, per decision).

## Adversarial review round (commit 2)

- **Bookkeeping unlink had no dirent-type check** — a hand-made *directory* named `.mstream-partial-*` inside an orphan dir hit `fs.unlink` → EISDIR/EPERM every run: permanent per-run error + unremovable ghost (a regression vs master; verifiers repro'd it live). Now files-only; a prefixed directory falls through to normal recursion. Pinned.
- **Catch-up date-key granularity halved cadence** — verifier simulation showed an always-off-at-window machine backing up every *other* day (the morning catch-up's row date masked that evening's miss). `missedDailyWindow` now compares against yesterday's window *moment*, keyed on `finished_at ?? started_at` (a marathon run finishing this morning covered yesterday's window — no pointless walk right behind it). Pinned.
- **Untrusted-branch DST carve-out had zero coverage** (verifier mutation: deleting it left all 95 tests green) — new test drives it through the fail-utimes preload; mutation-pinned.
- Test hardening from split findings: the long-path test now *pins* the live path at 237 chars so only the trash target (263) crosses MAX_PATH — the earlier construction let the live path float to 255, which would have failed on exactly the strict hosts the fix targets; trash bucket names are discovered by `readdir` instead of recomputing the UTC date (midnight-rollover flake).

## Validation

13 tests in `backup-batch6.test.mjs`; 10 of 13 fail with the fixes reverted (the three passes are new-coverage-only: crash recovery + the long-path pair on long-paths-enabled hosts). Full backup + task-queue suites 56/56, backup-api 18/18, lint at master baseline. Rig smoke on the big server: see comment below.

### DIFF
--- a/src/backup/manager.js
+++ b/src/backup/manager.js
@@ -271,21 +271,35 @@ export function scheduledWindowServed(lastRun, dailyAtHour, now) {
 
 const DAY_MS = 24 * 60 * 60 * 1000;
 
-// Anacron-style catch-up: was a whole scheduled window MISSED — i.e. is
-// the most recent attempt older than yesterday's local date? A server
-// that was off during last night's window (NAS shut down overnight,
-// power cut) previously skipped that day silently and waited for the
-// NEXT night — a machine that's always off during its window never
-// backed up at all. Date-key comparison is lexicographic ('YYYY-MM-DD').
+// Anacron-style catch-up: was YESTERDAY's scheduled window missed? True
+// when the destination's most recent attempt ended before yesterday's
+// window MOMENT (yesterday's local date at daily_at_hour) — if any
+// attempt had served that window or anything after it, a newer row
+// would exist. A server that's off during its nightly window previously
+// skipped the day silently and waited for the NEXT night — a machine
+// that's ALWAYS off at its window never backed up at all.
+//
+// Compared against the window moment, not yesterday's date: a morning
+// catch-up run is dated the day it ran, and a date-only comparison
+// counted it as serving that day's window — so an always-off-at-window
+// machine got backups only every OTHER day, and a midnight-straddled
+// run masked the following day's missed window.
+//
+// Keyed on finished_at ?? started_at: a marathon run that STARTED days
+// ago but finished this morning did cover everything up to its finish —
+// firing a catch-up right behind it would be a pointless third walk.
+// (While it's still running, started_at applies and this returns true —
+// harmless: the dedup gate drops the trigger while the run is active.)
+//
 // Null lastAttempt is NOT a missed window: a freshly-created daily
 // destination waits for its first regular window rather than firing the
-// moment it's saved. (DST edge: now-24h can land an hour into the wrong
-// day around transitions — worst case the catch-up fires a day early or
-// late, which the once-per-day dedup absorbs.)
-export function missedDailyWindow(lastAttempt, now) {
+// moment it's saved.
+export function missedDailyWindow(lastAttempt, dailyAtHour, now) {
   if (!lastAttempt) { return false; }
-  const yesterdayKey = localDateKey(new Date(now.getTime() - DAY_MS));
-  return sqliteUtcToLocalDateKey(lastAttempt.started_at) < yesterdayKey;
+  const y = new Date(now.getTime() - DAY_MS);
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
Rig smoke on the big server (real library, 200 files incl. >16M): **20/20 PASS** on this branch — fresh mirror, idempotent re-run, tree diff + content hashes, mid-run kill + recovery (21 deleted files restored byte-identical, no leftover partials), dest-corruption repair (old copy in trash), final idempotence.

PR #743, opened 2026-07-11
fix(backup): NFC/NFD matched-pair convergence (audit #6 — last open finding)

IrosTheBeggar:
Closes the **last open finding (#6)** of the backup audit — everything else in the 53-finding ledger is fixed, accepted-partial, or an explicit won't-fix.

## The bug

The merge-walk pairs entries by NFC-normalized key but built every dest-side path from `srcEntry.name`. A destination tree written in NFD — rsync from a Mac is the classic source — on a normalization-*sensitive* filesystem (ext4, NTFS) therefore stat'd ENOENT for every matched pair: the file was **recopied in full under the NFC name**, the NFD original was **never trashed**, and later runs paired the equal sort keys in arbitrary readdir order, producing trash+recopy churn that never converged.

## The fix (commit 1)

- `syncMatchedPair`'s early branches (link removal) act through the dest entry's **actual bytes**.
- Pairs that survive to the file/dir paths go through `convergeDestName()`: byte-different matched names — which with a pure-NFC `toKey` means a normalization variant and *nothing else*; case-different names never pair, so batch 2's deliberate case-insensitive-dest fold is untouched — get a cheap same-directory rename to the source's bytes, and all downstream operations use the converged path. Directory pairs converge then recurse, so an NFD subtree straightens out top-down in one run with **zero copies**.
- A same-inode guard lets APFS (normalization-insensitive, byte-preserving) rewrite its stored bytes through the rename; rename failure degrades to correct-but-unconverged mirroring with one stderr warning per run.

## The review round (commit 2) — a real regression caught and rooted out

The adversarial review (14 findings → 10 refuted, 4 confirmed) live-repro'd a genuine regression in the first cut: a source that *legitimately holds both spellings as distinct files* (a library merged from Mac-rsync'd and Linux-native trees), mirrored onto a dest that enumerates in a different order (legal on ext4 — per-directory hash seeds), got **cross-paired** by the positional equal-key merge. The sibling guard then routed each pair through the *other* variant's name — silently and stably **swapping the two files' contents on the mirror**. Pre-fix master was accidentally correct here, so this was a regression of commit 1, not of master.

Fixed at the root rather than patching the guard: the merge now gathers the full equal-key run on both sides and **pairs byte-equal names first**, then remaining entries positionally. Cross-pairing becomes impossible under any enumeration order, and duplicate cleanup becomes deterministic (a source spelling always claims its byte-identical dest twin; dest surplus is always the phase-1 orphan). Equal-key runs are length 1 except when variants coexist, so the common path is untouched.

Also from the review: the dest-symlink rewiring got its own mutation-pinned test, and the sibling test was reframed as the deterministic-pairing pin it actually is (the no-clobber guard is provably mid-run-race-only belt-and-braces now — its comment says so honestly).

## Tests

6 tests in `backup-nfd.test.mjs` + a new `reverse-readdir.cjs` fixture that deterministically simulates ext4's order divergence (NTFS CI hosts can never produce it naturally):

- NFD dest file: 0 copied / 0 trashed, byte-exact readdir convergence assertion, idempotent second run
- NFD directory subtree: recursive convergence, zero copies
- Pre-fix churn state (both spellings on dest): collapses to one converged entry, duplicate in trash, stable after
- Byte-exact dest twin: wins the pairing deterministically, variant trashed as orphan
- **Dual-spelling source + reversed dest enumeration: no content swap, 0 copies** (fails 1/6 against commit 1 alone)
- NFD-named dest symlink/junction: removed via its actual bytes, target untouched

Pin-checked: the original 4 tests fail 0/4 against master; the cross-pairing test fails against commit 1; the symlink test fails with its hunk reverted. Full backup + task-queue suites 62/62, backup-api 18/18, lint at baseline. Rig smoke: see comment.

### DIFF
--- a/src/backup/worker.mjs
+++ b/src/backup/worker.mjs
@@ -979,10 +979,65 @@ async function syncSrcEntry(srcEntry, srcDir, destDir, relPath) {
   emitProgress();
 }
 
-// Both source and dest have an entry with the same name. Decide what to do.
+// One stderr notice per run when a Unicode-normalization convergence
+// rename fails (see convergeDestName) — the run still operates
+// correctly through the dest's existing bytes, so this is log-only.
+let nfdRenameNoticed = false;
+function noteNfdRenameFailure(p, err) {
+  if (nfdRenameNoticed) { return; }
+  nfdRenameNoticed = true;
+  process.stderr.write(`Warning: could not converge Unicode-normalization variant name ${p}: ${err.message}; operating via the existing name\n`);
+}
+
+// A matched pair whose names are byte-DIFFERENT differs purely in
+// Unicode normalization: toKey folds nothing but NFC, so case-different
+// names never pair (the deliberate case-insensitive-dest fold in the
+// two-phase walk is untouched by this). Converge the dest entry to the
+// source's bytes with a cheap same-directory rename and return the dest
+// path all subsequent operations should use.
+//
+// Why converge instead of just reading through srcEntry.name (the old
+// behaviour): on normalization-SENSITIVE filesystems (ext4, NTFS) the
+// source-bytes path simply doesn't exist when the dest was written in
+// NFD (rsync from a Mac is the classic source) — the stat was ENOENT,
+// the file was recopied in full under the NFC name, the NFD original
+// was never trashed, and later runs paired the equal sort keys in
+// arbitrary order, churning trash+recopy forever (audit finding #6).
+//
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
Rig smoke on the big server (real library, 200 files): **20/20 PASS** on this branch — fresh mirror, idempotence, tree diff + content hashes, kill/recovery, corruption repair, final idempotence.

PR #745, opened 2026-07-11
feat(federation): stream proxy — "From your peers" rows are now playable (discovery-over-federation P4)

IrosTheBeggar:
Phase 4 of discovery-over-federation (plan: `~/.claude/plans/federation-discovery-bridge.md`) — the payoff phase. Requires nothing new on the peer: it streams through the grants #739's endpoint already relies on.

## Server

**`GET /api/v1/federation/peers/:id/stream/<vpath-path>`** ([federation-stream.js](../blob/feat/federation-stream-proxy/src/api/federation-stream.js)) — a thin byte pipe over `fedFetch`:
- Forwards `range`/`if-range`/conditionals to the peer's `/media/<path>`; copies back the streaming headers (`content-type/length/range`, `accept-ranges`, `etag`, `last-modified`); pipes the body with teardown on client disconnect (an abandoned seek must not keep pulling bytes over the bridge).
- Normal local-user auth via the same `?token=` query every audio URL uses; **not** on the federation-key allowlist, so a peer can never chain proxies through us.
- No local path judgement beyond per-segment re-encoding — the path lives in the **peer's** vpath namespace and its wall + key grants decide readability. Unknown peer → 404 local; peer 404s relay; unreachable → 502; federation off → 403.

## Webapp

- Fed Discover rows now **queue** (click or the + icon) instead of copying; MusicBrainz link stays.
- New `addFederationSongWizard` — deliberately not `addSongWizard`: no transcode rerouting, waveform prefetch, live-playlist save, or metadata lookup (all resolve LOCAL paths). Lite metadata from the fed row renders the queue entry and now-playing card directly.
- The song object carries a `federation: {peerId, peerName}` marker, and the degrade guards key on it: the waveform fetcher treats remote tracks like external streams (no 404 spam), and the Discover panel clears rather than showing the previous song's results while a peer track plays. Rating/lyrics for remote tracks fail with the existing error paths (documented degrade, per plan).

## Verification

- **e2e** (two real servers over iroh, extends the aggregate suite — 9/9): proxied bytes are **byte-identical** to the file on the peer; `bytes=0-99` → 206 + correct `content-range` + exact first 100 bytes; unknown-peer 404, ungranted-library 404 relay; federation-off 403 for both aggregate and stream.
- **Live browser run** on the two-server rig: clicked the "Remote Hit" row → queued with the proxy URL + federation marker → advanced to it → the audio stack pulled `206 Partial Content` through the proxy and **the track played to completion** (player `currentTime` reached full duration); now-playing card and tab title showed "Remote Hit - Nova"; waveform requests fired only for the local track; zero console errors.

Phase 5 (snapshot fallback) stays deferred per the plan — this closes the arc the user asked about: federated discovery results you can actually play.

### DIFF
--- a/src/api/discovery-federation.js
+++ b/src/api/discovery-federation.js
@@ -40,6 +40,26 @@ const PEER_LIMIT_CAP = 100;
 // surface in `searched.unreachable`, not as an error.
 const PEER_TIMEOUT_MS = 4000;
 
+// AbortSignal.timeout on the fetch only bounds the HTTP request THROUGH an
+// established bridge — a dead peer stalls in the iroh DIAL that rebuilds
+// the bridge first, which the signal never covers (measured ~29s against a
+// stopped peer in the 2026-07-11 WAN smoke). Race the whole fedFetch call
+// against a deadline instead; a dial that eventually succeeds in the
+// background still warms the bridge for the next query.
+export async function fedFetchWithDeadline(fedClient, peer, apiPath, opts, ms) {
+  let timer;
+  try {
+    return await Promise.race([
+      fedClient.fedFetch(peer, apiPath, opts),
+      new Promise((_resolve, reject) => {
+        timer = setTimeout(() => reject(new Error(`no answer within ${ms}ms (dial included)`)), ms);
+      }),
+    ]);
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
 export function setup(mstream) {
   // Similar-but-not-owned tracks across the federated peers.
   //   filePath        the local seed track ("<vpath>/<relpath>")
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
**Two-machine WAN smoke completed** (local Lab A ↔ rum.st `~/mStream` smoke checkout, both on this branch, paired over real iroh — relay/holepunch path, 18,555 real EffNet-embedded tracks on the peer):

- Pairing + capability: ticket add → TOFU redeem → health `discovery: {effnet-discogs, 1280-d, 18555}`; warm bridge round-trip **17ms**.
- Real-model aggregate: **~310–540ms** end-to-end (5KB vector out, 18.5k-track scan on the peer, grant-scoped rows back). Similarity quality visibly sane: Icarus seeds → Chase & Status / Seven Lions / ODESZA at 0.83–0.88 with Discogs tags + MBIDs.
- Panel + playback in a real browser: "From your peers" rows rendered from rum-st, row click queued a real Chase & Status MP3, **played + seeked (2:00) over the WAN**; 20.6MB FLAC pulled at **4.2 MB/s** with correct mid-file 206s; zero console errors.
- **Cold-bridge concern from the PR review: measured fine** — after the 5-min idle teardown, a cold dial + query completed in **541ms** (~7× headroom under the 4s budget).
- **One real bug found and fixed here** (last commit): with the peer *offline*, the aggregate took ~29s — `AbortSignal.timeout` never covered the iroh **dial**, only the fetch through an established bridge. `fedFetchWithDeadline` races the dial too: offline aggregate now **4.1s** with `unreachable:1`, offline stream play → crisp **502 @ 15s**; happy path unchanged. New e2e regression test pins it (suite now 10/10).

Also hit during setup (not this PR's code): a lab DB whose `user_version=57` came from the open #716 branch (model-genres V57) broke the V58 migration — the documented version-slot-collision gotcha; repaired the lab DB by hand. Production lineage unaffected (rum.st migrated 57→58 cleanly).

PR #751, opened 2026-07-16
feat(discovery): auto-forget catalog peers after a retention window

IrosTheBeggar:
## The problem

Once a peer's signed announcement is heard, its catalog entry lives forever. A server that died months ago still shows in every operator's Discovery page as a bare "offline" — indistinguishable from one that dropped off five minutes ago. (Spotted via a long-dead test lab box that never left the list.)

Related gap fixed along the way: adding a peer to `discoveryP2p.blockedPeers` stopped *new* announcements from being recorded, but never removed or hid the already-recorded entry.

## The fix

New `discoveryP2p.peerRetentionDays` (default **30**, `0` = keep forever). An hourly pass in `discovery-catalog.js` (`pruneStalePeers`) drops:

- entries not heard from within the retention window, and
- blocked peers of any age.

A pruned peer re-enters the catalog the moment it announces again — forgetting is never permanent.

### Guardrails

- **Neighbor gate**: the pass only runs while the sidecar reports ≥1 gossip neighbor. With zero neighbors, silence is evidence of our own isolation, not of peers being gone — a server that boots after months offline must never wipe its whole catalog. (Live peers re-announce every ~15s, so everyone reachable has refreshed `updatedAt` long before the first pass at +1h.)
- **Shelf pin**: peers whose snapshot is on the local shelf are never pruned — the catalog row is what makes the downloaded snapshot visible/manageable in the UI, so forgetting it would orphan disk usage invisibly. Snapshot removal stays an explicit operator action; after that the row ages out normally. The pin-set is injected as a callback from `discovery-p2p-stack` (the catalog keeps not importing `discovery-peer-dbs`; the dependency already points the other way).
- **Timestamp honesty**: same-seq heartbeats refresh `updatedAt` only in memory (`record()` deliberately doesn't save them). Each hourly pass now schedules a save, so persisted timestamps stay ≤1h stale across restarts.

## Admin surface (max-storage pattern)

- `POST /api/v1/admin/discovery/p2p/peer-retention` — live, no restart; the next pass reads the config fresh. Joi mirrors the config bounds (0–3650).
- Value exposed in the p2p status payload.
- Discovery page: "Forget offline servers: after N days of silence [edit]" line + modal (0 renders as "never").
- The catalog table's Online column now shows how long a silent peer has been gone — `offline · 45d` — with the full timestamp as a hover title.

## Testing

- New `test/unit/discovery-catalog-prune.test.mjs` (6 tests): stale/blocked/unparseable-timestamp dropped; fresh + shelf-pinned kept; keep-set shields blocked peers; retention 0 disables age pruning but not blocked-drop; pruned peer re-enters on next announcement. Crafts `catalog.json` before the module's lazy first load to control entry ages.
- Full `discovery-p2p` integration suite (covers the stack start/stop paths the prune timer is wired into): 36/36.
- Full unit run: 309 pass / 0 fail.
- Browser click-through on a live instance with a seeded catalog: panel line + modal save (200, persisted, re-rendered), 400 on out-of-range, offline ages in the table.

## Behavior-change note

Retention defaults to **on at 30 days** for existing servers. Operators who want the old keep-forever behavior set the value to 0 in the new modal (or config).

### DIFF
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -494,6 +494,7 @@ export function setup(mstream) {
       serverName: config.program.discoveryP2p.serverName,
       serverDescription: config.program.discoveryP2p.serverDescription,
       maxPeerDbStorageMb: config.program.discoveryP2p.maxPeerDbStorageMb,
+      peerRetentionDays: config.program.discoveryP2p.peerRetentionDays,
     });
   });
 
@@ -638,6 +639,21 @@ export function setup(mstream) {
     });
     joiValidate(schema, req.body);
     await admin.editMaxPeerDbStorageMb(req.body.maxPeerDbStorageMb);
+    res.json({});
+  });
+
+  // How long a silent peer stays in the catalog before the hourly prune
+  // pass forgets it (0 = forever). Live: the pass reads the config fresh, so
+  // no restart and no stack bounce. Bounds mirror the config Joi
+  // (discoveryP2pOptions.peerRetentionDays). Peers whose snapshot is on the
+  // local shelf are never pruned regardless of this value.
+  mstream.post("/api/v1/admin/discovery/p2p/peer-retention", async (req, res) => {
+    requireP2pEnabled();
+    const schema = Joi.object({
+      peerRetentionDays: Joi.number().integer().min(0).max(3650).required(),
+    });
+    joiValidate(schema, req.body);
+    await admin.editPeerRetentionDays(req.body.peerRetentionDays);
     res.json({});
   });
 
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -564,6 +564,24 @@ export function setup(mstream) {
     res.json({});
   });
 
+  // Forget a catalog peer right now instead of waiting out the retention
+  // window. Same pin rule as the automatic prune: a peer whose snapshot is
+  // on the local shelf stays — forgetting it would orphan the downloaded
+  // file invisibly (remove the snapshot first, then forget). Reversible by
+  // nature: the peer re-enters the catalog on its next announcement.
+  mstream.post("/api/v1/admin/discovery/p2p/forget", (req, res) => {
+    requireP2pEnabled();
+    const schema = Joi.object({ endpointId: Joi.string().hex().length(64).required() });
+    joiValidate(schema, req.body);
+    if (discoveryPeerDbs.get(req.body.endpointId)) {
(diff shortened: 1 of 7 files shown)


IrosTheBeggar:
**Live two-server smoke (2026-07-16)** — real boot path, real sidecars, prune interval shrunk to 15s via the new `MSTREAM_TEST_DISCOVERY_PRUNE_MS` override (commit b8f6eeda):

Setup: server A booted with a pre-seeded catalog — `Test Lab A` (45 days silent), `Attic NAS` (6 days), `Living Room` (fresh) — retention 30, auto-fetch off, community seeds off.

1. **Isolation gate**: A alone (`joined: true, neighbors: 0`). Three prune passes fired over 45s → **nothing forgotten**, all 3 peers still listed. Silence while isolated proves nothing, so the catalog is untouched. ✅
2. **Real prune**: server B booted with A's ticket in `bootstrapPeers` → A reports `neighbors: 1` → next pass logs `[discovery-catalog] forgot 1 peer(s) (not heard in 30d): aaaaaaaaaaaa…` → catalog API returns only `Living Room, Attic NAS`. Exactly the stale peer dropped, the 6-day peer kept. ✅
3. **Persistence**: `discovery-p2p/catalog.json` on disk rewritten to the 2 surviving entries. ✅
4. Zero errors in either server's log.

IrosTheBeggar:
Two additions since opening:

- `b8f6eeda` — `MSTREAM_TEST_DISCOVERY_PRUNE_MS` env override for the prune interval (integration/smoke testability, mirrors the shelf reconciler's override).
- `86dcab5f` — per-peer **[Forget]** action on offline catalog rows: drop a dead server now instead of waiting out the retention window. Same shelf-pin rule as the automatic prune (409 while a downloaded snapshot exists); reversible by nature since one announcement re-creates the entry. Live-verified in the browser (row disappears, catalog.json updated) + unit-tested.

Follow-up stack: #752 (befriend-a-server ticket box) → #753 (runtime blocklist UI). Merge bottom-up: this PR first.

IrosTheBeggar:
Follow-up live smoke (2026-07-23, details on #753): [Forget] exercised against a *live announcing* peer — forgotten row reappeared in 13s via its next real announcement, and the 409 shelf-pin path verified with a real downloaded snapshot. All #751 behaviors now observed live, none inferred.

PR #753, opened 2026-07-16
feat(discovery): runtime blocklist UI — block/unblock peers from the catalog

IrosTheBeggar:
**Stacked on the join-box PR** (merge order: #751 → join box → this). 

## The problem

`blockedPeers` is the public discovery network's only abuse lever, and it was config-file-only — on a network where any stranger's server can appear in your catalog with an arbitrary name and description. A runtime blocklist UI has been on the wishlist since the 6.16.0 soft-launch.

## The fix

- **[Block]** action on every catalog row. One server-side action: appends to the config blocklist, drops any downloaded snapshot (unpins the blob, re-fetchable later), and forgets the catalog row — blocked means "doesn't exist", present tense included. Blocking your own endpoint id is refused (400).
- **Blocked servers** list below the table (truncated endpoint ids) with per-id **[Unblock]**. Unblocking restores nothing proactively — the peer re-enters the catalog on its next announcement (~15s while online).
- The status payload now exposes `blockedPeers`.

Everything downstream was already blocklist-aware (`record()` refuses announcements, holds beacons ignored, fetch refuses, the retention prune drops lingering entries) — this PR just gives the lever a handle.

## Testing

- Live click-through: Block → row disappears + blocked section lists the id + config file updated; Unblock → section empties + config cleared. Toasts on both.
- Full discovery-p2p integration suite run on this branch.

### DIFF
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -495,6 +495,7 @@ export function setup(mstream) {
       serverDescription: config.program.discoveryP2p.serverDescription,
       maxPeerDbStorageMb: config.program.discoveryP2p.maxPeerDbStorageMb,
       peerRetentionDays: config.program.discoveryP2p.peerRetentionDays,
+      blockedPeers: config.program.discoveryP2p.blockedPeers,
     });
   });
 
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
**Live enforcement smoke (2026-07-23)** — the earlier click-through used inert seeded peers; this run used a *genuinely announcing* peer (server B: 35s tone + `test-fake` embedding model → real snapshot, real 15s announce loop), so the enforcement semantics are now observed, not inferred:

1. **Block is composite**: one `POST /block` → config `blockedPeers` gained B, catalog row dropped, downloaded snapshot dropped (`shelfBytes: 0`). ✅
2. **Blocked = ignored**: through **50s ≈ 3+ real announce cycles**, B never re-entered the catalog — `record()` genuinely refuses blocked announcements. ✅
3. **Unblock is live**: `POST /unblock` → B reappeared **5 seconds** later via its next announcement; config cleared. ✅
4. **Self-block refused**: blocking A's own endpointId → `400 "cannot block this server itself"`. ✅

Also exercised on the same run (relevant to #751): manual [Forget] of the live peer → catalog emptied → **reappeared in 13s** via real announcement; and the shelf-pin rule with a *real* fetched snapshot (20KB, BLAKE3-verified transfer) → forget returned `409 "peer has a downloaded snapshot — remove the snapshot first"` until the snapshot was removed. ✅

PR #758, opened 2026-07-18
feat(scanner): filesystem watcher for near-instant targeted scans

IrosTheBeggar:
## What

Watches each library root (chokidar) and turns filesystem events into **targeted scans through the existing machinery** — the watcher never touches the index itself, it only enqueues through the same task-queue dedup every other trigger uses, so a missed or duplicated event can never corrupt anything. Worst case is a redundant subtree scan or falling back to the interval scan.

Control loop (modeled on Navidrome's watcher):

- **Relevance filter**: scan-ignore rules (blocklist + LIVE dot-entry flags) plus watched extensions — configured audio formats, folder art (jpg/jpeg/png), lyric sidecars (lrc/txt).
- **Target mapping**: file events → parent directory; deletion events walk UP to the nearest still-existing ancestor; the library root itself degrades to a full vpath scan.
- **Coalescing**: per-vpath pending set, debounced until the library is QUIET for `watcherWait` seconds (default 10 — a torrent writing 200 files becomes one subtree scan), re-armed at 3× while a scan is already running. `awaitWriteFinish` (2s stability) keeps half-written files from triggering.
- **Flush**: nested targets collapse to their shallowest ancestor (segment-boundary safe) → `addSubtreeScanTask`, whose scoped sweep (#757) makes deletions land too.

**Default OFF** (`scanOptions.watcherEnabled`): most CIFS/NFS mounts emit no change events, so the `scanInterval` loop remains the delivery mechanism there — the watcher is an accelerator for local disks, never a replacement. Live admin toggle (persist + start/stop in-process, no reboot) + admin-panel row; directory add/remove re-syncs the watched set; boot/`reboot()` evaluate the flag idempotently. Watcher errors (inotify limits, permissions) degrade to a warn + quiet watcher, never a crash.

## Tests

- `test/unit/library-watcher.test.mjs`: pure mapping/coalescing logic (relevance incl. live flag semantics, walk-up, segment-boundary collapse) + ScanCoalescer under node:test **mock timers** (debounce re-arm, 3× active-scan back-off) + one real-chokidar round trip over a temp dir (10/10).
- `admin-scan-params` gains the four-part coverage for `watcher-enabled` — the happy-path flip exercises the real start/stop side effect inside the booted test server (22/22).
- Torrent completion-watcher suites green (33/33) — the other subtree-scan producer.

## Live smoke (real server)

Enabled the watcher via the admin endpoint on a running instance, then **without any manual scan trigger**: dropped a new file → auto-indexed in **12s**; deleted it → auto-converged in **10s**. The scan log shows `(1 scanned)` / `(0 scanned)` — the watcher-fired scans visited only the affected subtree, and the deletion line composed all three scanner PRs: scoped sweep removed the row, move re-homing paired its byte-identical twin.

## Notes

- New dependency: `chokidar` ^4.0.3 (v4 — no glob engine; recursive watching that works on the 3-OS matrix; Node's bare fs.watch recursive is newer on Linux and leaks watch descriptors on large trees).
- `watcherWait` is config-file-only for now (applied when watchers (re)start); the admin UI gets the boolean toggle.
- No schema changes; scanners untouched (no parity impact).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "archiver": "^8.0.0",
         "busboy": "^1.6.0",
+        "chokidar": "^4.0.3",
         "command-exists": "^1.2.9",
         "cookie-parser": "^1.4.7",
         "essentia.js": "^0.1.3",
@@ -2025,6 +2026,21 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
+      }
+    },
+    "node_modules/chokidar": {
+      "version": "4.0.3",
+      "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-4.0.3.tgz",
+      "integrity": "sha512-Qgzu8kfBvo+cA4962jnP1KkS6Dop5NS6g7R5LFYJr4b8Ub94PPQXUksCw9PvXoeXPRRddRNC5C1JQUR2SMGtnA==",
+      "license": "MIT",
+      "dependencies": {
+        "readdirp": "^4.0.1"
+      },
+      "engines": {
+        "node": ">= 14.16.0"
+      },
+      "funding": {
+        "url": "https://paulmillr.com/funding/"
       }
     },
     "node_modules/color": {
@@ -4095,6 +4111,19 @@
       },
       "funding": {
         "url": "https://github.com/sponsors/yqnn"
+      }
+    },
+    "node_modules/readdirp": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/readdirp/-/readdirp-4.1.2.tgz",
+      "integrity": "sha512-GDhwkLfywWL2s6vEjyhri+eXmfH6j1L7JE27WhqLeYzoh/A3DBaYGEj2H/HFZCn/kMfim73FXxEJTw06WtxQwg==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 14.18.0"
+      },
+      "funding": {
+        "type": "individual",
+        "url": "https://paulmillr.com/funding/"
       }
     },
     "node_modules/roarr": {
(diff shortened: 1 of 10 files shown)


IrosTheBeggar:
**Edge-case smoke (live server, no manual scan triggers): 17/18 checks pass — the one failure is a pre-existing scanner limitation the smoke surfaced, not a watcher defect.**

Verified live, correlated against scan-log timestamps:
- single add auto-indexed in 12s; recursive folder delete auto-converged in 10s
- **burst coalescing**: 8 rapid file drops → exactly ONE scan line (`8 files processed (8 scanned)`)
- **nested collapse**: files in `Nest/` + `Nest/Deep/` in one burst → ONE scan (`2 scanned`)
- **folder rename with a playlist ref** → `1 moved track(s) re-homed (1 reference(s) rewritten)` — watcher + scoped sweep + move re-homing composing
- **negatives stayed quiet** (zero scan lines in each window): `#recycle` churn, an irrelevant `.pdf`, a dot-file while `ignoreDotFiles` is on
- **live dot-flag flip**: toggling `ignoreDotFiles` off applies to the very next event (root-level dot file → full-vpath scan indexed all 3 dot files)
- **live library add**: `PUT /admin/directory` re-synced the watched set — a file dropped into the new library auto-indexed
- **off-toggle**: no auto-scan while disabled; a manual rescan then found the file (control)

**Surfaced (pre-existing, filed separately):** dropping a `cover.jpg` beside UNCHANGED tracks fires the correct targeted scan, but the scan reports `2 unchanged` and links nothing — folder-art discovery only runs when a track is re-parsed (the mtime fast-path skips it), same behaviour as a manual rescan today. The watcher makes this gap user-visible; fix tracked as its own task (detect folder-image drift for fast-pathed directories — the V50 art snapshot machinery is most of the way there).

PR #798, opened 2026-08-01
Remove local Claude configuration

BubbleMoth:
This PR removes `.claude/settings.local.json` from version control.

The file contains machine-specific Claude configuration and is already covered by the existing `.claude/*` rule in `.gitignore`, so it shouldn't be tracked anyway.

The tracked file may also include local configuration, such as credentials or machine-specific settings, that is not intended for the repository.


IrosTheBeggar:
Thanks for fixing this. PR is merged

PR #799, opened 2026-08-01
fix(boot): a broken console pipe must not kill an orphaned server

IrosTheBeggar:
## The bug

When the process that spawned mStream — the desktop app, Docker, npm, a service manager — dies without reaping it, the server keeps running but its stdout/stderr pipes lose their reader. The next console write raises **EPIPE**, and winston's Console transport writes to `process.stdout` with no `'error'` listener, so the runtime escalates it to an uncaught exception. A healthy orphan drops dead the moment it logs *anything*.

Observed in the field (macOS desktop-app smoke test, v6.19.2 standalone binary): an orphaned server answered pings normally until an admin `POST` made it log the request — it exited mid-probe, and its crash report went down the same dead pipe, so nothing was ever captured. The mstream_music client now recycles orphans instead of adopting them specifically to route around this; the server surviving is the real fix.

`exitOnError: false` in `logger.js` doesn't cover this: the error fires on the raw stream, outside winston. This PR extends that same existing policy — a logging failure must not kill the process — to the one channel outside winston's reach.

## The fix

`src/util/stdio-guard.js` — swallow `'error'` events on `process.stdout`/`process.stderr`, imported as the **first** import of `cli-boot-wrapper.js` so the guard is installed before any module can write (ESM import order guarantees this).

Scoping, refined in review:

- **Bun runtime only.** The shipped standalone binaries are Bun builds, and they are what supervisors spawn and orphan — the guard installs only when `process.versions.bun` is present, so **node/npm installs keep Node's stock fail-fast stdio semantics, unchanged from every released version**. This is a distribution-scoping decision, not a technical boundary: the underlying bug is runtime-agnostic (the regression test proves Node dies identically), and the gate is documented as deliberately widenable if the crash class ever matters for Node deployments (e.g. systemd units whose journald restarts). Detection is `process.versions.bun` rather than `isBunStandalone` because stdio-guard must import nothing, and a runtime gate is testable without compiling a binary.
- **Server only — workers keep fail-fast stdio.** A worker's stdout is not a console, it's the line-protocol IPC channel to the server that spawned it; that pipe's read end can only die with the server itself, so EPIPE there means "your consumer is gone" and dying is the correct teardown. The guard no-ops when argv carries `--mstream-worker=`, matching Node-mode workers (which fork loose scripts and never load it). Without this, an orphaned scanner/embedding worker would grind to completion as a zombie instead of exiting on its next protocol line.
- **One breadcrumb when it fires.** The first failed write pushes `[stdio] <stream> write failed (…) — console output is no longer being delivered` into the in-memory ring, so the admin live-log viewer records why console output stopped. Routed straight into the ring — not through winston, whose Console transport would write to the very stream that just broke.

mStream is a daemon, not a shell filter — losing the console must never take the (packaged) server down. The file and in-memory ring transports keep logging either way; normal console output is unaffected (the listener is inert until a write has already failed at the OS level).

## Verification

- **Regression test** (`test/integration/stdio-epipe-orphan.test.mjs`) pins both halves of the runtime gate, simulating the dead parent by destroying the test's read ends of the child's stdio pipes and triggering log writes via failed logins (`winston.warn` in `api/auth.js` fires unconditionally):
  - **Node leg** (always runs): the server must **die** on its first post-breakage write — stock semantics preserved, guard must not install. Verified: exits within ~0.5s.
  - **Bun leg** (`execPath: 'bun'`, skipped where bun isn't installed): the server must survive, keep answering ping, and the `[stdio]` breadcrumb must appear via `/api/v1/admin/logs/recent`.
- **Pre-gate evidence** the guard itself works (from the earlier revisions of this branch, where it was verified under Node before being scoped to Bun): unguarded server exits code 1 on first write; guarded server survived the same test plus a true-orphan repro (parent process actually dead) through 3 log-triggering requests.
- `npm run test:unit` — 373 pass. Integration slice through the modified helper (`notfound-codes`, `admin-access`, `random-route`) — 113 pass. ESLint clean.

Known limit: with the gate in place, **no automated job currently executes the guarded path** — the dev machine has no bun (Bun leg skips) and CI's build matrix compiles but doesn't run tests. Before a release leans on this, either run the Bun leg once on a machine with bun (`bun --version` present → `node --test test/integration/stdio-epipe-orphan.test.mjs`), or add a small CI test job — the workflow already installs bun for builds.

### DIFF
--- a/cli-boot-wrapper.js
+++ b/cli-boot-wrapper.js
@@ -1,5 +1,8 @@
 #!/usr/bin/env node
 
+// Must stay the first import: installs the stdout/stderr error guards before
+// any other module can write to the console (see src/util/stdio-guard.js).
+import './src/util/stdio-guard.js';
 import { join } from 'path';
 import { maybeRunWorker } from './src/util/worker-process.js';
 import { appRoot } from './src/util/esm-helpers.js';
--- a/cli-boot-wrapper.js
+++ b/cli-boot-wrapper.js
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
-// Must stay the first import: installs the stdout/stderr error guards before
-// any other module can write to the console (see src/util/stdio-guard.js).
+// Must stay the first import: installs the stdout/stderr error guards (Bun
+// runtime only — see src/util/stdio-guard.js) before any other module can
+// write to the console.
 import './src/util/stdio-guard.js';
 import { join } from 'path';
 import { maybeRunWorker } from './src/util/worker-process.js';
--- a/src/logger.js
+++ b/src/logger.js
@@ -219,6 +219,18 @@ export function setBufferCapacity(n) {
   }
 }
 
+// Out-of-band breadcrumb from the stdio guard (src/util/stdio-guard.js): a
+// console write just failed, so this one event CANNOT be logged through
+// winston — the Console transport would write to the very stream that broke.
+// Push straight into the ring so the admin live-log viewer still records why
+// console output stopped. (The on-disk file transport is deliberately not
+// invoked: this path stays free of anything that could itself fail.)
+export function noteStreamFailure(streamName, code) {
+  pushEntry('warn',
+    `[stdio] ${streamName} write failed (${code || 'unknown error'}) — ` +
+    'console output is no longer being delivered; live/file logging continues');
+}
+
 // Read recent log entries for the admin live-log viewer. `sinceSeq` is the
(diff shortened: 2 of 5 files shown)


IrosTheBeggar:
Closing in favor of a different design direction: rather than surviving the loss of its console, the server should treat a lost supervisor as a shutdown signal — self-terminate cleanly instead of either crashing (today) or living on as an unowned orphan holding the port (this PR). A successor PR implements that.

PR #803, opened 2026-08-03
fix(mac): give the .app double-click a visible outcome

### LINKED ISSUE #802: mStream-6.19.2-darwin-arm64 Binary fails to load macOS 26.5.2 (25F84)
Attempting to start mstream from the mStream-6.19.2-darwin-arm64 binary on macOS 26.5.2 (25F84) fails. The icon bounces around for a while on the taskbar, then stops (icon remains). Right click * show all windows - nothing but the message "application is not responding." Happy to provide any other information but not sure where to look. Console logs did not appear to illuminate the issue.

IrosTheBeggar:
Fixes #802.

## What the bug actually was

The darwin bundles wrap the faceless CLI server binary in `mStream.app` so Finder shows an icon — but a regular `APPL` bundle is expected to check in with the window server, and a CLI process never does. So a Finder double-click bounced the Dock icon into a permanent **"Application Not Responding"** while the server was, in fact, up and serving on port 3000 the whole time. With stdout at /dev/null under LaunchServices, no window, no browser open, and file logging off by default, there was zero visible evidence of life — "fails to load" was the reasonable read.

Reproduced 1:1 on the reporter's exact macOS build (26.5.2 / 25F84): `lsappinfo` showed `type="Foreground"` + `!cgsConnection` (the ANR state) with `/api/v1/ping` answering 200 underneath. Gatekeeper translocation was ruled out — a quarantined copy ran in place, un-translocated.

## The fix: make the double-click mean something

All desktop behavior keys off the `__CFBundleIdentifier` env marker, which LaunchServices sets only when launched *as* the bundle — terminal, daemon, and CI invocations keep pure CLI behavior (verified: the shell child is still the listening pid).

- **Info.plist** (`scripts/build-bun.mjs`): `LSUIElement` so macOS stops expecting a window-server checkin — no bounce, no ANR. Plus `NSLocalNetworkUsageDescription`/`NSBonjourServices` to caption the macOS 15+ Local Network consent prompt that default-on mDNS triggers.
- **`src/util/mac-app-launch.js`** (new): on a Finder launch the process re-spawns itself detached and exits, hopping out of LaunchServices' running-app registry — otherwise later double-clicks get delivered to the existing process as reopen Apple Events a faceless server can't receive (visibly: nothing happens). Once listening, it opens the web UI in the default browser. Fatal boot errors raise a native osascript alert instead of dying silently.
- **`src/server.js`**: every response now carries an `X-Mstream` marker header, so a relaunch that finds the port taken can probe it, recognize the earlier instance regardless of auth mode, refocus the browser at it, and exit 0 as redundant-not-failed. A port held by something else gets a clear alert.

Net effect: double-click → browser opens with your music. Double-click again later → browser opens again. Real failure → real error dialog.

## Two latent crash paths fixed for all platforms

Surfaced while testing, pre-existing on master:

1. **A failed `listen()` (e.g. `EADDRINUSE`) was an uncaught crash.** `ws`'s WebSocketServer forwards the wrapped HTTP server's `'error'` events onto the wss; with no listener there, the re-emit throws mid-dispatch and kills the process before any later-registered server handler can run. `remote.js` now listens (log-only), and `server.js` handles listen errors properly — clean message + nonzero exit under both Node and Bun.
2. **`serveIt()` rejections escaped `cli-boot-wrapper.js` as unhandled rejections** (e.g. bad SSL certs). Now caught: logged, alerted on app launches, exit 1.

## Reviewer notes

- The bundle id string `io.mstream.server` intentionally appears in both the plist template and `mac-app-launch.js`; comments on both sides cross-reference the pairing.
- `X-Mstream: true` deliberately carries no version. mStream already announces itself on the LAN via `_mstream._tcp` mDNS by default, so the header reveals nothing new.
- The pre-existing `no-empty` lint errors in `remote.js`/`server.js` are untouched.

## Verification

On macOS 26.5.2: LSUIElement state via `lsappinfo`, browser open on first click, exactly one server process across repeated double-clicks, redundant app-launch exits 0, clean `EADDRINUSE` failure from a terminal under both runtimes, terminal foreground contract intact. Suite: **2047 pass / 0 fail** (unit + integration + full), 7 new unit tests for the launch-detection helpers.

## Known limitations

- **No user-facing Quit for the macOS app (deferred).** As an `LSUIElement` agent the server has no Dock presence, so a Finder-launched instance can only be stopped via Activity Monitor or `kill` — re-clicking the app just reopens the browser. A proper Quit is deferred to the planned macOS menu-bar status item (the tray/autostart effort), which will drive a graceful-shutdown path rather than a bare kill. Documented inline in `scripts/build-bun.mjs` and `src/util/mac-app-launch.js`.

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -212,21 +212,36 @@ jobs:
             if curl -sf -o /dev/null "http://127.0.0.1:3401/api/v1/ping"; then UP=1; break; fi
           done
           if [ "$UP" != "1" ]; then echo "FAIL: detached child never served — the app-launch hand-off is broken"; cat "$RUN/applaunch.log"; exit 1; fi
-          # The detached child must be a DIFFERENT process than the one we ran.
           CHILD=$(pgrep -f "applaunch-config.json" | head -1)
           echo "ok: launcher exited 0, detached child $CHILD serving on 3401"
-          # Relaunching while it runs must exit 0 (redundant, not failed) and
-          # leave exactly one server alive — the nonce probe proving identity.
+
+          # Relaunch, full Finder path: detaches again, and the detached copy is
+          # the one that probes. Its stdio is /dev/null by design, so the only
+          # observable contract here is "exit 0 and don't disturb the running
+          # server" — asserting on its log would test nothing (that mistake is
+          # why this step first shipped red).
           __CFBundleIdentifier=io.mstream.server "$BIN" -j "$RUN/applaunch-config.json" > "$RUN/applaunch2.log" 2>&1
           RC2=$?
-          sleep 2
+          sleep 4
+          COUNT=$(pgrep -f "applaunch-config.json" | wc -l | tr -d ' ')
+          curl -sf -o /dev/null "http://127.0.0.1:3401/api/v1/ping"; STILL=$?
+          if [ "$RC2" != "0" ]; then echo "FAIL: relaunch exited $RC2, expected 0"; cat "$RUN/applaunch2.log"; pkill -f applaunch-config.json; exit 1; fi
+          if [ "$COUNT" != "1" ] || [ "$STILL" != "0" ]; then echo "FAIL: relaunch disturbed the running server (count=$COUNT alive=$STILL)"; pkill -f applaunch-config.json; exit 1; fi
+          echo "ok: redundant relaunch exited 0 and left the running server alone"
+
+          # Relaunch with the detach already done (MSTREAM_MAC_APP_DETACHED=1 is
+          # what the detached copy itself carries), so the probe runs in THIS
+          # process where its result is observable. This is the assertion that
+          # actually covers the nonce identity check.
+          __CFBundleIdentifier=io.mstream.server MSTREAM_MAC_APP_DETACHED=1 "$BIN" -j "$RUN/applaunch-config.json" > "$RUN/applaunch3.log" 2>&1
+          RC3=$?
           COUNT=$(pgrep -f "applaunch-config.json" | wc -l | tr -d ' ')
           curl -sf -o /dev/null "http://127.0.0.1:3401/api/v1/ping"; STILL=$?
           pkill -f "applaunch-config.json" 2>/dev/null
-          if [ "$RC2" != "0" ]; then echo "FAIL: relaunch against a running instance exited $RC2, expected 0"; cat "$RUN/applaunch2.log"; exit 1; fi
-          grep -q "already running" "$RUN/applaunch2.log" || { echo "FAIL: relaunch did not recognize the running instance (nonce probe failed)"; cat "$RUN/applaunch2.log"; exit 1; }
+          if [ "$RC3" != "0" ]; then echo "FAIL: in-process relaunch exited $RC3, expected 0 (redundant, not failed)"; cat "$RUN/applaunch3.log"; exit 1; fi
+          grep -q "already running" "$RUN/applaunch3.log" || { echo "FAIL: the nonce probe did not recognize our own running instance"; cat "$RUN/applaunch3.log"; exit 1; }
           if [ "$COUNT" != "1" ] || [ "$STILL" != "0" ]; then echo "FAIL: relaunch disturbed the running server (count=$COUNT alive=$STILL)"; exit 1; fi
-          echo "SMOKE OK: app-launch detach, serve, and redundant-relaunch all behave"
+          echo "SMOKE OK: app-launch detach, serve, redundant relaunch, and nonce probe all behave"
 
       - name: Smoke test iroh native binding
         # Proves the standalone binary can dlopen the bundled @number0/iroh .node
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -122,11 +122,29 @@ jobs:
           CODE=$(curl -s -o /dev/null -w '%{http_code}' "http://127.0.0.1:3399/api/v1/ping" 2>/dev/null)
           echo "HTTP up=$UP after ~${i}s ; /api/v1/ping -> ${CODE:-000}"
           grep -iE "scan started|scan complete" "$RUN/boot.log"
+
+          # (4) soft reboot must survive under the BUN runtime. Node's test
+          # suite can't catch this class: reboot()'s delayed sweep once hit
+          # the NEW server, and Bun's closeAllConnections (unlike Node's)
+          # stops the listener outright — the binary kept running with no
+          # port. Trigger a reboot-requiring config change, require the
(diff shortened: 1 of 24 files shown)


IrosTheBeggar:
Extended smoke pass over this PR surfaced a **pre-existing** bug now fixed in 25f379d6: on every Bun standalone binary, any admin-panel change that soft-reboots the server (port, address, trust-proxy, UI switch) killed the listener ~1s after it came back up — process alive, port dead, log claiming success. Root cause is a two-layer interaction: `reboot()`'s delayed `closeAllConnections()` sweep captures the module-level `server` variable (reassigned to the NEW server by the time it fires), and Bun's `closeAllConnections()` — unlike Node's sockets-only version — stops the server outright. Verified broken at the pre-PR commit under Bun and invisible under Node, which is why the test suite never caught it; the CI bundle smoke now asserts reboot survival under the Bun runtime.

Full smoke matrix run on macOS 26.5.2 (arm64 native + x64 via Rosetta), all green:
- all three failure dialogs render end-to-end (bad config, bad SSL certs, port taken by a non-mStream app) and survive the exiting process
- https self-signed probe and explicit `127.0.0.1` bind: redundant relaunch recognizes the running instance, exits 0
- two simultaneous first launches: exactly one server survives, no spurious dialog (×2 rounds)
- app-mode admin reboots ×2: recover in ~1s, sustained past the sweep window
- CI-parity steps on both darwin bundles: self-dispatch, default boot, serve+scan, reboot survival; iroh OK on arm64, unavailable on x64 per the known upstream npm gap (no darwin-x64 prebuilt)

Still needs a human at the GUI: the Gatekeeper unapproved-quarantine first-run dance, and the Local Network prompt caption (TCC state isn't readable programmatically).

IrosTheBeggar:
Follow-up to the smoke report: the win-x64 failure the new reboot smoke caught was a third pre-existing layer — Windows doesn't grant the immediate same-process port rebind unix does, so the reboot's fresh `listen()` hit `EADDRINUSE` against the old, still-releasing listener and died (pre-#803 this same race was an uncaught crash; it was invisible because nothing exercised reboot). Fixed in 1017f566: `serveIt` takes a `relisten` flag set only by `reboot()` — EADDRINUSE retries every 250ms for up to 5s, then falls through to the normal fatal path; first boots keep fail-fast semantics. Verified with a port-holder simulation under Node and Bun (release after 1.2s → 5 retries → exactly one boot callback → 200).

CI is now green on all seven targets, with the bundle smoke asserting reboot survival under the Bun runtime on win-x64, linux-x64, and darwin-arm64.

IrosTheBeggar:
## Audit of this PR + Bun-runtime bug hunt

Ran a 31-agent audit over the branch diff and the shipped-binary paths, every finding adversarially verified with runtime repros under bun 1.3.14 and node v26.5.0. 20 findings confirmed, 1 refuted on re-verification (below). Fixes are in `cdfd5418` (this PR's own defects) and `9e30400b` (pre-existing Bun divergences), each reproduced-before / verified-after.

### Critical regression this PR introduced — fixed

`reboot()` has no in-flight guard, and `close()` on an already-closing server still runs its callback, so **two quick admin saves start two `serveIt()`s racing for the port**. Before this branch that race died as an uncaught `'error'`; the relisten retry in `1017f566` turned it into something worse — the loser burns its 5s budget and `process.exit(1)`s the process hosting the healthy winner. Deterministic repro: **DEAD 2/2** at `1017f566`, **ALIVE 5/5** after the fix (bun + node). On a Finder-launched `.app` the death even read as success, since the probe found the sibling and exited 0.

### Other defects in this PR — fixed

- **The wss `'error'` listener I added doesn't cover per-connection errors.** `ws` raises protocol failures on the individual socket, and `verifyClient` skips auth entirely on a fresh install — so **six bytes from an unauthenticated client killed the server** under Node (`c1 80 00 00 00 00` → dead; survives + logs now). Shipped Bun binaries were immune, the asymmetry running backwards for once.
- **`25f379d6`'s forced-close sweep is a no-op under Bun** — its shim nulls the internal handle inside `close()`, so `closeAllConnections()` early-returns and a reboot during an active transfer never completes. Replaced with per-instance socket tracking (works on both runtimes; verified with a rate-limited in-flight request).
- **`reboot()`'s `serveIt()` had no rejection handler** — a cert rotated out from under a reboot died as a raw unhandled rejection with the old server already gone: the exact silent death #802 was about, relocated.
- **The `.app` probe never settled under Bun** against a squatter that accepts and stalls, and `X-Mstream` alone "proved" identity — so any local process could set it and have us open the user's browser at its page **on the real UI's origin, where the client keeps its token**. Identity now rests on a per-boot nonce published only to loopback callers and mirrored to a `0600` file; the probe settles on `'close'`. Legit relaunch still exits 0 and reopens the browser; a spoofed header now gets the alert.

### Pre-existing Bun bugs — fixed

- **AcoustID has been dead in every shipped binary**: `task-queue` launches the `acoustid` worker but `maybeRunWorker()` had no case, so the self-dispatched child exited `Unknown mStream worker role`. Node forks the script directly and never consults the switch — structurally invisible to the suite. A new parity test now fails if any role lacks a case.
- **mDNS dead on arrival under Node on macOS**: `reusePort` exists only on Linux/FreeBSD, so bind failed `ENOTSUP` and discovery self-disabled for the process lifetime (`bind ENOTSUP 0.0.0.0:5353` before, advertising after).
- **Timeout guards leaked pending promises under Bun** (`destroy(err)` emits no `'error'`): the ffmpeg download and lyrics lookup hung forever instead of failing, and the boot scan chains behind ffmpeg. Measured: never settled vs rejected in 505 ms.
- Unguarded spawn `'error'` at the ssl-test and image-compress workers (a spawn failure was a dead server); a leaked SQLite handle per soft reboot; an unguarded federation bridge socket.

### Refuted on verification

The audit flagged `supervision.js`'s console-loss watchdog as inert under Bun. It reproduces only for `console.log`, which Bun routes around `process.stdout.write` — the server logs through winston's `console._stdout.write`, which **does** emit `'error'`. Verified end to end on the real server at the unmodified commit: self-terminates with code 74 in ~3.6 s. No change made; the proposed wrapper would have been overhead on every write.

### CI

The reboot smoke could pass **having rebooted nothing** — a non-200 trigger still set `REBOOT_OK=1` (proven live). It now asserts the trigger, requires a second serve line, and exercises overlapping reboots via `/config/address` (unlike `trust-proxy`, it has no unchanged-value short-circuit — with a no-op second POST the test passes against the bug). Added a darwin smoke of the app-launch flow itself, which had **zero** coverage despite being this PR's headline, and where `LSUIElement` makes any regression invisible.

Full suite 2051 pass / 0 fail; darwin bundle smoke green; overlapping reboots verified on the real `.app`.

PR #805, opened 2026-08-04
feat(federation): per-key bandwidth limits + key expiry (V62)

IrosTheBeggar:
PR 1 of the discovery→federation migration arc (plan: federation requests over the discovery network; this lands the abuse-control half first since it hardens existing federation on its own). Bandwidth limits + key expiry, **one V62 migration** (the interim V63 was folded in before merge — commit 3 — since the branch never shipped).

## Bandwidth limits

Three per-key caps on federation keys, enforced at the auth wall where every `x-federation-key` request already flows. **0 = unlimited on all three, and existing keys migrate to unlimited — upgrades change nothing until an admin sets a limit.**

- **`stream_kbps`** — token-bucket rate cap on `/media` responses, bucket **shared across the key's concurrent streams**. Implemented by wrapping `res.write`/`res.end` (compression-middleware technique) with strict FIFO dispatch and pipe-compatible backpressure (`false` + `'drain'`), so throttled bodies stay byte-identical.
- **`daily_mb`** — per-UTC-day transfer quota over everything the key pulls. Over quota → **429 + `Retry-After`** (seconds to UTC midnight) on byte-heavy routes **only**: browse/health keep answering so the peer's UI can say *why* playback stopped. An in-flight response is never cut (overshoot ≤ one file, documented).
- **`max_streams`** — concurrent `/media` response cap → 429.

Accounting is an in-process accumulator flushed to `federation_key_usage` every ~15s (no DB write per request). Quota checks read DB baseline + unflushed remainder, so bursts can't outrun the flusher and restarts can't forget a day. One row per key per UTC day, pruned at 90 days — the table's ceiling is keys×90 rows regardless of traffic.

## Key expiry

`federation_keys.expires_at` (NULL = never; every existing key stays that way). **One semantic: past the cutoff the key fails the auth wall AND the iroh pipe handshake** — it time-boxes a friend's access and quietly retires never-redeemed tickets. Details that matter:

- The expiry check runs **before the TOFU block** — a dead ticket can't bind on its way out (proven in the handshake suite).
- The `expired` flag is **computed in SQL** against `datetime('now')` on every key read — never via JS `Date` parsing, which reads SQLite's format as local time and would drift the cutoff by the UTC offset.
- An already-open pipe is **lazily severed** on the key's first post-expiry request.
- The limits route treats `expiresAt` as tri-state (absent = unchanged, null = never, future ISO = set/renew) — renewing an expired friend is one edit, and saving bandwidth tweaks can't silently restart the clock. The key row and usage history survive expiry.
- Tickets carry expiry as an informational **`e` field** (old parsers ignore unknown fields by design; the server row is the truth). The add-peer preview shows "valid until …". Spec updated in `docs/federation-ticket.md`.

## Surface

- Config `federation.limits` (8000 kbps / 2048 MB / 3 streams) — **pre-fills the mint dialog only**; per-key values live on the row, enforcement never reads config.
- Mint accepts optional limits + `expiresAt`; new `POST /api/v1/admin/federation/keys/:id/limits` edits both live; key list exposes limits, live `usage_today_bytes`, `expires_at`, `expired`. openapi documented.
- Admin UI: mint modal (three limit fields + Expires-days, "0 = unlimited/never"); keys table Limits ("8 Mbps · 2 GB/day · 3 streams", "expires in 30d" / red "expired") + Today columns; edit modal with current-expiry display and renewal field.

## Testing

- Schema suite: fresh defaults (limits 0, expiry NULL), single-hop pre-V62 upgrade (unlimited AND never-expiring), usage UPSERT accumulation, cascade delete, SQL cutoff math incl. ISO normalization.
- `federation-limits.test.mjs` e2e over plain HTTP: config-default mints, live edits (+404/400), byte-identical throttled streams, **exact** quota math (4×256 KiB = 1 MiB → 5th pull 429), concurrency cap with slot release, flush-to-disk proof, and the full expiry lifecycle (400 on past dates → alive → 401 wall-wide → renewal → clear-to-never, ticket `e` round-trip).
- `federation-handshake.test.mjs` (real iroh): expired key rejected **without TOFU-binding**, renewal re-arms the same ticket.
- Regression: all federation + discovery-over-federation + schema suites green. Lint clean on changed files.
- UI click-through on a live preview server; fresh-boot verified `user_version = 62` with all four columns from the single migration.

⚠ If you booted an earlier commit of this branch against a real DB, delete that DB (or `PRAGMA user_version = 61`) before running this one — the interim V62/V63 stamps predate the merge.

### DIFF
--- a/docs/federation-ticket.md
+++ b/docs/federation-ticket.md
@@ -34,10 +34,16 @@ mstrfed<V>:<base64url(JSON payload)>
   "k": "fedk_…",           // REQUIRED — the minted read-only API key
                            // ('fedk_' + 32 random bytes base64url)
   "n": "Paul's mStream",   // optional — display name for the add-peer UI
-  "l": ["Music", "Vinyl"]  // optional — granted library (vpath) names.
+  "l": ["Music", "Vinyl"], // optional — granted library (vpath) names.
                            // Informational preview only: the live grant
                            // list comes from GET /api/v1/federation/health
                            // after pairing.
+  "e": "2026-09-03T12:00:00Z" // optional — ISO expiry of the key.
+                           // Informational preview only: the minting
+                           // server's row is the truth, and an expired key
+                           // fails its auth wall + pipe handshake
+                           // regardless of what the ticket says. Absent =
+                           // never expires.
 }
 ```
 
@@ -78,6 +84,13 @@ payload missing `t` or `k`.
   every HTTP request, and any live connections. Rotating the server's
   `federation.secretKey` changes its EndpointId and invalidates the `t` in
   every issued ticket (peers must re-add).
+- **Expiry** is per-key and server-side (`federation_keys.expires_at`,
+  NULL = never): past the cutoff the key fails both the auth wall and new
+  pipe handshakes, and a pipe that was already open is severed on the
+  key's next request. An expired-but-never-redeemed ticket can no longer
+  TOFU-bind, which quietly retires lost tickets. The admin can renew
+  (set a new future date) or clear the expiry at any time — the key row
+  and its usage history survive expiry.
 - **"Friend reinstalled" case:** their endpoint identity changed, so TOFU
   rejects them. The admin UI's reset-binding action clears the binding
   without re-minting; the next successful handshake re-binds.
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -6071,6 +6071,8 @@ paths:
                     daily_mb: { type: integer, description: Per-UTC-day transfer quota. 0 = unlimited. }
                     max_streams: { type: integer, description: Concurrent /media response cap. 0 = unlimited. }
                     usage_today_bytes: { type: integer, description: Bytes served to this key today (UTC), live — includes traffic not yet flushed to the DB. }
+                    expires_at: { type: string, nullable: true, description: "UTC cutoff ('YYYY-MM-DD HH:MM:SS'); null = never. Past it the key fails the auth wall and the pipe handshake." }
+                    expired: { type: integer, description: 1 when expires_at has passed. The row survives expiry so the key can be renewed. }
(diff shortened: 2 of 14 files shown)


IrosTheBeggar:
**Two-server live smoke round** (real iroh pairing + the federation stream proxy, so every byte crosses B's fedFetch bridge and A's tunnel bridge before hitting the metered wall — the path the plain-HTTP integration tests can't reach): **14/14** after one fix.

```
PASS  mint: all six keys minted with tickets
PASS  pairing: five peers dial A over iroh and pass health (TOFU bound)
PASS  unlimited: 200 + byte-identical through B-proxy → iroh → A — 115ms
PASS  throttle: 300KiB at 800kbps is paced (≥1s) and byte-identical — 2116ms
PASS  quota: four pulls pass, fifth is 429 through the proxy — 5th=429
PASS  quota: health/browse still answers for the over-quota key
PASS  concurrency: first stream 200 + intact, parallel second 429
PASS  concurrency: slot frees after the first stream closes
PASS  expiry: short-lived key streams fine before its cutoff
PASS  expiry: past the cutoff the stream 401s and health fails
PASS  expiry: renewal on A revives the same ticket for B
PASS  pre-expired ticket: iroh handshake rejected, key never TOFU-bound
PASS  usage: A reports per-key usage_today_bytes covering every pull
PASS  logs: A warned about quota, concurrency, and expired-key rejections
```

Found + fixed (`dc9d0ea5`): the expired-key **lazy pipe sever raced its own 401** — cutting the iroh connection in-line with the rejection killed the response mid-bridge, so the peer saw 502 "peer unreachable" instead of the clean auth error. The sever is now delayed 2s (unref'd): the 401 flushes, requests in the window still die at the wall, and the coasting pipe goes right after. Re-verified live plus the limits/handshake suites (13/13).

PR #812, opened 2026-08-05
perf(dlna): stop enumerating network adapters per DIDL item, memoize the browse surface, bound Search

IrosTheBeggar:
Audit PR-H — the DLNA findings from the [2026-07-30 perf-landmine audit](https://github.com/IrosTheBeggar/mStream/pull/792) (M6, M10, M14 + low-hangers), plus the mechanism the audit missed, which turned out to dominate every one of them.

Follows #792 (A), #793 (D), #794 (F), #809 (C), #810 (B), #811 (E).

## The thing the audit missed

The audit measured DLNA at the SQL/JS level. This PR's rig drives **real SOAP over HTTP** against `dlna.setup()`'s own route table, and that surfaced the actual dominant cost:

`trackItem()` builds a media URL for every track and `containerArtXml()` an art URL for every container. Both call `getBaseUrl()` → `getLocalIp()` → **`os.networkInterfaces()`**, which costs ~13 ms on Windows. The browse surface was paying that syscall **once per DIDL element**.

Measured on the 25k-track audit fixture (8,000 `user_metadata` rows), `RequestedCount=0`:

| | wall | event loop blocked |
|---|---|---|
| `Search '*'` | **61,276 ms** | **61,181 ms** |
| `tracks-N` browse | 57,101 ms | 57,034 ms |
| `favorites` browse | 11,073 ms | 22,599 ms |
| `albumartists-N` browse | 4,417 ms | 7,118 ms |

**One unauthenticated SOAP request pinned the server for a full minute.** DLNA is opt-in, but once on, the control surface takes no credentials.

`getLocalIp()` now memoizes for 5 s — long enough to collapse a 10,000-item response into a single lookup, short enough that a DHCP renew, VPN flip or adapter swap still self-heals within seconds. An explicitly configured `address` bypasses both the lookup and the cache, exactly as before.

## Browse memos (M10 + the dirs-mode low-hanger)

Library *shape* — artists, album-artists, albums, genres, track count, and the sorted filepath list — only changes when a scan changes the library. That is precisely what `SystemUpdateID` already signals to control points, so the memo keys on it, with:

- a 60 s TTL as a **safety net**, not the invalidation (a writer that skips the bump goes stale for a minute, not forever);
- an LRU entry cap — the filepath list is the only large entry, and retaining one copy per browsed library costs what the uncached code allocated and threw away on *every click*;
- an explicit clear on every bump, so those lists are released immediately.

Smart-container totals aggregate play history and ratings, which never move `SystemUpdateID`, so they get a 5 s window instead — enough to collapse the nine aggregates one root-container click fires.

## Bounded responses (M6, M14)

**Search** picks a 500-item page when `RequestedCount` is 0 — UPnP leaves that choice to the server, `SearchCriteria` of `'*'` matches every track on the whole server, and Search is only ever issued by control points that page. `TotalMatches` stays truthful, and an explicit `RequestedCount` is still honoured up to `MAX_BROWSE_COUNT`.

**Browse is deliberately left alone.** Renderers drill into a container expecting its listing and some don't page, so its ceiling is unchanged — a 10,000-item Browse is still ~320 ms, down from 57 s.

**Favorites** joins its siblings at `SMART_LIMIT`. It was the odd one out: an uncapped count paired with a `LIMIT -1` fetch. ⚠️ **Behaviour change** — a library with thousands of 4-star tracks now shows 200 in that container, matching Recently Played / Most Played / Shuffle.

## Also

- `xmlEscape` probes once for the four strip classes instead of always running them, then escapes the five entities in a single pass. Output is byte-identical — pinned against the verbatim nine-pass original over 6,174 hostile strings (control chars, both lone-surrogate halves, noncharacters, valid astral pairs, all five entities). ~3× on per-field escaping, ~13% on the document pass.
- **V64** indexes `tracks.year` (partial, `WHERE year IS NOT NULL`) — the By-Year surface full-scanned three ways.
- `getAlbumArtistById` / `getAlbumArtistAlbums` branch on `artistId === 0` instead of `COALESCE(al.artist_id, 0) = ?`, so `idx_albums_artist` applies.

## After, same rig

| | before | after | |
|---|---|---|---|
| `Search '*'` count=0 | 61,276 ms | **43 ms** | 1,412× |
| `tracks-N` browse count=0 | 57,101 ms | 327 ms | 175× |
| `favorites` browse | 11,073 ms | 32 ms | 351× |
| `albumartists-N` browse | 4,417 ms | 9 ms | 496× |
| `mostplayed` browse | 2,646 ms | 41 ms | 65× |
| `year-N` browse | 1,866 ms | 12 ms | 154× |
| `recentplayed` browse | 1,523 ms | 37 ms | 41× |
| `lib-N` browse (M10) | 222 ms | 5 ms | 43× |
| `music` root browse (M14) | 181 ms | 2 ms | 86× |
| `folders-N` browse | 90 ms | 23 ms | 4× |
| 5-click renderer session | 595 ms | 44 ms | 14× |

## Deliberately not done

The audit's proposed um-driven UNION-ALL rewrite of the smart-container hash joins. I measured it on the same fixture: **66 → 59 ms, ~10%** — not the O(plays)-vs-O(tracks) win it was expected to be. `user_metadata` is only 3× smaller than `tracks` there, and the rewrite adds a temp b-tree GROUP BY plus a second correlated probe back into `um`. Row counts matched exactly (866 / 4820 / 6329), so the rewrite is correct — just not worth the risk. The memo is the real fix. Recorded here so it isn't re-chased.

## Validation

- **New suite** `test/db/dlna-surface.test.mjs` — 20 cases: V64 index + plan, the `getLocalIp` memo, the escape equivalence sweep, memo hit/TTL/bump-invalidation, the Favorites cap (with 260 rated tracks), and the Search bounds incl. a full paging walk. 20/20.
- `test/integration/dlna.test.mjs` 77/77 · schema/migration ring 59/59 · `npm test` **2136 pass / 0 fail**.
- Lint clean on all changed files (the one `ssdp.js` `no-empty` is pre-existing, just shifted).
- **Live smoke, 16/16** — one booted server, DLNA same-port, 6,000-track second library, 400 rated tracks, driven by raw SOAP with **no credentials** while a client-side pinger watches `/api/` latency. Covers the bounded Search + a paging client reaching all 6,009 tracks, a 40-click `lib-N` burst (worst warm click 21 ms), a 40-click root burst (9 ms), the Favorites cap, By-Year, the full root→lib→folder→track drill, the advertised media URL streaming real bytes, and **a real `force-rescan` proving the memo cannot serve a stale listing** (childCount 8 → 0 as the rescan purged fileless rows).

One smoke assertion failed on the first run and was my harness, not the code: same-port DLNA puts `/media` behind the main auth wall — only separate-port mode serves media credential-free. Pre-existing and out of scope here.

## Remaining audit backlog

G (subsonic), I (wrapped).

### DIFF
--- a/src/api/dlna.js
+++ b/src/api/dlna.js
@@ -23,6 +23,16 @@ const subscribers = new Map();
 // consuming excessive memory (~5MB at 512 bytes/item and this cap).
 const MAX_BROWSE_COUNT = 10000;
 
+// UPnP lets the server pick the page size when RequestedCount is 0. Browse
+// keeps MAX_BROWSE_COUNT there — a renderer drilling into a container expects
+// its listing, and some don't page. Search is a different shape: it spans the
+// whole server rather than one container, `SearchCriteria` of '*' matches
+// every track, and the action is only ever issued by control points that do
+// page. Choosing a page rather than 10,000 items keeps one unauthenticated
+// SOAP request from assembling a multi-megabyte DIDL document. An explicit
+// RequestedCount is still honoured, up to MAX_BROWSE_COUNT.
+const DEFAULT_SEARCH_COUNT = 500;
+
 // Hard cap on concurrent GENA subscribers. Each entry is tiny but we still
 // don't want a malicious client to grow the Map without bound by SUBSCRIBEing
 // repeatedly without UNSUBSCRIBEing.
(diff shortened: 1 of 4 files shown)


IrosTheBeggar:
## Smoke round 2 — 23/23

Round 1 (16/16, in the PR body) had a blind spot worth recording: `test/helpers/server.mjs` pins `address: '127.0.0.1'`, and `getLocalIp()` early-returns on an explicitly configured address **before** the memo. So round 1 never exercised the headline fix — its speed came from the address short-circuit plus the Search cap, not from the change this PR is about.

Round 2 boots with `address: '::'` so the real `os.networkInterfaces()` path runs, on a 12,000-track library.

| leg | result |
|---|---|
| **A** — the memo on the real adapter path | 10,000-item Browse in **377 ms = 0.038 ms/item**, against ~13 ms/item pre-fix (~340×). Advertised host is a genuine LAN address (`192.168.1.71`), identical across all 10,000 items. Unauthenticated `Search '*'` 25 ms, bounded at 500, `TotalMatches` 12,009. |
| **B** — separate-port DLNA | Own port serves device.xml + SOAP; `getBaseUrl()` correctly advertises the **DLNA** port, not the API port; media streams credential-free (8,739 bytes) — the mode where that is by design. |
| **C** — description coherence | device.xml keeps every service URL **relative**, so no cached IP can desync it; and the host the DIDL advertises really does serve device.xml (200) and SOAP (200). |
| **D** — memo isolation | Two libraries return distinct view counts, unchanged when re-browsed in the opposite order (no key collision). Every memoised view still drills to real children, including the full genre → gartist → album → track chain. |
| **E** — admin `dlna.browse` change | First view flips `Folders` → `Genres` immediately; the memo keys on mode, so the setting is not masked. |
| **F** — `shareUserData: false` | Per-user containers stay out of the memoised root listing on first **and** repeat browse; direct browse to a gated container still 701s. |
| **G** — V64 upgrade path | A database rewound to v63 migrates to v64 with all 12,009 tracks intact, and the planner picks it: `SEARCH tracks USING COVERING INDEX idx_tracks_year (year=?)`. |
| **H** — concurrency | 8 concurrent renderers × 12 browses across the memoised endpoints, all well-formed, 610 ms total, loop max 120 ms. |

Two failures on the first pass of this round were both harness bugs, not product defects: my device.xml check assumed an absolute `<URLBase>` (there isn't one — the URLs are relative, which is correct UPnP), and leg E's title regex forgot the DIDL is XML-escaped inside `<Result>`.

**Honest remaining number:** the deliberately-uncapped 10,000-item Browse still blocks ~320–440 ms. That is the protocol-compatibility tradeoff called out in the PR body — down from 57 s.

PR #813, opened 2026-08-05
perf(db): index-friendly plans for the default-UI read paths

IrosTheBeggar:
The 2026-07 [perf-landmine audit](https://github.com/IrosTheBeggar/mStream/pull/792) findings in `src/api/db.js` and `src/api/search.js` that never got assigned to one of the lettered fix PRs — the deferred loose ends, minus the velvet-only one.

Follows #792 (A), #793 (D), #794 (F), #809 (C), #810 (B), #811 (E), #812 (H).

All numbers are **endpoint-level over HTTP** against the 25k-track audit fixture, not statement-level.

## The library filter is usually a no-op — so say so

`libraryFilter` now reports `coversAllLibraries`: whether its clause actually excludes anything. Two hot paths drop the clause when it doesn't, which is what lets SQLite satisfy their `ORDER BY` / `GROUP BY` from an index instead of probing `idx_tracks_library` across the whole visible set and sorting the result in a temp b-tree.

| | before | after |
|---|---|---|
| `POST /db/recent/added` limit=50 | 103.2 ms | **7.1 ms** |
| `POST /db/recent/added` limit=200 | 159.5 ms | **9.7 ms** |
| `POST /db/genres` | 57.9 ms | **3.7 ms** |

**The gate is the important part.** The obvious version of this fix — pinning the planner onto `idx_tracks_created_at` unconditionally — is a large *regression* for a scoped caller: a user who can see one small library of old tracks would walk the entire index to fill a page. Measured **0.05 ms → 13.7 ms**. So the shortcut fires only when the clause is provably a no-op, and the scoped path keeps its old plan (verified unchanged: 23.1 → 24.4 ms, 8.5 → 8.2 ms).

Compared as a set rather than by length — a federation key supplies `user.libraryIds` directly and could repeat an id or name a deleted one.

## artists-albums: the shape was the problem, not the predicates

Written as an OR of three album-id tests AND-ed to a tracks join, SQLite drove the whole query from tracks — a full `idx_tracks_library` scan per artist click, on the hottest default-UI path. Collapsing the three sources into one `UNION ALL` id set gives the planner a small candidate list to seek albums by, and visibility becomes an EXISTS probe per candidate rather than a driving scan.

**23.7 ms → 1.8 ms** (7.6 → 1.8 ms when scoped — this one helps every caller).

`UNION ALL`, not `UNION`: duplicates are harmless in a membership test and deduping them would cost a temp b-tree. The outer `SELECT DISTINCT` is retained — two album rows can share `(name, year, art)` when the same release is credited to two artists, and the original collapsed those. Pinned by a test that builds exactly that fixture.

## search: EXISTS instead of a materialised id list

`id IN (SELECT … FROM tracks WHERE <scope>)` made SQLite materialise every visible `artist_id` / `album_id` in the library *before* FTS had narrowed anything down — a whole-library pass on every keystroke. EXISTS runs per candidate, and FTS only ever emits 30.

Statement level: **24.5 → 1.6 ms** (artists), **27.8 → 2.9 ms** (albums). Whole `/db/search` endpoint: 736.7 → 498.9 ms — it does considerably more than these two queries, so this is the honest share.

## genre-songs: optional paging

A dominant genre returns the whole library in one response (**10,500 rows / 8.05 MB** for a 42%-share genre at 25k tracks, audit M8) and there was no way for a caller to ask for less. `limit`/`offset` are optional and default to everything, so the existing contract is untouched.

⚠️ **This is the server half only — nothing gets faster until a client opts in.** With a page: 553 ms / 8.05 MB → **48.8 ms / 0.38 MB**.

**Deliberately not done:** a default cap. "Play this whole genre" is the point of the endpoint; silently truncating it would be a functional regression dressed up as an optimisation.

## Two compatibility bugs caught in self-review

Both were mine, in the first commit, and both are now fixed and pinned:

- the new Joi schema was strict, and this route has **never had one**. It's reachable from clients outside this repo — the Flutter app, and federated peers via `federation-auth`'s allow-list — so a strict schema would have started 400ing bodies that used to be accepted. Now `.unknown(true)`.
- paging read `req.body` rather than `joiValidate`'s coerced value, so `limit: "50"` would have been bound into `LIMIT` as text.

## Validation

- **New suite** `test/db/db-read-paths.test.mjs` — 26 cases. Routes are mounted on a bare express app so `req.user` can be driven directly; per-user library visibility is the whole design and an integration harness can't vary it cheaply. Covers the gate itself (including a stale federation id), oracle comparisons against the verbatim pre-PR queries for recent/added, getGenres and artists-albums, the DISTINCT-collapse case, scoping on every path, and the paging contract.
- Affected integration ring 45/45 · `npm test` **2152 pass / 0 fail** · lint clean.
- **Live smoke, 17/17** — one booted server, two *real* users (one all-access, one single-vpath), 8,000 seeded tracks that are the newest in the DB. Proves the fast path is fast, the scoped user never sees the hidden library through any of the four endpoints, `ignoreVPaths` turns the shortcut off, the shortcut's genre counts match the database, paging walks the whole set, and the event loop stays responsive (max 27 ms) under a 75-call browse burst.

## Not in scope

The remaining audit items in these files are watch-list tier and the audit's own measurements argue against fixing them: `search.js:204` lyrics LIKE fallback (rare path), `random.js:388` (proposed fix measured at only ~28%), `manager.js:558` (Last.fm opt-in), `admin.js:1819` (3 ms today). The velvet `/db/genre/albums` raw-bind nit is left alone — that UI is slated for removal.

Remaining audit backlog after this: **G** (subsonic) and **I** (wrapped).

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1375,6 +1375,11 @@ paths:
     post:
       tags: [Library]
       summary: List tracks in a genre
+      description: >
+        Returns every track in the genre by default. `limit`/`offset` are
+        optional; omitting both preserves the full-list response. Worth using
+        on a dominant genre, where the unpaged response can span most of the
+        library.
       requestBody:
         required: true
         content:
@@ -1385,6 +1390,15 @@ paths:
                   required: [genre]
                   properties:
                     genre: { type: string }
+                    limit:
+                      type: integer
+                      minimum: 1
+                      maximum: 10000
+                      description: Page size. Omit for the whole genre.
+                    offset:
+                      type: integer
+                      minimum: 0
+                      description: Rows to skip. Usable without `limit`.
                 - { $ref: "#/components/schemas/IgnoreVPaths" }
       responses:
         "200":
--- a/src/api/db.js
+++ b/src/api/db.js
@@ -524,6 +524,14 @@ export function setup(mstream) {
     // temp b-tree. The outer DISTINCT is retained from the original — two
     // distinct album rows can share (name, year, art), e.g. the same album
     // present in two libraries, and the original collapsed those.
+    //
+    // The name tiebreak exists because `ORDER BY al.year DESC` alone leaves
+    // same-year order to the plan, the old and new plans disagree on it, and
+    // the UI renders response order — without it, an artist's same-year
+    // albums would visibly reshuffle on upgrade (25k-fixture sweep: 200 of
+    // 653 artists). Pinning ties alphabetically is deterministic across
+    // plans and SQLite versions; the sort b-tree already exists, so it's
+    // free.
     const albumRows = d().prepare(`
       SELECT DISTINCT al.name, al.year, al.album_art_file
       FROM albums al
@@ -539,7 +547,7 @@ export function setup(mstream) {
             AND t2.album_id IS NOT NULL
       )
       AND EXISTS (SELECT 1 FROM tracks t WHERE t.album_id = al.id AND ${filter.clause})
-      ORDER BY al.year DESC
+      ORDER BY al.year DESC, al.name COLLATE NOCASE
     `).all(String(req.body.artist), String(req.body.artist), String(req.body.artist), ...filter.params);
 
     const albums = albumRows.map(r => ({
--- a/src/api/db.js
+++ b/src/api/db.js
@@ -637,13 +637,20 @@ export function setup(mstream) {
     // audit M8) and there was previously no way for a client to ask for a
     // page. This is the server half only — nothing gets faster until a
     // caller opts in.
+    // `.unknown(true)` on purpose: this route has never had a schema, and it
+    // is reachable from clients this repo does not contain (the Flutter app,
+    // federated peers — see federation-auth's allow-list). Rejecting fields
+    // that used to be ignored would be a breaking change smuggled into a
+    // perf PR. The new paging fields are still validated.
     const schema = Joi.object({
       genre: Joi.string().required(),
       limit: Joi.number().integer().min(1).max(10000).optional(),
       offset: Joi.number().integer().min(0).optional(),
       ignoreVPaths: Joi.array().items(Joi.string()).optional()
-    });
-    joiValidate(schema, req.body);
+    }).unknown(true);
+    // Use the COERCED value, not req.body: Joi turns `limit: "50"` into a
+    // number, and binding a string into LIMIT would be a different query.
+    const { value: query } = joiValidate(schema, req.body);
 
     const filter = libraryFilter(req.user, req.body?.ignoreVPaths);
 
@@ -668,10 +675,10 @@ export function setup(mstream) {
 
     // SQLite treats LIMIT -1 as "no limit", and OFFSET is meaningless without
     // one — so the unpaged call keeps exactly its old plan and result.
-    const pageSql = (req.body.limit != null || req.body.offset != null)
+    const pageSql = (query.limit != null || query.offset != null)
       ? 'LIMIT ? OFFSET ?' : '';
     const pageParams = pageSql
-      ? [req.body.limit ?? -1, req.body.offset ?? 0] : [];
+      ? [query.limit ?? -1, query.offset ?? 0] : [];
 
     const rows = d().prepare(`
       ${trackQuery(req.user?.id, { includeGenres: false })}
--- a/src/api/db.js
+++ b/src/api/db.js
@@ -172,10 +172,30 @@ export function libraryFilter(user, ignoreVPaths) {
     libIds = libIds.filter(id => !ignoredIds.has(id));
   }
 
-  if (libIds.length === 0) { return { clause: '1=0', params: [] }; }
+  if (libIds.length === 0) { return { clause: '1=0', params: [], coversAllLibraries: false }; }
+
+  // Does this filter actually filter anything? When the caller can see every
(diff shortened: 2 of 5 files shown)


IrosTheBeggar:
## Post-open audit

Re-reviewed the full diff plus everything it touches; ran an independent at-scale sweep (25k fixture, old query text vs new, **14/14 checks**).

**Verified safe:**
- **Gate consistency under mutation** — `getAllLibraries()` is the in-memory cache; `invalidateCache()` clears users+libraries+userLibraries together, and both mutation paths order correctly: `addDirectory` invalidates **before the scanner can insert a track** for the new library (no window where a restricted user's `coversAllLibraries` stays true while invisible tracks exist), and `deleteLibraryRows` drains tracks **before** deleting the library row (no orphan-track window for the unscoped path; crash mid-drain leaves the row, so scoping still applies on resume).
- **Equivalence at scale** — old-vs-new identical result sets for 653 artists (599 reached via `album_artists`/`track_artists` side doors, seeded — the fixture shipped with empty M2M tables), plus a scoped variant (64 artists actually narrowed); `recent/added` identical id sequences at LIMIT 50/200/1000; getGenres fast path byte-identical; 63 FTS IN-vs-EXISTS probes identical (deterministic tiebreak added to both forms so rank ties can't masquerade as filter differences); genre-songs page-concat == unpaged with plan equality under LIMIT.
- Plans: fast paths order from `idx_tracks_created_at` (no temp b-tree), genres path never touches `tracks`, artists-albums no longer drives from `idx_tracks_library`. All 34 `libraryFilter` call sites consume `.clause`/`.params` positionally — the third field is invisible to them. Joi errors map to **400** in live code (server.js:446).

**Found & fixed (b26fd14d):** the rewrite changed the plan, and `ORDER BY al.year DESC` alone leaves same-year order to the plan — 200 of 653 artists returned identical album *sets* in a different within-year *order*, and the default UI renders response order (no client-side sort). Pinned ties with `al.name COLLATE NOCASE`: deterministic across plans/SQLite versions, measured free. New pin test (27 cases now); affected ring re-run green.

**Noted, accepted:** genre-songs with a missing or non-string `genre` now 400s (was a silent `String()` coercion → usually `[]`); no client sends either shape. `artists-albums-multi` (velvet-stubs) still has the old OR-of-three — out of scope per the no-velvet instruction, and velvet is slated for removal.

Live-server smoke round next.

IrosTheBeggar:
## Smoke round 2 — 31/31

One real booted server (testlib fixtures + 8k-track seeded biglib), two real users over authenticated HTTP. **Oracle-driven**: every recent/added assertion compares the API's exact row sequence against the same query run straight on the database — fast path, scoped path and flipped-gate states are proven by row identity.

New coverage this round (round 1 never touched runtime mutation):

- **Live vpath revoke/regrant** — revoking `biglib` from the covers-all user scopes the *same token* on the very next request (== testlib oracle); regranting restores the fast path (== both-libs oracle). Proves the per-request vpath re-derivation + `invalidateCache` chain the gate depends on.
- **Live add-directory** — `PUT /admin/directory` with a real MP3, real scan. The new library's track becomes the newest row in the whole DB and the ex-covers-all user still doesn't see it (gate correctly OFF, ungranted); granting the vpath flips covers-all back on and that row leads recent/added. The scoped user is untouched throughout.
- **Paging law** — pages tile the unpaged sequence exactly; limit-only == head; offset-only == tail; offset past the end → `[]` not an error.
- **Client tolerance + 400 contract** — Flutter-shape body (unknown fields + string `limit`) → 200 with coerced paging; limit 0 / limit 20000 / offset −1 / missing genre / numeric genre → 400 with `{error}` body.
- **Determinism (b26fd14d)** — five identical artists-albums responses; year DESC dominates, NOCASE name breaks the same-year tie.
- **Event loop** — 21.6 MB of full-genre responses (3k rows each), max /api/ ping 183 ms.
- **Default-UI click-through** on a live preview — artists → albums (year-DESC render), genres with counts, recently added, genre → songs; zero console errors.

Timings live: recent/added 9 ms first / 11 ms worst-of-20, genres 4 ms, artists-albums 8 ms, search 7 ms.

Full suite on the tiebreak HEAD: **2159 tests, 2153 pass, 0 fail, 6 skipped**. Platform CI green on b26fd14d.

IrosTheBeggar:
**CI failure diagnosed — a flake in a pre-existing suite, not this PR's code.**

The failing test was `discovery-infra-churn.test.mjs` (PR #809's H5 suite): *"safety net: unpublished drift rebuilds once the debounce window passes."* Its inside-the-window probe had an **80 ms budget** (build → upsert → getIndex before the shortened debounce window elapses), and the windows-latest·3/3 runner was demonstrably crawling (a block that takes ~1 s locally took 14.7 s there) — so the safety-net rebuild fired *early*, which is correct product behaviour the test margin didn't allow for. The same file passed on all three ubuntu and all three macOS shards; this PR touches neither discovery nor that suite.

Deflaked in a0015cc0: window 80 → 1000 ms, and the probe is now elapsed-guarded — if a runner genuinely overshoots the window, a rebuild is right, and the cached-inside-the-window property is independently pinned time-free by the published-epoch test above it. The unique claim (rebuild after the window) stays asserted unconditionally. Ran 3× locally, 13/13 each.

Same flake class as the 2026-07-28 Windows clock-skew lesson: timing assertions on spawned/CI processes need seconds-wide margins.

Issue #795, opened 2026-07-30
Desktop player: minimise to system tray — and where should desktop-player patches land?


Wired4ncer:
Hi — two things bundled, since the second one gates the first.

### The ask: minimise to tray in the desktop player

On Linux the desktop player has no way to get out of the window list. Closing it stops playback, so a long listening session means it permanently occupies a taskbar slot. Minimising to the system tray is the usual answer, and it's also handy on Windows.

Right now I'm working around it with `kdocker`, which wraps the window and hides it to the tray. That works, but it's an external dependency, it needs a per-user config key set by hand to make the minimise button (rather than only the tray icon) route to the tray, and it can't show anything about what's playing.

### Why this looks like a small change

Unless I'm reading the shipped build wrong, the player's Electron main process is a single 30-line file. From the `app.asar` of the AppImage I'm running (`X-AppImage-Version=6.0.0`, whose `app-update.yml` points at this repo), `/index.js` is:

```js
const { app, BrowserWindow } = require('electron')

function createWindow () {
  const win = new BrowserWindow({ autoHideMenuBar: true, backgroundColor: '#1e2228', width: 1200, height: 800, webPreferences: { contextIsolation: true, nodeIntegration: false } })
  win.loadFile('./index.html')
}
...
```

There's no `Tray` in it (`new Tray` appears nowhere in the player's asar), so this would be additive rather than a refactor:

- create a `Tray` on ready, with a context menu (Show / Quit)
- `win.on('minimize', e => { e.preventDefault(); win.hide() })`
- tray `click` → `win.show()`
- keep it behind a preference if you'd rather not change default behaviour

Two things I'd flag as the fiddly parts rather than pretend they're free:

- **Tray icon assets.** The asar already ships `velvet/build/tray-icon.png` / `tray-icon-osx.png`, which look vestigial from the server-mode desktop build — those could probably be reused.
- **Linux activation is genuinely ambiguous.** Per the Electron docs, `click` fires on "activation" but the StatusNotifierItem spec doesn't define what activates: single left click in some environments, double click in others. So a tray whose *only* way back is a left click will feel broken for some users — the context menu "Show" entry needs to be the reliable path, not a nicety.

### The blocking question

[...]

PR #817, opened 2026-08-08
feat(desktop): mStream tray launcher crate (phase 1b)

IrosTheBeggar:
Phase 1b of the binary-releases roadmap: the tray launcher as a standalone crate. Phase 1c (separate PR) wires it into the bundles — staging via build-bun.mjs, the `mStream.exe`/`.app`-executable rename, `mstream-server` rename, mac-app-launch.js retirement, and the bundle smoke.

## What this is

`rust-launcher/` — "mStream Desktop", the future double-click entrypoint. One binary, two faces:

- **Double-click / login item** → spawns `mstream-server --supervised` in the background (no console flash) and puts an icon in the tray: **Open mStream · Quick Connect · Start at login (checkbox) · Restart server · Quit**. The launcher holds the server's stdin; the server's existing supervision contract (stdin EOF → clean shutdown) makes orphaned servers structurally impossible — even a force-killed launcher takes its server down.
- **Terminal** → transparent pass-through to the server binary: same argv, real console stdio (`CONIN$`/`CONOUT$` on Windows, `exec()` on unix), mirrored exit code. Terminal users keep exactly today's CLI, including `-j`, `--portable`, and `--quick-connect-off-by-default` (#816).

## Decisions implemented (from the approved plan)

- **Autostart on by default**, one-click off: configured exactly once on first run (marker in `launcher.json`), after which the user's tray toggle — or the scriptable `--autostart=enable|disable|status`, which the installer/uninstaller will reuse — is never overridden. Login-item launches pass `--autostarted` and come up silent (tray only, no browser tab).
- **Quick Connect as the standard connection**: the tray item deep-links to `/#quick-connect`, and the webapp change here opens the existing Quick Connect modal on that hash (no-op on servers that don't share the code).
- **Single management model**: `paths.rs` deliberately mirrors `src/util/boot-config.js`'s config ladder (explicit `-j` → `MSTREAM_CONFIG` → legacy/portable next-to-binary → per-OS data home) so launcher and server always agree on config, port, and data. Cross-referenced comments on both sides.
- The server child's env drops `__CFBundleIdentifier`, so when the macOS `.app` runs the launcher as its bundle executable, bundle-launch markers never leak into the managed child.
- Fatal paths alert visibly (MessageBox / osascript) — a desktop launch must never die silently (#802's lesson, applied to the launcher that will replace that fix).

## Verification (Windows, debug build, against a real server binary)

- Console pass-through: `--version` answered by the server, worker exit codes mirrored.
- Autostart: `--autostart=enable` wrote `HKCU\...\Run\mStream = "...mstream-launcher.exe --autostarted"`; `disable` removed it (verified via `reg query`, left clean).
- Tray mode: server spawned and serving in ~2s, tray icon constructed, lock file held; a second instance exited 0 after logging that it yielded to the running one.
- Quick Connect end-to-end: `/api/v1/iroh/code` returned `shared:true, available:true` with a real pairing code under the launcher-managed server, and a browser at `/#quick-connect` had the modal open with the QR rendered (DOM-asserted).
- Orphan prevention: `taskkill /F` on the launcher alone → the server exited on the closed supervision pipe and port 3000 went connection-refused.

## CI

`build-rust-launcher.yml`: release build + `clippy -D warnings` on windows/macos/ubuntu for PRs touching the crate (Linux deps: gtk3 + ayatana-appindicator + xdo). Build-only by design — artifact staging belongs to 1c. Note: as a brand-new workflow file it may not trigger on this PR itself (same GitHub quirk build-bun.yml hit in #670); it's wired for every PR after merge, and I've validated build + clippy-clean locally on Windows.

## Review notes

- Linux tray reliability follows the #795 discussion: the context menu is the reliable interaction (StatusNotifierItem "activation" is ambiguous across DEs), and a missing tray host degrades gracefully — the server keeps running and the fact is logged.
- muda toggles a `CheckMenuItem` before the event fires, so the handler reads `is_checked()` as the *desired* state and reverts it if the OS registration fails.
- The exit watcher polls `try_wait()` rather than blocking `wait()` (which would hold the child mutex across the quit path), and both helper threads are generation-stamped so a watcher from before a Restart can't misreport the replacement child.


IrosTheBeggar:
## macOS smoke (arm64) — PASS

Runtime smoke on real hardware — macOS 26.5.2, Apple Silicon, Rust 1.94.1, release build — complementing the Windows verification above and the CI leg (which is build + clippy only). Server: real darwin-arm64 binary built via `bun scripts/build-bun.mjs --bundle`, staged in the phase-1c shape (`mstream-launcher` + `mstream-server` siblings, webapp/ + bin/ + iroh `.node`).

**Console pass-through (unix `exec()` path)**
- `--version` under a real pty answered by the server (6.19.3, exit 0); a failing boot (`-j /nonexistent`) mirrors the server's exit 1
- No sibling binary → clear "no server binary… (or set MSTREAM_SERVER_BIN)" message, exit 1; both `MSTREAM_SERVER_BIN` and `--server-bin` overrides work
- A no-tty launch dispatches to the desktop face (the Finder double-click shape) — verified by running without `--tray`

**Tray mode**
- Server spawned and serving in ~2 s; menu-bar icon rendered (visually confirmed against `build/icon.png`); `server is up` logged; no "tray unavailable"
- Single instance: second launcher exited 0 after logging that it yielded to the lock holder
- Launcher/server agreement: the server wrote conf/db/caches to `~/Library/Application Support/mStream` — exactly the config path the launcher logged; nothing written next to the binaries

**Orphan prevention**
- `kill -9` and SIGTERM on the launcher both took the server down within ~3 s; the server log shows `[supervision] supervisor closed stdin — shutting down`, and port 3000 went connection-refused

**Autostart (LaunchAgent)**
- `--autostart=enable` wrote `~/Library/LaunchAgents/mStream.plist` with `ProgramArguments = [<launcher path>, --autostarted]` + `RunAtLoad`; `status` tracked it; `disable` removed it and left the directory clean; unknown subcommand exits 2. The plist is file-only (never `launchctl load`ed), which is right for login-time semantics
- First-run policy (isolated `$HOME`): the first tray run enabled autostart and wrote the `autostartConfigured` marker; after a user disable, a relaunch did **not** re-enable

**Quick Connect end-to-end**
- `/api/v1/iroh/code` → `shared:true, available:true` with a real pairing code (`iroh.darwin-arm64.node` loads fine on darwin-arm64)
- Browser at `/#quick-connect` → modal open (`hystmodal--active`, DOM-asserted), 240 px QR rendered, top-bar button revealed

### Note: the headline results need #816

The branch predates #816 (merged yesterday). A server built from this branch alone diverges from the launcher: it writes `save/` next to the binary while the launcher resolves the data-home profile (both still default to port 3000, and the launcher converges to the legacy path on the next run once `save/` exists), and Quick Connect stays dark (`iroh.enabled` defaults false pre-#816). Everything above was verified against this branch **with master merged in**. Worth a rebase/merge before 1c — that also makes `paths.rs`'s cross-references to `src/util/boot-config.js` / `test/unit/boot-config.test.mjs` resolve on the branch.

### Residual gap

Clicking the tray menu items themselves (Open / Quick Connect / Restart / Quit) couldn't be automated here (osascript assistive access is TCC-denied in the test shell). Menu construction was verified visually, the deep-link target by direct navigation, and the handlers are the shared muda path already exercised on Windows — one manual click-through on a Mac closes the gap. Also minor: the iroh endpoint reported `online:false` (no relay connection during the test window) — irrelevant to the modal, which gates on `available`.

PR #818, opened 2026-08-08
fix(waveform): full-length decode, no phase cancellation, real Opus coverage

IrosTheBeggar:
Three correctness bugs in waveform generation, found while auditing the
generator. Each is reproduced by a fixture in the test suite.

## 1. Waveforms were cut short and time-misaligned

The rust decoder binned bars against the container's declared frame count
— and symphonia's MPEG reader also *ends the packet stream* at that
count. With no Xing/Info header the count is extrapolated from the
**first frame's bitrate**, so a VBR file that opens denser than its
average loses its tail.

Measured on a 60 s fixture with bursts at 0/25/50/75/99.5 %:

| | bar positions | outcome |
|---|---|---|
| before | 1, 220, 441, 661 | last burst past bar 799, **dropped** |
| after | 1, 200, 400, 600, 795 | all present |

Same audio, same encoder, differing only in section order: opening loud
decoded **47.5 s** of 60 s, opening quiet decoded **60.0 s**. So roughly
10 % of the track was missing and every x-position disagreed with the
playhead.

Binning now runs through a bounded peak pyramid over the frames
*actually decoded*, and MPEG sources are wrapped in a `MediaSource` that
declines to state its length so the reader has nothing to extrapolate
from. That also retires the buffered path, which silently truncated
anything past ~12 minutes and could allocate 120 MB for a single file —
the pyramid is 512 KB per thread.

## 2. Out-of-phase stereo rendered as a blank bar

The ffmpeg fallback asked for `-ac 1`, which **sums** channels. An
anti-phase stereo fixture that the rust engine renders at 32/255
throughout came back as **800 zero bars**, and that got cached. Real
mid/side-widened masters lose energy the same way, partially.

Channels now stay interleaved, and both engines reduce a frame with
`max|ch|` — which never cancels, and makes the two engines describe the
same audio the same way. They previously disagreed by ~22 % on broadband
content (mean bar 58.6 rust vs 45.4 ffmpeg on pink noise).

## 3. Opus was reported as a permanent failure and never got a waveform

symphonia has no Opus decoder, so the pass wrote a `symphonia` failure
marker for every Opus track: a permanent failure count in the admin panel
for files that are perfectly fine, and no proactive waveform, so every
first play paid a cold ffmpeg decode behind an empty progress bar.

The pass now leaves codecs it structurally cannot decode out of its plan
entirely — no marker, honest totals — and a new ffmpeg half
(`src/db/waveform-fallback.js`) covers them, plus anything symphonia
genuinely did fail on, clearing markers it disproves. It runs from the
same task, holding the queue slot until both halves finish, so the
single-task rule still holds and the pass isn't "done" while work remains.

## Cache generation bump

All three fixes change the bytes produced for the same audio, so a fix
that didn't invalidate the cache would only ever apply to newly added
tracks. Artifacts move to `<hash>.w2.bin` / `<hash>.w2.failed`; v1 names
are neither read nor honoured, and are swept at boot from
`ensureCacheDir()` — the one path both rust-equipped and rust-less
deployments take.

The webapp's localStorage prefix moves to `wf2:` alongside, purging old
keys. Waveforms are cached client-side per filepath with no version in
the value, so without this the server-side fix would never reach anyone
who had already played the track.

**Accepted cost:** genuinely undecodable files lose their old `ffmpeg`
marker in the sweep and are retried once before it's rewritten.

## Testing

`npm test` green on the full tree — integration 680 tests / 678 pass /
0 fail / 2 skipped in 252 s, matching master's 249 s. New coverage:

- full-length decode across four MPEG header/bitrate shapes
- anti-phase stereo on both engines
- Opus excluded from the plan with no marker; symphonia markers still excluded
- the sweep's name discrimination (keeps current, drops v1, ignores bystanders)
- nine cases over the fallback: Opus pickup, marker retry-and-clear,
  ffmpeg-verdict not retried, dedup, vanished copies, abort

The fallback suite caught a real defect while being written: an
ffmpeg-failed Opus key arrives via the extension query rather than the
marker query, so gating only the marker path re-spawned a doomed 30 s
decode on every pass, forever.

`rust-parser/` is source-only here — `bin/rust-parser/*` is CI-managed.

## Merge sequencing note

`rust-parser/` is source-only here — `bin/rust-parser/*` is CI-managed and
rebuilt on push to master. Between this merging and that rebuild landing,
master runs new JS against a v1 binary: the pass would write `<hash>.bin`,
the boot sweep would delete those, and coverage would read 0 done. Nothing
breaks (the on-demand endpoint still generates and writes `.w2.bin`), but
the pass does wasted work each run until the binaries update, so it's worth
letting the rebuild land promptly after merge.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1118,7 +1118,7 @@ paths:
                         disabledReason:
                           type: string
                           nullable: true
-                          enum: [config, no-ffmpeg, no-api-key, no-binary, binary-unsupported, runtime-unavailable, null]
+                          enum: [config, no-ffmpeg, no-api-key, no-binary, binary-unsupported, binary-generation-mismatch, runtime-unavailable, null]
                         state:
                           type: string
                           enum: [idle, queued, running, disabled]
--- a/rust-parser/src/main.rs
+++ b/rust-parser/src/main.rs
@@ -44,6 +44,12 @@ const NUM_BARS: usize = 800;
 // are ignored here and swept at boot by the JS side.
 const WAVEFORM_EXT: &str = ".w2.bin";
 const WAVEFORM_FAILED_EXT: &str = ".w2.failed";
+// Deferred to the ffmpeg half: no decoder for this codec. A SEPARATE
+// artifact, not a line inside `.failed`, because every consumer
+// classifies these by filename — the JS coverage counter must stay
+// filename-only, and counting a deferral as a failure misreports a
+// perfectly good file. Mirrors DEFERRED_EXT in src/db/waveform-lib.js.
+const WAVEFORM_DEFERRED_EXT: &str = ".w2.deferred";
 
 // ── Config (matches what task-queue.js passes) ──────────────────────────────
 
@@ -4466,7 +4472,10 @@ fn load_waveform_cache_names(dir: &Path) -> HashSet<String> {
     if let Ok(entries) = fs::read_dir(dir) {
         for entry in entries.flatten() {
             if let Some(fname) = entry.file_name().to_str() {
-                if fname.ends_with(WAVEFORM_EXT) || fname.ends_with(WAVEFORM_FAILED_EXT) {
+                if fname.ends_with(WAVEFORM_EXT)
+                    || fname.ends_with(WAVEFORM_FAILED_EXT)
+                    || fname.ends_with(WAVEFORM_DEFERRED_EXT)
+                {
                     names.insert(fname.to_string());
                 }
             }
@@ -4549,18 +4558,16 @@ fn run_waveform_scan(json_str: &str) -> Result<(), Box<dyn std::error::Error>> {
         // for a transient quirk) must not stop this pass from trying, and
         // a generated .bin is exactly what unblocks the endpoint again.
         // Mirrors the engine-line discipline in src/db/waveform-lib.js.
-        // Both lines mean "this pass already decided about this key":
-        // `symphonia` is a decode failure, `no-decoder` is a codec we have
-        // no decoder for. Either way re-running would repeat the same work
-        // for the same answer, so both exclude the key from the plan. Only
-        // the ffmpeg fallback can move them, and a success there deletes
-        // the marker outright.
+        // A `symphonia` line means this pass tried and failed; re-running
+        // would repeat the work for the same answer. Only the ffmpeg
+        // fallback can move such a key, and a success there deletes the
+        // marker outright.
         let symphonia_failed: HashSet<&String> = existing
             .iter()
             .filter(|name| {
                 name.ends_with(WAVEFORM_FAILED_EXT)
                     && fs::read_to_string(cache_dir.join(name.as_str()))
-                        .map(|c| c.contains("symphonia") || c.contains("no-decoder"))
+                        .map(|c| c.contains("symphonia"))
                         .unwrap_or(true) // unreadable marker — assume ours
             })
             .collect();
@@ -4589,6 +4596,7 @@ fn run_waveform_scan(json_str: &str) -> Result<(), Box<dyn std::error::Error>> {
                 None => continue, // no content hash to key the cache on
             };
             if existing.contains(&format!("{}{}", key, WAVEFORM_EXT))
+                || existing.contains(&format!("{}{}", key, WAVEFORM_DEFERRED_EXT))
                 || symphonia_failed.contains(&format!("{}{}", key, WAVEFORM_FAILED_EXT))
             {
                 continue;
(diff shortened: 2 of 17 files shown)


IrosTheBeggar:
**Adversarial review completed and fixes pushed** (2687348a).

Six attack areas (rust decoder, cache sweep, fallback module, task-queue integration, JS/webapp, claims-vs-tests), every finding re-verified by a skeptic pass whose default was to refute: **12 confirmed, 6 refuted**. Verdict was *merge with fixes* — no blockers, nothing loses user data, but three majors shared one root cause in `generateOne`'s error classification. All confirmed findings are fixed in the new commit; each was reproduced against the unfixed head first.

**Majors fixed:**
- **Self-re-enqueue loop** — an unwritable cache dir made every round count phantom progress and chain forever, re-forking the full rust pass each time. The chain gate now requires *durable* progress (landed files/markers) plus a strictly shrinking backlog.
- **Disk errors recorded as ffmpeg's verdict** — a cache-write failure inside the decode `try` wrote a permanent `ffmpeg` marker for perfectly good audio, with no path that ever clears it. Persistence moved outside the try; failed writes count nothing and retry next pass.
- **Stale-binary delete/re-decode cycle** — the binary now answers `--wf-generation` and the server refuses the pass on generation mismatch (same pattern as the hash-generation gate). This also resolves the merge-sequencing note in the PR description: in the CI rebuild window the pass skips cleanly instead of doing wasted work, and both rust test suites probe the same way so labeled CI skips rather than fails.

**Minors fixed:** hash-shape guard on the sweep (no more deleting `pytorch_model.bin`), Opus-in-`.ogg` skipped silently + picked up by the fallback (was counted failed + marked), multi-copy walk (an unreadable copy no longer poisons a key with a good copy on disk), unique temp staging names, async `planWork`, and the overstated engine-parity comment corrected.

**Verification:** 40/40 across the three waveform suites (7 new regression tests), 899 tests / 0 failures on the scanner+unit+db shards, and a 13/13 end-to-end smoke against a booted server — all four codec cases (mp3, vorbis-ogg, bare opus, opus-in-ogg) got `.w2.bin` from the pass chain with zero failure markers, and the committed pre-w2 binary fails the probe exactly as the gate expects.

What the reviewers attacked and could not break, for the record: the PeakPyramid arithmetic, HiddenLenSource, the base64 webapp cache (zero findings), the endpoint-vs-fallback write race (benign by construction), and shutdown/kill-queue handling.

IrosTheBeggar:
**Round-2 adversarial review done, fixes pushed** (1ffa01ef).

Seven attack areas weighted toward the *fix commit* (newest, least-reviewed code), each finding re-verified by a skeptic defaulting to refute: **14 confirmed, 7 refuted**. Verdict was again *merge with fixes* — no blockers, no data loss — but round 2 caught two regressions the previous fix commit introduced, plus one defect from the first commit that round 1 missed.

**Regressions from 2687348a, now fixed:**

- **`NoDecoder` black-holed files.** Making it a fully silent skip assumed it only meant Opus-in-Ogg. It doesn't: symphonia's AAC decoder rejects *anything* not LC/≤2ch/no-SBR, so **5.1 AAC in `.m4a` and `.aac`** hit the same arm — and the fallback only routed `.opus`/`.ogg`. Those files got no waveform from either pass, ever, while the rust pass re-probed them on every scan forever. It now writes a `no-decoder` marker: not a failure line, but a durable trace that routes the key to ffmpeg whatever the container.
- **The generation gate disabled the ffmpeg half too.** It returned before the spawn, and the fallback's only call site is inside the rust child's close handler — so a stale prebuilt silently killed the Opus coverage this PR exists to add, on every deployment between merge and the CI rebuild. Both bails now run the ffmpeg half standalone.

**From the first commit, missed by round 1:** a momentarily unreadable file (ACL change, AV lock, stale mount) earned a **permanent** `ffmpeg` marker — excluded from every plan, endpoint 500s forever, cleared by nothing. `existsSync` is a stat, not an open. The walk now opens and reads a byte, and re-checks after a decode failure.

**Three of my tests would have passed on reverted code.** The worst was the cache-write regression test: it obstructed the final path with a directory named `<key>.w2.bin`, which `planWork` classifies as *cached* — so the run did nothing and every assertion passed vacuously. All three are fixed and mutation-checked (reverting each hunk now fails).

**Verification:** 902 tests / 0 failures (scanner+unit+db), 680 / 0 (integration, baseline parity). Two end-to-end smokes: the normal path (16/16) now covers 5.1 AAC in `.m4a`, 5.1 AAC in `.aac` and Opus-in-MP4 — all of which produced *nothing* before this commit — and a stale-binary run (13/13) proves Opus coverage survives with the rust half gated off.

Also added `binary-generation-mismatch` as a `disabledReason` so a sidelined rust half is visible in the admin panel rather than showing as enabled-and-idle.

One nice thing the smoke surfaced: `bare.opus`, `hidden-opus.ogg` and `opus.m4a` share a single `audio_hash`, so one cache file covers all three paths — V60 content hashing and the multi-path walk doing exactly what they should.

IrosTheBeggar:
**Round-3 adversarial review done, fixes pushed** (b5c10aa6).

Seven attack areas weighted at `1ffa01ef`, every finding re-verified by a skeptic defaulting to refute: **8 confirmed, 9 refuted. No blockers. Verdict: ship it.**

**The headline result: `1ffa01ef` introduced no regression** — the first fix commit in this PR that didn't. Reviewers checked each of its hunks against both `HEAD~1` and `origin/master` and proved the `no-decoder` route end-to-end on a real 5.1-AAC `.m4a` fixture: deferred by the rust pass, picked up by the ffmpeg half, marker cleared, rust re-run plans zero.

Two minors fixed, both cases where **the code didn't do what its comment claimed**:

- **A deferral was counted as a failure.** Routing `no-decoder` through a line inside `.w2.failed` meant the coverage counter — which classifies by filename alone — reported perfectly good 5.1-AAC and Opus-in-Ogg tracks as decode failures and subtracted them from `remaining`. That's the exact misreport this pass exists to remove, in a new form. Deferrals now have their own `<key>.w2.deferred` artifact, kept filename-classifiable on purpose: the coverage counter is the hot status path behind a prior production CPU-spin incident and must never read N marker bodies per poll. Verified: 1 cached + 1 deferred + 1 genuinely failed now reports `done=1, failed=1, remaining=1`.
- **The `binary-generation-mismatch` badge was unreachable.** The gate returns `enabled: true` (degraded, not disabled), but the admin UI only read `disabledReason` inside its `state === 'disabled'` branch — so the label was dead code and the operator saw a plain "Idle", indistinguishable from a healthy server, exactly in the window the gate exists for. The API side was always correct; only the UI wiring was missing.

Plus the review's cheap-insurance list: a rejection handler on `runWaveformFallbackOnly` (nothing could be made to reject it, but if anything ever did, `activeTask` would never release and the queue would wedge), and no more synthetic `completed` `lastRun` when the ffmpeg half did nothing.

**Verification:** 43/43 waveform suites, 902 / 0 (scanner+unit+db), 680 / 0 (integration, baseline parity), lint at parity. Both smokes re-run and extended to assert no deferral survives: normal path **17/17**, stale-binary path **14/14**.

Worth noting what three rounds of attack could *not* break: the PeakPyramid arithmetic, `HiddenLenSource`, `isReadable`'s cross-platform behaviour, `shouldChain`'s termination, the queue bookkeeping, and the webapp base64 cache (zero findings across all three rounds).

PR #821, opened 2026-08-12
feat(bundles): the launcher becomes the bundle's face (phase 1c)

IrosTheBeggar:
Phase 1c of the binary-releases roadmap — the last structural piece before **6.20.0**. The tray launcher (#817) becomes what users double-click; the server becomes `mstream-server` everywhere; #802's transitional machinery retires.

## The bundle shape

| | double-click face | server |
|---|---|---|
| Windows | `mStream.exe` (launcher — icon + VersionInfo via new `build.rs`, version read live from `package.json`) | `mstream-server.exe` |
| macOS | `mStream.app` → `CFBundleExecutable` = the launcher (menu-bar app; `LSUIElement` now describes exactly what ships, and #802's deferred Quit lives in the tray menu) | `Contents/MacOS/mstream-server` |
| Linux | `mstream-desktop` + `Terminal=false` `.desktop` | `mstream-server` |
| linux-arm64 / musl | **server-only by design** (headless targets — Pi/Alpine/NAS); plist + `.desktop` writers fall back to the server, docs say so | `mstream-server` |

One name for the server across every platform (the bundle dir carries the arch), and it's the launcher's sibling-resolution contract from #817 verbatim — the macOS agent's smoke on that PR already validated this exact layout.

## Launcher binary distribution

`bin/rust-launcher/`, CI-committed, exactly the rust-parser model: `build-rust-launcher.yml` gains build + self-test (`--autostart=status` — the one CLI path a headless runner can prove end-to-end) + commit jobs for the four desktop targets, with the same exec-bit and rebase-retry dances. Launcher PRs stay source-only. The chicken-and-egg for *this* PR: `build-bun.yml` builds the leg's launcher in-job whenever no committed binary exists, so PR bundles and their smokes always cover the real product shape.

## Retired (#802's bridge, as planned)

`mac-app-launch.js`, `instance-proof.js`, the challenge/proof headers in `server.js`, the Finder detach hook, their unit tests, and the elaborate CI nonce smokes — the launcher owns single-instance, browser-open, and failure dialogs with a real event loop. Every durable piece stays: the reboot hardening, clean listen-error exits, `LSUIElement`, the Local Network consent keys, and the read-only `dataRoot` messaging (now a winston line; the launcher does dialogs).

## New CI smoke

Launcher lifecycle on all three native legs: spawn → sibling resolution → health-event in the launcher log → hard-kill → **no orphan** (the property the whole design rests on), xvfb-backed on Linux, tray-tolerant (runners may have no StatusNotifier host — graceful degradation is itself the asserted path). `pgrep -x`, never `-f`: a cmdline pattern matches the xvfb-run wrapper — and, in one memorable local run, the polling shell's own script.

## Verified

- **Integrated Windows bundle, zero env overrides**: `mStream.exe` → found `mstream-server.exe` → served → console passthrough answers `--version` → hard-kill took the server down, port freed.
- **Integrated Linux bundle in a Docker container**: `mstream-desktop` lifecycle end-to-end **including Quick Connect `available:true` through the bundle's own iroh `.node`**.
- Full suite: **2258 pass / 0 fail** (two cancellations under deliberate parallel-build load both reproduce green in isolation, 15/15 and 4/4; the admin-access one exposed a leak-on-ready-timeout in its private boot helper — fixed here, matching the shared helper's behavior; a follow-up chip audits the other files with private helpers).
- Two local-host bundler fixes found by building bundles on a dev machine: GNU tar (first on PATH under git-bash) reads absolute `C:\` paths' drive colon as a remote hostname — broke local zips and had been **silently dropping the iroh addon from every locally cross-built bundle**. Both call sites now use cwd-relative paths.

## After merge

The first master push touching `rust-launcher/**` (this merge) commits the four launcher binaries; from then on bundles stage from `bin/`. Then: tag **v6.20.0** whenever you're ready — build-bun produces the draft release with the new bundle shape. The mstream-website download page still describes the old artifacts and `update-website.yaml` is still dispatch-only pending that rewrite.


IrosTheBeggar:
## macOS smoke checklist (for the Mac agent — post results as a comment here)

This PR makes the `.app`'s executable the tray launcher and deletes the old Finder-detach machinery, so the items below are exactly the ones only a real Mac + LaunchServices can prove. The #817 smoke ran the launcher from a shell in a hand-staged layout; **this pass must go through the actual `.app`**.

### Setup

Prefer the CI-built artifact from this PR (tests the real product bytes):

```bash
gh run list --repo IrosTheBeggar/mStream --branch claude/bundle-launcher-1c --workflow "Build Bun Server" --limit 1
gh run download <run-id> --repo IrosTheBeggar/mStream -n mStream-darwin-arm64
unzip mStream-*-darwin-arm64.zip && cd mStream-*-darwin-arm64
```

Fallback (build locally on the branch): `cargo build --release --manifest-path rust-launcher/Cargo.toml && mkdir -p bin/rust-launcher && cp rust-launcher/target/release/mstream-launcher bin/rust-launcher/mstream-launcher-darwin-arm64 && bun scripts/build-bun.mjs --target=darwin-arm64 --bundle` (the bundler stages the launcher only when that file exists — CI has its own in-job build step, local builds don't).

Sanity first: `ls mStream.app/Contents/MacOS/` should show `mStream` (launcher) + `mstream-server` + `webapp/` + `bin/`, and `defaults read "$PWD/mStream.app/Contents/Info" CFBundleExecutable` → `mStream`.

### 1. LaunchServices launch (the #802 class, final form)

- [ ] `open mStream.app` → menu-bar icon appears (compare to `build/icon.png`), **no Dock tile**, **no bounce/ANR**
- [ ] Browser opens at `http://localhost:3000` once the server is up; `/api/v1/ping` → 200
- [ ] Process split is right: `pgrep -x mStream` (launcher) and `pgrep -x mstream-server` (server) both alive; server's parent is the launcher
- [ ] Data lands in `~/Library/Application Support/mStream` (conf/db/logs) — **nothing new written inside the .app**
- [ ] Quick Connect: `curl -s localhost:3000/api/v1/iroh/code` → `shared:true, available:true` + a code (the bundled `iroh.darwin-arm64.node` loading)
- [ ] `~/Library/Application Support/mStream/logs/launcher.log` shows `starting server` → `server generation 1 spawned` → `server is up`

### 2. Re-click behavior — MEASURE, don't assume (open design question)

The old detach machinery handled the fact that macOS routes a re-open of a running app as an AppleEvent, not a new process. It's deleted; the launcher's single-instance lock assumes a second process runs and yields (opens the browser).

- [ ] With the app running, `open mStream.app` again. Record exactly what happens: new process that yields + browser opens? Nothing visible (reopen event swallowed by the tray loop)? Either result is DATA — if nothing happens, report it as such (the menu-bar icon is now the persistent affordance, so this may be acceptable; we'll decide with the observation in hand).

### 3. Tray menu click-through (the one gap no smoke has ever closed — needs TCC assistive access, or hand-click and narrate)

- [ ] **Open mStream** → browser at the server
- [ ] **Quick Connect** → browser at `/#quick-connect`, modal opens with the QR rendered
- [ ] **Start at login** toggle: on → `~/Library/LaunchAgents/mStream.plist` exists with `--autostarted` in ProgramArguments + RunAtLoad; off → file gone
- [ ] **Restart server** → new `mstream-server` pid, still serving within ~15s
- [ ] **Quit** → the graceful path nothing has driven yet: server exits via supervision (`supervisor closed stdin` in `server-console.log`), menu-bar icon disappears, **no surviving `mstream-server`**, port 3000 refused

### 4. First-run autostart policy (use an isolated `$HOME` so your real LaunchAgents stay clean)

- [ ] First tray-mode run auto-enables (plist appears) and writes the `autostartConfigured` marker in `launcher.json`
- [ ] Disable (menu or `--autostart=disable`), relaunch → **not re-enabled** (the user's choice sticks)

### 5. Terminal faces

- [ ] `./mStream.app/Contents/MacOS/mStream --version` from a terminal → the server's version, exit 0 (exec passthrough)
- [ ] `./mStream.app/Contents/MacOS/mStream --autostart=status` → prints on **stdout** (`| grep -c abled` sees it)
- [ ] `./mStream.app/Contents/MacOS/mstream-server` directly → normal CLI boot with console output

### 6. Orphan prevention on this layout

- [ ] `kill -9` the launcher (never the server) → server exits within ~5s, port freed

Notes for scripted phases: `--tray --no-open` suppresses the browser; `MSTREAM_LAUNCHER_SKIP_AUTOSTART=1` skips the first-run autostart (use it everywhere EXCEPT the deliberate §4 test). Report per-item PASS/FAIL/OBSERVED with the §2 narrative verbatim.

IrosTheBeggar:
# macOS smoke (real CI artifact) + pre-release adversarial review

**Setup**: CI artifact `mStream-6.19.3-darwin-arm64` from run 31236986673's sibling for this PR (run **31596489697**, built at HEAD `50f5f56`), on macOS 26.5.2 / Apple Silicon. Sanity: `Contents/MacOS/` = `mStream` + `mstream-server` + `webapp/` + `bin/` (rust-parser, rust-server-audio, `iroh.darwin-arm64.node`); `CFBundleExecutable` → `mStream`; exec bits intact after unzip. ✅

## §1 LaunchServices launch — PASS (every item)

- ✅ `open mStream.app` → menu-bar icon (visually matched to `build/icon.png`), **no Dock tile** (`lsappinfo` reports `type="UIElement"`), no bounce/ANR
- ✅ Browser auto-opened at `http://localhost:3000` once up (real default browser, web player rendered); `/api/v1/ping` → 200
- ✅ Process split: `mStream` (launcher) → parent of `mstream-server`; **`__CFBundleIdentifier` present in the launcher's env, stripped from the server child's** (`ps eww` on both)
- ✅ Data in `~/Library/Application Support/mStream` (conf/db/logs/caches + launcher.json/lock); **nothing new written inside the .app**
- ✅ Quick Connect: `/api/v1/iroh/code` → `shared:true, available:true` + real code (bundled `iroh.darwin-arm64.node` loads)
- ✅ launcher.log: `starting server` → `server generation 1 spawned` → `server is up`
- 👁 OBSERVED: first run pops the macOS **"App Background Activity"** notification (BTM announcing the auto-registered LaunchAgent). Every new user sees this once — expected OS behavior for the autostart-on-by-default decision, noting it for the record.

## §2 Re-click behavior — OBSERVED (measured, verbatim)

> With the app running, `open mStream.app` again: exit 0, **no second launcher process ever appeared** (pid set identical before/after), **zero new launcher.log lines** — the single-instance lock path never executed. LaunchServices routed the re-click as a reopen AppleEvent to the running instance, and the tao event loop **swallows it silently**: no browser, no visible reaction of any kind. The menu-bar icon is the only affordance after a re-click. The lock-yield + browser-open path still covers bare-binary launches and a *second copy* of the .app at a different path — it is specifically same-.app re-clicks that go dark.

Related: the review found **tao 0.30 exposes the reopen event** (`Event::Reopen`, mapped from `applicationShouldHandleReopen`) — a single match arm in `tray_app.rs`'s event loop opening the server URL would restore the re-click affordance the old machinery had. Recommended.

## §3 Tray menu click-through — BLOCKED (TCC), equivalents proven

`osascript`/System Events assistive access is denied on this runner (error -1719) and granting it is a system security-settings change I won't make. What was proven by equivalent paths:

- ✅ **Quick Connect target**: browser at `/#quick-connect` on this exact bundle's server → modal `hystmodal--active`, QR rendered (DOM-asserted)
- ✅ **Start at login** underlying `set_enabled`: proven via `--autostart` CLI both real-HOME and isolated-HOME
- ❌ Not driven: **menu-click Open / Restart server / graceful Quit** (the Quit supervision-close path specifically). Still needs one human click-through on a Mac.

## §4 First-run autostart policy (isolated $HOME) — PASS

✅ First tray run wrote the LaunchAgent plist + `"autostartConfigured": true` marker; ✅ after `--autostart=disable`, relaunch did **not** re-enable.

## §5 Terminal faces — PASS

- ✅ `./mStream --version` (pty) → `6.19.3`, exit 0 (exec passthrough)
- ✅ `./mStream --autostart=status 2>/dev/null | grep -c abled` → `1` — status prints on **stdout**
- ✅ `./mstream-server` direct → normal CLI boot (banner, DB init, "Access mStream locally"), clean Ctrl+C exit 0

## §6 Orphan prevention — PASS

✅ `kill -9` launcher → `[supervision] supervisor closed stdin — shutting down` in server-console.log, server gone and port 3000 refused in <5 s.

## Extra (release-relevant): the .app has no bundle codesign seal

`codesign --verify --deep --strict mStream.app` → **"code has no resources but signature indicates they must be present"**; `spctl --assess` rejects it. Both binaries are only linker-ad-hoc signed and there's no `_CodeSignature/CodeResources`. My `gh`-downloaded copy carried no quarantine so LaunchServices ran it — but a **browser-downloaded release zip is quarantined**, and a malformed-seal app tends to get the *"mStream is damaged and can't be opened"* dialog with no right-click bypass, worse than the normal unidentified-developer flow. Cheap mitigation in the bundler after staging: `codesign --force --deep -s - mStream.app` (ad-hoc seal). Real fix eventually: Developer ID + notarization.

Machine state after the run: LaunchAgent removed via the product's own `--autostart=disable`, data home deleted, no stray processes — clean.

---

# Adversarial review (70 agents: 7 lenses → dedup → 2–3 adversarial verifiers per finding → completeness critic)

42 raw findings → 30 unique → **30 confirmed, 0 refuted** (every finding survived independent code-traced refutation attempts), + 6 completeness-critic gaps. Triaged for the release:

## 🔴 Fix before merge (these bite the v6.20.0 release directly)

1. **Windows CI self-test asserts on stdout the GUI exe bypasses — the binary-commit pipeline never produces its first commit** (`build-rust-launcher.yml:131` + `platform.rs:89`). Under Actions, bash.exe has a hidden console, so `AttachConsole` succeeds and `--autostart=status` goes to CONOUT$ instead of the command-substitution pipe → `$out` empty → win leg red → `commit-binaries` (needs: build) skipped **for all four targets**. Master goes red at the merge, `bin/rust-launcher/` never populates. Fix in `console_out`: prefer `GetStdHandle(STD_OUTPUT_HANDLE)` when it's a valid pipe/redirect and only fall back to CONOUT$ — which also fixes real users piping `mStream.exe --autostart=status`. (Ironic counterpart: the macOS pass above proves the *unix* stdout contract works.)
2. **linux-x64 launcher inherits ubuntu-latest's glibc ≥ 2.34 floor** (`build-rust-launcher.yml:88` and the in-job fallback in build-bun.yml). Dead `mstream-desktop` on RHEL/Rocky 8, Ubuntu 20.04, Debian 11, Amazon Linux 2 — the exact distros build-bun.mjs's musl-parser fallback exists to keep working, and with `Terminal=false` the failure is perfectly silent. Verifier note: ubuntu-22.04 does NOT fix this (still 2.35); needs a ≤2.31 sysroot/container or cargo-zigbuild, or a documented floor + loud failure.
3. **The health probe trusts any process on the port — EADDRINUSE opens the browser at a stranger and hides the boot failure** (`tray_app.rs:226/286`, `server.rs:71`). Port already occupied (React/Grafana/anything on 3000) → first TCP probe succeeds instantly → `ServerUp` accepted (generation still matches — it's only bumped by `stop_current`, not by a child death) → `ever_up=true`, browser opens on the foreign app, and when the real child exits with EADDRINUSE ~1 s later the `!ever_up` gate suppresses the boot-failure dialog. This is a regression of the exact impostor scenario the deleted `instance-proof.js` existed to prevent. Minimal fix: don't announce `ServerUp` for a generation whose child already exited, and treat an instant first-probe success (before the child could bind) as suspect; better, probe an identity endpoint (`/api/v1/ping`) instead of a bare TCP connect.
4. **The GNU-tar "fix" silently writes an uncompressed tar named `.zip` for locally built Windows bundles** (`build-bun.mjs:242`). GNU tar can't create zips; unrecognized `-a` suffix → plain tar, exit 0. CI is unaffected (pwsh resolves System32 bsdtar) — this is the local git-bash path the fix was for. Prefer `%SystemRoot%\System32\tar.exe` explicitly on Windows, or verify the output starts with `PK\x03\x04`.
5. **The v6.20.0 Windows launcher will report FileVersion 6.19.3** (`build-rust-launcher.yml:25` + `build.rs`). The commit workflow's paths filter is `rust-launcher/**` only, so the release version bump never rebuilds; the committed exe keeps the version from the last crate change. Add `package.json` to the paths filter (commit step already no-ops on identical bytes) or dispatch the workflow before tagging.

## 🟠 Release checklist (decisions/process, not necessarily this PR's code)

- **Fresh no-account desktop install serves the Iroh pairing code (embedding the connect secret) to any LAN peer** (`boot-config.js:95` + `api/iroh.js`; #816 provenance, but v6.20.0 makes it the product's out-of-box posture): public mode grants admin to every request, server binds `::`, so any Wi-Fi peer can fetch the code and hold a persistent tunnel into the library. Decide before release: gate the code behind a created account / admin network, or accept + document.
- **Autostart entries pin the absolute install path and are never re-asserted** (`autostart.rs:72`): first run in `~/Downloads/mStream-6.20.0-…/`, then move to `/Applications` (or upgrade to the next versioned folder) → login item silently dead while the tray checkbox still shows enabled. One-liner: when `is_enabled()`, re-run `set_enabled(true)` on each tray boot to refresh the path.
- **Tag-in-the-rebuild-window ships a stale launcher** (`build-bun.yml:98`): the in-job build only fires when the committed binary is *missing*, never stale — a tag pushed while build-rust-launcher is still compiling stages pre-merge binaries, green smokes and all. Process rule: wait for build-rust-launcher green before tagging (or add a provenance check).
- **mstream.io still advertises the removed Electron installers** (`update-website.yaml` disabled) — the release's headline feature won't be on the download page until the website rewrite lands; needs to be on the v6.20.0 checklist.
- **Default-on autostart has never been exercised end-to-end on Windows/Linux** (CI smokes set `MSTREAM_LAUNCHER_SKIP_AUTOSTART=1`; only macOS has a real-login verification — above).
- **.app codesign seal** (empirical, above) — one bundler line.

## 🟡 Confirmed, post-release backlog (sev-ordered highlights of the remaining 21)

| Finding | Where |
|---|---|
| Autostart entries unquoted → breaks for install paths with spaces | `autostart.rs:33` |
| `Event::Reopen` one-liner restores macOS re-click | `tray_app.rs:262` |
| Tray URLs hardcode `http://` + boot-time port snapshot; admin port/SSL changes strand Open/Quick Connect/probe | `paths.rs:112`, `tray_app.rs:134` |
| Single-instance lock is global-per-user; contradicts the `--portable` carve-out comment | `tray_app.rs:50` |
| `read_port` rejects string `"port"` Joi coerces → launcher probes 3000 while server listens elsewhere | `paths.rs:79` |
| Empty-string `XDG_DATA_HOME`/`LOCALAPPDATA`: Rust treats as set, JS as unset → split data homes | `paths.rs:26` |
| `--server-bin` override keeps the *launcher's* exe_dir for the legacy-save/portable anchor → ladder divergence | `paths.rs:65` |
| Launcher source changes get no integrated bundle test until after commit to master | `build-bun.yml:13` |
| Two rapid launcher merges race the commit job into a rebase conflict that can keep stale binaries | `build-rust-launcher.yml:179` |
| darwin-x64 leg has no assertion the launcher actually staged | `build-bun.mjs:165` |
| `--supervised` as a `-j` value is eaten by the launcher's flag parser → supervision silently lost | `main.rs:76` |
| Probe/Open assume loopback; `address` pinned elsewhere → "never up" | `server.rs:72` |
| Headless-Linux tray panics via raw gtk `expect` instead of clean error | `tray_app.rs:99` |
| `ensure_default_on` panics on valid-but-non-object `launcher.json` | `autostart.rs:77` |
| `build.rs` keys Windows resources off HOST os, not compile target | `build.rs:8` |
| README "Binaries" section still describes the v6.13 shape | `README.md:50` |
| Stale comments: launcher cites deleted `mac-app-launch`; server.js cites retired relaunch-exit-0 | `server.rs:34`, `server.js:506` |
| Win terminal passthrough rebinds child to CONIN$/CONOUT$ → piping/redirection of server output is clobbered | `platform.rs:56` |
| Committed launcher binaries (~20 MB) will ship inside the npm tarball | `.gitignore:26` |
| launcher.log / server-console.log grow unbounded for an always-on login item | `server.rs:21` |
| `.instance-nonce` untracked-noise after the .gitignore removal | `.gitignore:42` |

Full finding descriptions, failure scenarios, and both verifier rationales for every item are archived; happy to expand any of them into issues on request.

IrosTheBeggar:
# Review follow-ups landed: 5 merge blockers + 8 cheap backlog items

Two batches pushed since the review comment above.

## `f1750df3` — the five merge blockers

1. **Windows `console_out`/`console_err` honor a real std handle** (pipe/`$()`/redirect) before falling back to `CONOUT$` — the binary-commit self-test was capturing an empty string while the status line went to the runner's hidden console. **Proven on a real windows-latest runner** via `workflow_dispatch dry_run=true` ([run 31604951556](https://github.com/IrosTheBeggar/mStream/actions/runs/31604951556)): `self-test --autostart=status -> 'disabled'`. Also fixes real users piping `mStream.exe --autostart=status`.
2. **linux-x64 launcher builds in a `debian:11` container** — the ubuntu-latest link left `mstream-desktop` with a GLIBC_2.34 floor, silently dead (`Terminal=false`) on Debian 11 / Ubuntu 20.04-era distros. Dry-run measured **`max glibc symbol version: 2.30`**, and a new `objdump` assertion fails the job loudly if the floor regresses. install.md documents the floor + the `mstream-server` fallback; build-bun's in-job fallback carries a "don't tag while bin/rust-launcher is empty" caveat.
3. **The health probe is an identity probe now** (`GET /` must answer 200 and identify as mStream) **and `ServerUp` is ignored when the child already exited** — a port squatter no longer gets the browser pointed at it while the child's EADDRINUSE death goes un-dialoged (the regression vs the retired instance-proof machinery). Verified live on macOS both ways: a foreign `http.server` on 3000 → no false "server is up", `server exited unexpectedly` + the boot-failure alert; real server → announces normally. CI lifecycle smokes on linux/win exercise the new probe too (all green).
4. **Local Windows bundles prefer `System32\tar.exe`** — GNU tar (first on PATH under git-bash) can't create zips and was silently writing an uncompressed tar named `.zip` — **and every host verifies the `PK` magic** after archiving, so a wrong archive fails the build instead of shipping.
5. **`package.json` joined the binary-commit workflow's push paths** so the release version bump re-stamps the Windows VersionInfo (v6.20.0's `mStream.exe` would otherwise report 6.19.3), with the "wait for the commit workflow before tagging" release rule documented inline.

## `10f0b7ca` + `f397d967` — cheap high-value backlog items

- **macOS re-click restored**: tao's `Event::Reopen` is now handled — re-clicking the running `.app` opens the browser at the server. Measured end-to-end through LaunchServices: the reopen AppleEvent that §2 found silently swallowed now logs `reopen event - opening browser` and opens the tab (`--no-open` still suppresses for smokes).
- **Two panic guards**: headless Linux (no `DISPLAY`/`WAYLAND_DISPLAY`) fails cleanly with a logged, explained error instead of gtk's raw panic; a `launcher.json` holding valid non-object JSON (`[]`) no longer panics `ensure_default_on` on the GUI face — verified a `[]` state file boots and is rebuilt into a proper object.
- **Autostart path staleness fixed**: entries re-assert the current exe path on every configured-and-enabled boot, so versioned-folder upgrades and Downloads→/Applications moves no longer strand the login item on a dead path. Verified: move the binary, relaunch, the LaunchAgent follows; disable still sticks.
- **Autostart quoting**: the command path is quoted on Windows/Linux (auto-launch 0.5 writes it verbatim, so spaces word-split and launch nothing); macOS keeps the raw path — verified a space-containing install path lands intact in the plist.
- **Housekeeping**: README's Binaries section now describes the actual v6.20 bundle shape (was the v6.13 "batteries included coming soon" text); the stale `server.js` listen-error comment and the launcher's `mac-app-launch` reference updated to post-retirement reality; `.instance-nonce` is ignored again as legacy residue (pre-6.20 checkouts still carry one).

**Verification on this batch**: `cargo build` + `clippy -D warnings` clean on macOS and the Windows target; empirical macOS tests for reopen, both panic-guard paths, and path re-assert as described; all machine state cleaned after.

Still open from the review (deliberate decisions, not one-liners): the fresh-install **Iroh pairing-code LAN exposure**, `.app` **codesign seal**, SSL/port-change URL assumptions, single-instance lock scope, the config-ladder divergences, launcher-PR integrated-bundle testing, npm tarball weight, log rotation, and the website rewrite. Ranked list with scenarios is in the review comment above.

IrosTheBeggar:
# Independent audit of the review follow-up commits (Windows + Linux verification)

Audited all nine follow-up commits on the platforms the macOS pass couldn't exercise. **Verdict: merge-worthy.** Full suite on the audited head: **2265 pass / 0 fail / 0 cancelled** (clean machine).

## Independently verified

- **Windows stdio semantics** (`f1750df3`/`f5f46b44`): `--autostart=status | grep` and `$()` capture work on real Windows; the terminal face carries the server's usage-error text through a pipe and mirrors exit 1. The per-handle inherit-vs-rebind logic in the pass-through behaves as documented.
- **Identity probe** (blocker 3): positive path announces on the real bundle (Windows + Linux). Negative path proven twice — a foreign `http.server` on the expected port drew **zero** false announces with the child's EADDRINUSE death logged; and an accidental *real-mStream* squatter in the Linux container produced the exact designed outcome (identity passes, dead child suppresses the announce).
- **Supervision post-changes**: controlled single-pid launcher kill → server exited in ~1s, port refused (Linux container; Windows bundle equivalent also green).
- **Headless-Linux guard**: clean explained error, exit 1, no child spawned.
- **Rust**: build + `clippy -D warnings` clean on Windows; all 9 new unit tests pass on Windows and Linux. The paths-parity work (Joi port coercion, endpoint probing, server-dir anchoring, `-j` value protection, empty-env rules) is correct and pinned by tests.
- **`.npmignore`**: pattern-by-pattern diff against `.gitignore` — exact mirror with only the intended launcher-binary divergence. One copy-paste negation fixed in `5cefe011` (`save/sync-cdn`'s block negated `save/conf/README.md`).
- **Server-side gate** (`2fc20815`): `getAllUsers()` confirmed as the real account store (auth.js keys on the same call), fail-closed address handling, demo/source posture untouched, account creation restores LAN semantics.

## Flag for the maintainer (decision, not defect)

The pre-release review listed the fresh-install pairing-code LAN exposure as *"decide before release"* — `2fc20815` implements the loopback-only gate rather than waiting for that decision. The implementation is sound and the desktop pairing flow survives (the launcher opens the owner's local browser; the QR lives there). **Needs explicit ratification**: acceptable trade is "LAN devices on an accountless install don't see the Quick Connect code"; reverting is one clause + one test file if not.

## Notes (filed, no action)

- Version-only `package.json` bump + instant tag passes tree-hash provenance while Windows VersionInfo is stale — the documented "wait for build-rust-launcher's commit before tagging" rule is process, not enforcement.
- Micro-edges: a `ui:'subsonic'` desktop config may serve an index the "mStream" probe won't match; the "starting server" launcher.log line prints before event-loop build (cosmetic ordering on set-but-dead DISPLAY).
- Audit-process disclosure: three intermediate scares (phantom pre-check server, dead Xvfb, "supervision orphan") were all this audit's own container-harness bugs (zombie-confused pgrep, self-matching patterns in an init-less container); the pushed code was correct in each case.

IrosTheBeggar:
# Windows interactive click-through — PASS (real desktop, real clicks)

Companion to the macOS smoke above: the full tray-menu click-through on a real Windows 10 desktop, driven interactively against the audited head's bundle (`mStream.exe` + `mstream-server.exe` from `dist/stage`). This closes the "literal menu clicks" gap on Windows.

| Step | Result |
|---|---|
| Shell launch (ShellExecute, no console — the double-click environment) | ✅ GUI face: launcher + server up, port 3000, desktop-profile config generated in real `%LOCALAPPDATA%\mStream` with Quick Connect on |
| Browser-open on first run | ✅ New Firefox process spawned the same second as the `server is up` announce |
| First-run autostart | ✅ Real HKCU Run key written: `"…\mStream.exe" --autostarted` (quoted path) |
| Tray icon | ✅ Visible in the tray overflow with tooltip **"mStream Server - http://localhost:3000"** (post-announce tooltip); menu opens on left-click |
| Menu render | ✅ Open mStream · Quick Connect · ✓ Start at login · Restart server · Quit mStream (screenshot-zoomed; checkbox reflects state, later verified unchecked after toggle) |
| **Open mStream** click | ✅ Firefox foregrounded, new content process at the click second |
| **Quick Connect** click | ✅ Browser foregrounded at the deep link (modal behavior previously DOM-verified) |
| **Start at login** click (off) | ✅ Run key deleted + `autostart disabled` logged; menu re-render shows unchecked |
| **Restart server** click | ✅ `menu: restart server` → new generation spawned → new server pid → identity probe re-announced → serving |
| **Quit mStream** click | ✅ **The graceful path, driven by click for the first time anywhere**: `menu: quit` → server logs `[supervision] supervisor closed stdin — shutting down` → both processes exit, port refused, tray icon removed |

Environment notes: three ghost tray icons from earlier force-kill smokes appeared alongside the live one (standard Windows behavior — dead tray icons persist until hovered; force-kills don't run the tray cleanup, the graceful Quit does). Machine left clean: data home removed, registry clean, no processes.

With this, every checklist item is covered on Windows; the remaining platform gap is the same short menu click-through on macOS (TCC-blocked for the agent there) and a real login-time autostart fire on any platform.

PR #827, opened 2026-08-17
feat(launcher): tray 'View logs' — terminal with the full session's server log

IrosTheBeggar:
Adds a **View logs** item to the tray menu (grouped with Restart server). Clicking it opens a terminal window showing a short tail of `launcher.log`, then `server-console.log` **from line 1**, followed live (`-F`, so the per-session rotation never ends the view).

## Why whole-file instead of fetching the admin live-log buffer

The admin panel's live-log viewer is backed by an in-memory ring (`src/logger.js` — a `MemoryRingTransport` attached to the same winston root logger as the Console transport). Under the tray, the launcher captures that exact console stream into `server-console.log` starting at server spawn — so from line 1 the terminal shows everything the admin viewer has, **uncapped** (the ring keeps the last N entries at 4 KB each), plus any early-boot lines the ring already evicted, with winston's colors rendering natively in the terminal. Fetching `/api/v1/admin/logs/recent` instead would add an auth dependency (the wall, once accounts exist) for a strict subset of the file. The doc comment on `open_logs_terminal` records this equivalence so future logger-transport changes know the terminal depends on it.

## Per-OS "what is a terminal"

- **macOS** — writes an executable `view-logs.command` into the logs dir and `open`s it: Terminal runs it as a document, avoiding the AppleEvents automation-consent prompt that scripting Terminal directly would trigger.
- **Windows** — `powershell -NoExit … Get-Content -Wait` in a `CREATE_NEW_CONSOLE` window (the GUI-subsystem launcher has no console of its own to lend).
- **Linux** — first of `x-terminal-emulator` / `gnome-terminal` / `konsole` / `xfce4-terminal` / `xterm` that spawns (each has its own exec-argument dialect); with no emulator at all it degrades to `xdg-open` on the logs directory.

Failures are logged and can never take the tray down.

## Verification

- macOS: real end-to-end — the ignored manual test (`cargo test -- --ignored manual_open_logs_terminal`) opens an actual Terminal window; verified both log sections render, plus a direct BSD `tail -n +1 -F` whole-file-then-follow check, plus a live click-through on a running 6.20.0 bundle.
- `cargo build` + `clippy -D warnings` + tests green on darwin and the `x86_64-pc-windows-msvc` target.
- Windows/Linux terminal spawns are compile-verified; the command shapes are standard (`Get-Content -Wait`, per-emulator exec args) — worth one manual click on each platform before the next release.

### DIFF
--- a/rust-launcher/src/platform.rs
+++ b/rust-launcher/src/platform.rs
@@ -213,7 +213,10 @@ pub fn open_logs_terminal(logs_dir: &std::path::Path) -> Result<(), String> {
     {
         use std::os::windows::process::CommandExt;
         // The launcher is a GUI-subsystem exe with no console of its own to
-        // lend — give the tail a brand-new console window.
+        // lend — give the tail a brand-new console window. -Encoding UTF8 on
+        // both reads: the server writes UTF-8 (winston's em-dashes), and
+        // Windows PowerShell 5.1's Get-Content defaults to the ANSI codepage
+        // for BOM-less files — without it every "—" renders as "â€"".
         const CREATE_NEW_CONSOLE: u32 = 0x0000_0010;
         let dir = logs_dir.display().to_string().replace('\'', "''");
         std::process::Command::new("powershell.exe")
@@ -223,9 +226,9 @@ pub fn open_logs_terminal(logs_dir: &std::path::Path) -> Result<(), String> {
                 "-Command",
                 &format!(
                     "Set-Location '{dir}'; Write-Host '== launcher.log =='; \
-                     Get-Content .\\launcher.log -Tail 50 -ErrorAction SilentlyContinue; \
+                     Get-Content .\\launcher.log -Tail 50 -Encoding UTF8 -ErrorAction SilentlyContinue; \
                      Write-Host ''; Write-Host '== server-console.log: full server log for this session (following; close the window to stop) =='; \
-                     Get-Content .\\server-console.log -Wait"
+                     Get-Content .\\server-console.log -Wait -Encoding UTF8"
                 ),
             ])
             .creation_flags(CREATE_NEW_CONSOLE)
--- a/rust-launcher/src/platform.rs
+++ b/rust-launcher/src/platform.rs
@@ -173,3 +173,116 @@ pub fn fatal_alert(msg: &str) {
         eprintln!("mStream: {msg}");
     }
 }
+
+/// Tray "View logs": open a terminal window showing the two launcher-owned
+/// logs — a static tail of the quiet launcher.log, then server-console.log
+/// FROM LINE 1, followed live (-F, so the per-session rotation doesn't end
+/// the view). Whole-file is deliberate: the same winston stream that feeds
+/// the admin panel's in-memory live-log ring (src/logger.js — Console +
+/// MemoryRingTransport on one root logger) is what the launcher captures
+/// into server-console.log, and the capture starts at server spawn — so
+/// from line 1 this window shows everything the admin viewer has, uncapped
+/// (the ring holds the last N entries, 4KB each), plus anything the ring
+/// already evicted. Fetching /api/v1/admin/logs/recent instead would add
+/// an auth dependency (the wall, once accounts exist) for a subset of this
+/// file. Per-OS "what is a terminal" seams; the caller logs a failure —
+/// a missing terminal emulator must never take the tray down.
+pub fn open_logs_terminal(logs_dir: &std::path::Path) -> Result<(), String> {
+    #[cfg(target_os = "macos")]
+    {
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
## Windows audit + interactive smoke — PASS (one cosmetic nit)

**Static audit** (read the full diff, verified every load-bearing claim against the code):
- `logs_dir` is a real in-scope binding (tray_app.rs:39) and `server-console.log` is genuinely the launcher's stdout+stderr capture of the server child (server.rs:21, `create(true).append(true)`), so the whole-file-instead-of-admin-ring rationale holds. `server-console.log` is per-launcher-session (rotated to `.1` at launcher start, appended across in-session Restarts) — the "full session from line 1" claim is exact.
- The one edge I probed — first click before the file exists (`Get-Content -Wait` on a missing file errors and exits) — is **unreachable**: the file is created at spawn, and the menu only exists after spawn.
- Apostrophe-in-path escaping (`'` → `''`) verified against a real `It's a dir\logs` path: renders correctly.
- `cargo build` + `clippy -D warnings` + 11/11 tests green on windows-msvc.

**Interactive smoke** on this machine's live install (swapped the PR launcher into the daily-driver bundle, launched via Explorer ShellExecute — the real double-click path):
- Menu shows the new 6-item layout, View logs grouped with Restart server ✓
- Click → new PowerShell console opens with launcher.log tail (ending in the very `menu: view logs` line the click produced), the `== server-console.log ==` header, then the server log from line 1 (banner, migrations, tunnel-up, mDNS), winston colors rendering natively ✓
- **Live follow** ✓ — a bogus login attempt I fired at the API appeared in the window unprompted (`Failed login attempt ... Username: viewlogs-smoke-marker`)
- **Survives Restart server** ✓ — the window kept following straight into generation 3's full boot sequence, appended in place (the crash-loop-evidence policy working end-to-end)
- Closing the window stops it; launcher/server unaffected; `launcher.log` recorded `menu: view logs` → `menu: restart server` → `server generation 3 spawned` ✓

**One cosmetic nit (recommend fixing before merge, one token):** winston's em-dashes render as `â€"` in the window (`tunnel up â€" endpointId=...`). Windows PowerShell 5.1's `Get-Content` defaults to the ANSI codepage for BOM-less files, and the server writes UTF-8. Add `-Encoding UTF8` to both `Get-Content` calls in the Windows arm of `open_logs_terminal`. (Console font/codepage renders the decoded chars fine — it's purely the read encoding.)

Not a PR issue but worth knowing for anyone smoking this: launching the launcher via a tool's `Start-Process` from a process that owns a hidden console makes it take the **terminal face** (AttachConsole succeeds → passthrough, no tray, no launcher.log). Cost me ten confused minutes; Explorer/ShellExecute is the faithful GUI-launch path.

Verdict: **merge-worthy** with the `-Encoding UTF8` fix.

IrosTheBeggar:
## Linux smoke (Docker, Xvfb + real xterm) — PASS, all four arms of the Linux code path exercised

Drove `open_logs_terminal()` directly (temporary local env hook, not committed) in a debian:12 container with GTK3 + Xvfb + xterm as the ONLY emulator, logs dir at `/tmp/It's a dir/logs` to exercise `sh_quote`. Read what the user would SEE via xterm's own screen log (`-l -lf`).

**1. Real terminal path** ✓ — spawned with the correctly quoted cwd (`cd '/tmp/It'\''s a dir/logs'`), and the screen log shows the exact designed layout:
```
== launcher.log ==
[demo] launcher line 1
[demo] launcher line 2

== server-console.log: full server log for this session (following; Ctrl+C to stop) ==
server A
server B
server C (appended live)      <- LIVE FOLLOW: appended after the window opened
```
Debian-specific note worth knowing: on Debian/Ubuntu `x-terminal-emulator` resolves (via alternatives → `lxterm` locale wrapper) to xterm, so the chain's FIRST candidate wins on any Debian-family desktop — the ordering is right.

**2. Rotation survival (`-F`)** ✓ — renamed `server-console.log` → `.1` and recreated it while the window was following, exactly what a launcher restart does. coreutils tail printed `'server-console.log' has been replaced; following new file` and picked up the new file's first line. The per-session rotation genuinely does not end the view.

**3. No-emulator degrade** ✓ — with every emulator absent, `xdg-open` was invoked with the logs dir (`xdg-open called with: /tmp/It's a dir/logs`), returned Ok.

**4. Nothing at all** ✓ — no emulator AND no xdg-open: returned `Err(no terminal emulator found (tried x-terminal-emulator, gnome-terminal, konsole, xfce4-terminal, xterm) and xdg-open failed)` cleanly — the tray-app arm logs it, no panic.

Combined with the Windows smoke above and the mac agent's end-to-end, all three platform arms are now interactively verified. Still merge-worthy; the only outstanding item is the Windows `-Encoding UTF8` nit.

Issue #18, opened 2017-03-15
Make static vPaths and secrets easier to use


IrosTheBeggar:
That way sessions last across reboots

IrosTheBeggar:
We should force that the vPath and secret are both random or static. 

If the secret is static but vpath is not, a server reboot can cause the frontends to go out of sync

IrosTheBeggar:
Fixed this in the express version.  Closing

Issue #17, opened 2017-03-03
Playlists don't auto-remove on delete


IrosTheBeggar:
Playlist is successfully deleted but not removed from interface

IrosTheBeggar:
Also playlists should add to browser panel if the playlist view is open

Issue #12, opened 2016-12-29
New playlist button doesn't work


MaxLeiter:
The new playlist button doesn't work

the one in the top right:
![image](https://i.imgur.com/LqvnZ2R.png)

IrosTheBeggar:
That button adds songs all to the playlist.  Works only in the file-browser currently

PR #829, opened 2026-08-18
feat(release): manifest.json + one-line install scripts (P3 trust, part 2)

IrosTheBeggar:
## What ships

**Every release gets a `manifest.json`** — `{name, version, apiVersion, assets:[{file, sha256}]}`, the same shape as mstream-terminal-player's — generated in the release job from the downloaded bundles and uploaded beside them. The job also refuses to publish if any zip's filename doesn't carry the tag's version (the installers derive filenames from `manifest.version`, so a lagging package.json would produce an un-installable release).

**`install.sh`** (Linux/macOS, POSIX sh) and **`install.ps1`** (Windows, PS 5.1 floor, pure ASCII):

```sh
curl -fsSL https://raw.githubusercontent.com/IrosTheBeggar/mStream/master/install.sh | sh
```
```powershell
irm https://raw.githubusercontent.com/IrosTheBeggar/mStream/master/install.ps1 | iex
```

Both: fetch the manifest **first** → derive the exact asset for this machine (OS + arch + **glibc-vs-musl on Linux**, via ldd's banner / `ld-musl` presence) → download → verify sha256 → extract into a versioned folder under the data-home-adjacent app root → point `current` at it (symlink / junction) → wire a `mstream-server` command (`~/.local/bin` symlink / user PATH) → desktop integration (Linux: the bundle's own `.desktop` with `%INSTALL_DIR%` filled — that placeholder has been waiting for exactly this; mac: `~/Applications/mStream.app` link; Windows: Start Menu shortcut). **Re-run = upgrade in place**; data lives outside the app folder, never touched. Old versions kept beside `current` for rollback. `MSTREAM_RELEASE_BASE` points at an internal mirror (also how this was tested).

**Post-install proof: `mstream-server -V`** (commander — instant, no boot). This caught a real gap: **Bun's musl build dynamically links `libstdc++`/`libgcc`, which a minimal Alpine lacks** — a wall of loader `Error relocating`. The installer now reports it *at install time with the fix* (`apk add libstdc++ libgcc`) instead of at the user's first run. Worth a line in the musl bundle's README too (follow-up).

## Layout
| | app root | command |
|---|---|---|
| Linux | `~/.local/share/mstream/app/<bundle>/` + `current` | `~/.local/bin/mstream-server` |
| macOS | `~/Library/Application Support/mStream/app/` | same |
| Windows | `%LOCALAPPDATA%\Programs\mStream\<bundle>\` + `current` junction | user PATH |

This is also the layout P4's auto-updater will want: versioned folders + one pointer flip.

## Verified against the REAL v6.20.0 assets (local mirror serving them + a generated manifest)
- **Windows**: install · upgrade-in-place (stale file gone, junction valid) · tampered-hash refusal · installed server boots from the install dir (`-j` config on :3998, live :3000 untouched)
- **Debian 12**: 8/8 — picked `linux-x64`, layout, symlinks, boot, upgrade, tamper refusal, desktop entry substituted with the `current` path
- **Alpine 3.20**: picked `linux-x64-musl` (the decisive libc test); bare box → runtime hint printed; after `apk add` → `mstream-server 6.20.0 runs`
- macOS: same code path as Linux minus the `.desktop` arm; the `~/Applications` link + `.app`-internal server path need one real run (mac agent).

Also: `docs/install.md` leads with the one-liners; the bundle README points at the script for upgrades.

Note: the manifest step lands in the release job **downstream** of #826's signing/stapling in the build legs — no overlap. Once #826 merges, the stapled mac zips are what get hashed.

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -801,6 +801,47 @@ jobs:
           path: bundles
           merge-multiple: true
 
+      # manifest.json: the release's own inventory — version + sha256 per zip.
+      # install.sh / install.ps1 read it FIRST (to learn the exact asset name
+      # for their platform, then to verify the download), and the future
+      # launcher auto-updater will read the same file. Same shape as
+      # mstream-terminal-player's, so tooling that understands one
+      # understands both: {name, version, apiVersion, assets:[{file,sha256}]}.
+      # Bump apiVersion only on a breaking change to this shape.
+      - name: Generate sha256 manifest
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd bundles
+          version="${GITHUB_REF_NAME#v}"
+          {
+            echo '{'
+            echo '  "name": "mstream",'
+            echo "  \"version\": \"${version}\","
+            echo '  "apiVersion": 1,'
+            echo '  "assets": ['
+            first=true
+            for f in mStream-*.zip; do
+              sum=$(sha256sum "$f" | cut -d' ' -f1)
+              $first || echo ','
+              first=false
+              printf '    {"file": "%s", "sha256": "%s"}' "$f" "$sum"
+            done
+            echo ''
+            echo '  ]'
+            echo '}'
+          } > manifest.json
+          cat manifest.json
+          # Every asset name must carry THIS version: the installers derive
+          # the filename from manifest.version, so a bundle built from a
+          # package.json that lags the tag would be un-installable.
+          for f in mStream-*.zip; do
+            case "$f" in
+              "mStream-${version}-"*) ;;
+              *) echo "::error::$f does not match tag version ${version} - package.json and the tag disagree"; exit 1 ;;
+            esac
+          done
+
       - name: Attach bundles to the release
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -814,4 +855,4 @@ jobs:
           if ! gh release view "$TAG" --repo "$REPO" >/dev/null 2>&1; then
             gh release create "$TAG" --repo "$REPO" --title "$TAG" --generate-notes --draft
           fi
-          gh release upload "$TAG" bundles/*.zip --clobber --repo "$REPO"
+          gh release upload "$TAG" bundles/*.zip bundles/manifest.json --clobber --repo "$REPO"
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -27,6 +27,11 @@ on:
       # to codesign. An entitlements-only PR without this filter would merge
       # with zero CI and first run on the release tag.
       - 'build/**'
+      # The install scripts + manifest generator have their own PR-time
+      # contract check (installer-contract job below).
+      - 'install.sh'
+      - 'install.ps1'
+      - 'scripts/release-manifest.sh'
   workflow_dispatch:
     inputs:
       notarize:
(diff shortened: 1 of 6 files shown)


IrosTheBeggar:
## Adversarial review applied — findings 1–7, 10, 16 fixed (commit d68c3074)

An ultracode panel (8 lenses × 3 skeptics, 108 agents) reviewed this PR + the merged signing series. Full report is in the session; the fixes for the "merge with fixes" set are in:

| # | Finding | Fix |
|---|---|---|
| **1 HIGH** | Every published release lacked `manifest.json` → the one-liners 404'd on merge day | **Backfilled `manifest.json` onto v6.20.1** (7 real assets hashed with the workflow's exact logic; live-verified: `releases/latest/download/manifest.json` resolves, and `install.ps1` against real GitHub now installs 6.20.1 end-to-end). Both scripts explain a missing manifest (older release / newer `MSTREAM_VERSION` / manual download) instead of a bare 404; docs no longer cite v6.20.0. |
| **2 HIGH** | Upgrades never reached the login item — the launcher registers its *versioned* path, so every reboot silently booted the old version; deleting it per the hint = silent no-boot | After the `current` flip, the **new** launcher re-registers via `--autostart=enable` (its no-tray CLI) when the user has it enabled; a running instance from elsewhere is detected and the user told it stays on its version until Quit; delete hint reworded |
| **3 MED** | Same-version re-run `rm -rf`'d the live install (fatal for `--portable` state inside the bundle); ps1 tore through `bin\`/`webapp\` before hitting the locked exe | Already installed → keep, re-verify hash, re-point links only. `MSTREAM_FORCE=1` **moves aside**, never rm. ps1 refuses a forced replace while a process runs from that folder |
| **4 MED** | musl false-positive on glibc boxes with `/lib/ld-musl` (Debian `musl-tools`, Arch) | Detect glibc **positively** first; `MSTREAM_KEY` override |
| **5 MED** | 32-bit Pi OS on 64-bit kernel → unrunnable arm64 bundle, exit 0 | `getconf LONG_BIT` → refuse with the 64-bit-OS message |
| **6 MED** | `ln -sfn` into a pre-existing **real** dir nests the link (hand-installed `~/Applications/mStream.app` kept the old version) | `link()` helper: replace symlinks, refuse real dirs (never delete the user's app; say so), move aside a real `current`, verify |
| **7 MED** | Relative `MSTREAM_INSTALL_DIR` → dangling links | Canonicalize (`cd && pwd -P` / `Resolve-Path`) |
| **10 LOW** | `wget -q` swallowed errors | `-nv` + callers name the URL |
| **16 LOW** | ps1 rewrote user PATH as REG_SZ, freezing `%VAR%` entries | Raw read + `REG_EXPAND_SZ` write |

Also: `manifest.version` validated to tag-safe chars before it becomes a path component (defense in depth — the panel judged traversal unreachable, but it's free).

**Verified**: Windows 9/9 (incl. PATH kind + `%USERPROFILE%` entry preserved unexpanded; forced replace REFUSED while running with webapp intact; missing-manifest message against a dead base **and** against real GitHub pre-backfill) · Debian 12 19/19 (incl. autostart `Exec=` re-pointed to the new folder, running-instance warning, real-dir `current`/bin handling, relative dirs, ld-musl-on-glibc, `MSTREAM_KEY`, 32-bit refusal, wget messaging) · Alpine still picks musl and runs (no false-negative from the glibc-positive change).

**Not in this batch, tracked for follow-up**: #8 uninstall path, #9 keychain cleanup + #11 launcher VersionInfo stamping (both in the merged CI — separate PR), #12 prefix-only version assert, #13 CI coverage for the manifest/installer contract, #14 `.desktop` sed escaping, #15 `--clobber` on published releases, #17 symlinked `.app` vs Launchpad, #18 `main()` wrapper.

IrosTheBeggar:
## Smoke tests — real bundles, real GitHub, product driven end to end

Both tiers run against **actual published 6.20.1 assets via `releases/latest`** (no mirror, no overrides) using this branch's scripts.

**Tier 1 — Windows (this machine): 13/13.** `install.ps1` → real 6.20.1 in 16 s → Start Menu shortcut → `current` junction, user PATH edited as REG_EXPAND_SZ → then *drove* the installed product: `mstream-server` on :3977 served **HTTP 200, the web app, and the Quick Connect API** (loopback trust) → same-version re-run kept the install + state → uninstall removed root/shortcut/PATH entry, data home untouched. Isolated from my live install throughout (`--portable`, separate port + root).

**Tier 2 — Linux in Docker: Debian 12 (glibc bundle) and Alpine 3.20 (musl bundle), 12/12 and 10/10 for the product.** Both installed 6.20.1 in ~17 s, picked the right bundle, booted via the PATH command, served the web app; `--portable` state (inside the bundle) survived a same-version re-run — the exact case fix #3 protects; clean uninstall. Two assertion mistakes on my side, both instructive: the musl bundle *does* ship a `.desktop` (server-only, `Terminal=true` — designed), and Quick Connect is correctly *off* under `--portable` (desktop-profile-only default).

**One real finding, fixed (commit e971292a):** the launcher registers ONE per-user login item by app name, so on a machine that also had a hand-run copy of mStream (my dev launcher), the installer's re-point read "enabled" (true — for the *other* copy), took the slot over, and uninstall removed it outright. Correct for a one-mStream user; surprising with two. Both scripts now read where the login item actually points and only re-point/remove one **under the install root** (or the owned `~/Applications` copy on mac); otherwise leave it byte-identical and say so with the path. Tier 1 rerun 13/13 with my live autostart untouched both ways; Debian: a foreign login item survives install and uninstall unchanged.

**Not smoked here:** macOS — the `ditto` copy into `~/Applications`, LaunchAgent re-point, Spotlight indexing. Two-minute job for the mac agent: `curl -fsSL https://raw.githubusercontent.com/IrosTheBeggar/mStream/claude/install-manifest/install.sh | sh`, then `open ~/Applications/mStream.app`, then `MSTREAM_UNINSTALL=1` the same one-liner.

PR #830, opened 2026-08-18
feat(launcher): tray status line with server uptime

IrosTheBeggar:
## What

The tray menu now leads with a greyed, unclickable **status line**, and the icon tooltip carries the same state:

| Phase | Menu line | Tooltip |
|---|---|---|
| booting / after Restart | `Starting…` | `mStream Server - starting` |
| serving | `Running · up 3h 12m` | `mStream Server - http://localhost:3000 (up 3h 12m)` |
| server died / restart failed | `Stopped · use Restart server` | `mStream Server - stopped (use Restart server)` (text unchanged) |

Until now the menu said nothing about whether the server was up — the only signal was the tooltip, which Linux (appindicator) never shows. Beyond the number itself, a suspiciously short uptime is the one visible trace of a crash→Restart cycle.

## How

- `Phase { Starting, Running { since }, Stopped }` owned by the event loop; `since` is set at `ServerUp` — "up" means *serving*, not spawned — so a Restart or crash resets the count with the phase. Transitions: Init→Starting, ServerUp→Running, ServerExited→Stopped, Restart→Starting (Stopped if the spawn fails).
- The closure's top-of-pass `ControlFlow::Wait` becomes `WaitUntil(next minute boundary)` **only while Running** — one wake a minute (the exit watcher already polls at 2 Hz), none when stopped — and `NewEvents(ResumeTimeReached)` re-renders. Aligned to the boundary rather than "60 s from now" so the shown minutes are exact. `Exit` is sticky in tao, so Quit's `ControlFlow::Exit` survives the per-pass assignment (same property the previous unconditional `Wait` relied on).
- **Minute granularity on purpose:** muda's `set_text` is live on all three OSes, but Windows doesn't repaint an *open* popup menu until it's reopened — a ticking-seconds line would lie there, and would cost a wake per second for a number nobody watches.
- `format_uptime`: `uptime(1)`-style two units — `<1m` / `45m` / `3h 12m` / `3d 4h`.
- Uptime source is the launcher's own clock: it *is* the supervisor, and the server has no uptime API. No new deps. No `launcher.log` line changes (the CI smokes grep those).
- `docs/install.md`: menu list mentions the status line (and picks up "View logs" from #827, which it was missing).

Timer-only by design — no `TrayIconEvent` handler — so it stays independent of #828; if that lands, the sole merge point is `next wake = min(popup poll, phase.next_tick())` at the top of the closure.

## Verification

- `cargo test` 14/14 (3 new: formatter, boundary math incl. early-fire self-correction, phase strings), `cargo clippy --release -- -D warnings` clean, release build OK.
- **Windows, isolated smoke** (scratch `LOCALAPPDATA`, `MSTREAM_LAUNCHER_SKIP_AUTOSTART`, own config on :3999, launched through a `wscript` shim so the launcher takes the GUI face with those env vars; the live launcher on :3000 and the HKCU Run key verified untouched):
  - tooltip read **exactly** via MSAA `IAccessible` on the notification toolbar: `mStream Server - http://localhost:3999 (up 5m)` at 354 s elapsed;
  - polled across a boundary: `6m → 7m` observed **0.2 s after** the 420 s mark;
  - killed the server: `mStream Server - stopped (use Restart server)`, launcher survived, no fatal alert (`ever_up`), log `server exited unexpectedly`.
- **Linux (Docker, ubuntu:24.04 + Xvfb + session bus), added after opening:** menu layout, the minute tick (2m→3m at t=180.6 s), Restart, crash→Stopped, Restart-from-Stopped and menu Quit all verified by reading/clicking the menu over `com.canonical.dbusmenu` — the panel's own interface. Details in the comment below.
- **Not driven here:** the Windows menu line's own rendering (right-click on the tray is tier-blocked for the automation; it's the same string from the same `show_status` call as the tooltip above, and the Linux readout shows the exact strings) and macOS (same `set_text`/`WaitUntil` primitives; CI darwin lifecycle smokes passed).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -13,9 +13,10 @@ Keep them together; the bundle itself can live anywhere.
 **Just double-click it.** The desktop face of the bundle — `mStream.exe` on
 Windows, `mStream.app` on macOS, `mstream-desktop` on Linux — starts the
 server in the background, puts an mStream icon in your tray / menu bar
-(Open mStream · Quick Connect · Start at login · Restart server · Quit), and
-opens your browser at the player. Start-at-login is on by default; one click
-in the tray menu turns it off.
+(a status line — "Running · up 3h 12m", or Starting… / Stopped — then Open
+mStream · Quick Connect · Start at login · View logs · Restart server · Quit),
+and opens your browser at the player. Start-at-login is on by default; one
+click in the tray menu turns it off.
 
 **Terminal users lose nothing.** The same desktop binary run from a terminal
 behaves exactly like the server itself (same flags, output, and exit codes) —
--- a/rust-launcher/src/tray_app.rs
+++ b/rust-launcher/src/tray_app.rs
@@ -182,6 +182,12 @@ pub fn run(args: LauncherArgs) -> ! {
     // it IS the supervisor, and the server has no uptime API to ask.
     let mut status_item: Option<MenuItem> = None;
     let mut phase = Phase::Starting;
+    // Last time the timer path re-rendered the status. The status can only
+    // change on a minute boundary, so timer wakes are throttled to 1 Hz:
+    // any other timer that shares this loop's ControlFlow (a fast poll
+    // while a popup is open, say) would otherwise re-set the same tooltip
+    // text at that timer's rate. Phase changes render unconditionally.
+    let mut status_rendered_at: Option<Instant> = None;
     let mut ever_up = false;
     let mut opened = false;
     let url = paths::server_url(&ep);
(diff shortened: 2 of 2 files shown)


IrosTheBeggar:
## Linux smoke — all paths verified on the PR's CI-built artifact

Ran the `mStream-linux-x64` bundle from this PR's build-bun run (its `mstream-desktop` is the launcher step 7 built fresh from this branch — status strings confirmed in the binary) in a bare `ubuntu:24.04` container: runtime deps only (`libayatana-appindicator3-1`, `xvfb`, `dbus-x11`, `libglib2.0-bin`), Xvfb `:99`, a `dbus-launch` session bus, `MSTREAM_LAUNCHER_SKIP_AUTOSTART=1`, `--no-open --quick-connect-off-by-default`.

The appindicator exports the menu at `/org/ayatana/NotificationItem/tray_icon_tray_app/Menu` (`com.canonical.dbusmenu`), so the labels were read with `GetLayout` and the items *clicked* with `Event(id, "clicked", …)` — the same interface a panel uses, i.e. an exact readout, not a screenshot.

| step | observed |
|---|---|
| boot | `server generation 1 spawned` → `server is up` +2 s, HTTP 200; no "tray unavailable" line (tray built) |
| menu layout | `id=2 enabled=false label="Running · up 2m"` first, then separator · Open mStream · Quick Connect · separator · Start at login (checkmark) · separator · View logs · Restart server · Quit mStream |
| tick | polled across the 180 s boundary: `Running · up 2m` at t=177.6 s → **`Running · up 3m` at t=180.6 s** |
| Restart (Event id 10) | `Starting…` +0.6 s → `Running · up <1m` +1.7 s; new server PID; log `menu: restart server` / `server generation 3 spawned` / `server is up` |
| crash (SIGKILL server) | `Stopped · use Restart server` at +0.6 s; launcher alive; log `server exited unexpectedly`; no fatal alert (`ever_up`) |
| Restart from Stopped | `Starting…` → `Running · up <1m` +1.1 s; HTTP 200 |
| Quit (Event id 11) | launcher exited, server followed via supervision (`[supervision] supervisor closed stdin — shutting down`); log `menu: quit` |

The SNI `ToolTip` isn't set on Linux (tray-icon's gtk `set_tooltip` is a no-op, as the PR body says) — the menu line is the carrier there, as intended.

Remaining un-driven: **macOS** only (same muda `set_text` / tao `WaitUntil` primitives, and the CI darwin legs' lifecycle smokes passed).

PR #833, opened 2026-08-18
reboot: keep the listening socket on same-bind reboots (fixes win-x64 smoke flake + mid-scan reboot death); single-flight ffmpeg install

IrosTheBeggar:
## What

**1. Soft reboots no longer give up the listening socket** (`src/server.js`)
- `reboot()` parks the live listener behind a 503 stub (`Retry-After: 1`) instead of `server.close()`, tears down as before, and hands the current bind (`{port, address, ssl, cert-material sha256}`) to `serveIt()`.
- `serveIt()` **keeps the socket** when the bind is unchanged — swaps the rebuilt app in for the stub, runs the post-listen chores — and only **recycles** the listener when the bind changed (address/port/http↔https/rotated cert). The replacement server is created before the old one is closed, so a bad cert still fails with `BAD CERTS` before anything is torn down.
- A **same-port relisten** (address/TLS change) retries EADDRINUSE with backoff (250 ms → 1 s → 5 s) and logs the diagnosis at ~5 s / every 30 s, instead of `process.exit(1)` after 5 s. First boots and moved ports still exit immediately as before.

**2. ffmpeg-bootstrap: one install per reboot, not one per reboot *plus* one per re-serve** (`src/util/ffmpeg-bootstrap.js`)
- `downloadAndInstall()` is single-flight per target dir (`_install = {dir, promise}`), deliberately not cleared by `reset()`; `checkForUpdate()` shares it. A changed `ffmpegDirectory` mid-download gets its own install.
- `reset()` bumps a generation counter so a superseded resolution chain neither publishes resolved state nor nulls its successor's `_initPromise` (the old code could — one reboot could turn into a *third* download).
- `MSTREAM_FFMPEG_MIRROR` = base URL for the BtbN assets + `checksums.sha256` (and the darwin martin-riedl zips); plain http accepted for **loopback only**, everything else stays https-only. Doubles as an air-gapped-mirror knob and the unit test's seam.

**3. Smoke** (`.github/workflows/build-bun.yml`, win-x64 leg's step 4/5): step (4) now fails closed on `keeping the listening socket` present / `retrying listen` absent (positive proof, not just survival); step (5) accepts a `503` for the *second* of its two concurrent `/config/address` POSTs (it arrived inside the reboot window — under the old code that interleaving was a `000` FAIL nobody had hit yet; all four recent green runs saw sequential reboots) while still requiring one `200` and failing on 400/404/000.

**Tests**: `test/integration/reboot-keeps-listener.test.mjs` (Node + Bun legs; `MSTREAM_TEST_BUN_BIN` for machines where npm's `bun` shim isn't spawnable) hammers fresh connections through the reboot window (master: stream of `ECONNREFUSED`; now: only 503/200), asserts the ring-buffer log lines, and covers the address-change recycle. `test/unit/ffmpeg-bootstrap.test.mjs` is hermetic on all three OSes (loopback mirror serving a slow fake archive with wrong checksums, so nothing is ever installed) — with the single-flight disabled it fails with "archive fetched 2x".

## Why

The win-x64 `build-bun` smoke ([run 32094833102, job 95584019673](https://github.com/IrosTheBeggar/mStream/actions/runs/32094833102/job/95584019673)) failed at step (4): after the trust-proxy POST the relisten got EADDRINUSE for its whole 20×250 ms budget → `Unable to start mStream: port 3399 is already in use` → exit. Root cause: **Bun ≤ 1.3.14 on Windows creates inheritable socket handles**, so every spawned child holds a duplicate of the listen socket and `close()` in the parent releases nothing until that child exits (`netstat` keeps showing the port `LISTENING` under the dead listener; the kernel even completes handshakes that then hang). In the smoke the child was the ffmpeg download's PowerShell extractor — alive at the reboot only when the 130 MB download finished fast, hence intermittent. Fixed upstream in [oven-sh/bun#36938](https://github.com/oven-sh/bun/pull/36938) (merged 2026-08-05) but **not in any release** (1.3.14 = latest); CI's `bun-version: latest` and the shipped 6.20.x Windows binaries carry it.

Production impact was real, not just CI: with the staged binary, **saving any reboot-requiring admin setting during a scan killed the whole server** (the `rust-parser` child held the port → 5 s → exit) — same for a transcode, backfill or enrichment worker alive at that moment. Keeping the socket sidesteps the whole class on every runtime; the patient relisten covers the residual bind-change case; the Bun bump when it ships removes the runtime cause on top.

## Reviewer notes

- Verified with the rebuilt win-x64 bundle: reboot with the extractor alive → 70 ms on the same listener; reboot mid-scan → 95 ms, scan keeps running; address change with a live child → 11 EADDRINUSE retries over 6 s then up (old code would have died at 5 s). The workflow's smoke script run verbatim locally: `SMOKE OK` twice (once coalesced, once the 200/503 interleaving), one "Downloading ffmpeg" per boot across its three reboots (was three).
- Full `npm test`: 2271 pass, 0 cancelled (only failures = a pre-existing untracked local test file for routes not on master; not part of this PR). Lint clean on changed files (one pre-existing `require-await` warning in ffmpeg-bootstrap remains).
- Behavior change to be aware of: during a same-bind reboot, clients now get a **503** with `Retry-After` instead of connection-refused; a WebSocket upgrade attempt inside the window is dropped silently under Bun (no `'upgrade'` listener) — same net effect as before. The admin UI only toasts on reboot-triggering saves (no polling), so nothing there depends on refusals.
- Not done / parked: parking requests during the reboot window and replaying them into the new app instead of answering 503 (would let admin double-saves succeed; needs Bun body-stream validation).

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -543,6 +543,21 @@ jobs:
               echo "FAIL: no re-serve observed (boot lines ${BOOTS_BEFORE} -> ${BOOTS_AFTER}) — the reboot never happened"
               REBOOT_OK=0
             fi
+            # A same-bind reboot must KEEP the listening socket, never
+            # close()+listen() it. On Windows a live child (here: the ffmpeg
+            # download's PowerShell extractor, alive at this point on a fast
+            # runner) holds an inherited handle to the old listen socket under
+            # Bun <= 1.3.14 (oven-sh/bun#36936), so a relisten sits in
+            # EADDRINUSE until that child exits — this exact step used to die
+            # of it intermittently. Positive proof, not just survival.
+            if ! grep -q "keeping the listening socket" "$RUN/boot.log"; then
+              echo "FAIL: the same-bind reboot did not keep the listening socket (see boot.log)"
+              REBOOT_OK=0
+            fi
+            if grep -q "retrying listen" "$RUN/boot.log"; then
+              echo "FAIL: the same-bind reboot relistened (retrying listen seen) — the socket should never have been released"
+              REBOOT_OK=0
+            fi
 
             # (5) overlapping reboots must not kill the process. Two rapid admin
             # saves used to start two serveIt()s racing for the port; the loser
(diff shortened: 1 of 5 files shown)


IrosTheBeggar:
## Independent smoke: negative control on master, positive on this branch — production scenario, real bits

Built the win-x64 server from **master** and from **this branch**, pointed both at a real 30,000-file library (so the Rust scanner child outlives the reboot), and fired a reboot-triggering admin save (`POST /admin/config/trust-proxy`) **the instant `rust-parser` was alive** (`scanning:true`, child PID confirmed alive at the POST). Same harness, same library, same port; sampled the port every 250 ms for 20 s.

**Negative control (master, Bun 1.3.14 Windows) — outage reproduced exactly:**
```
Port 3977 not released yet after reboot — retrying listen (1 left)
Port 3977 not released yet after reboot — retrying listen (0 left)
Unable to start mStream: port 3977 is already in use
mStream fatal: port 3977 already in use
```
→ **server process dead 8 s after a routine setting save**, port samples `0x3 200x3` then nothing. That's the shipped 6.20.x Windows behavior for anyone who saves a reboot-requiring setting during a scan.

**Positive (this branch), identical scenario:**
```
Reboot: bind unchanged (http://127.0.0.1:3977) — keeping the listening socket
```
→ back serving in ~54 ms (263 ms at my sample resolution), **process alive, HTTP 200, and the scan kept running through the reboot** (`scanning:true` after). Port samples `0x1 200x79`.

Follow-up on the same branch: **two admin saves mid-scan** (5 s apart) → both `200`, the second correctly coalesced by `rebootInFlight` (one "Rebooting Server" line), and the **scan completed to all 30,000 tracks 29 s later** with the process alive and serving.

Also checked (as the launcher's author): the 503 stub is invisible to the desktop launcher — its health prober only runs until the first `ServerUp`, the exit watcher is process-level (`try_wait`), and the stub never exists at first boot. `--supervised` untouched. Merges cleanly onto master; CI 19/19 incl. full-ci.

Verdict: the bug is real and shipped, the fix works on real bits, and the design is the right long-term shape (it stays correct after Bun ships #36938 — at that point only the Bun-attributing comments/diagnostic need softening). Recommend merging into 6.20.2.

IrosTheBeggar:
## Linux smoke (Docker, real linux-x64 build of this branch, 30k-file library) — 13/13

Both reboot paths hand-driven with the `rust-parser` child confirmed alive at each trigger:

**A. Same-bind reboot mid-scan** → `keeping the listening socket`; **40/40 port samples 200** through the window (not even a 503 — the swap is that fast on Linux); process alive; scan kept running.

**B. Port change mid-scan (3977 → 3978, the recycle + relisten path)** → `Reboot: bind changed … recycling the listener` → `Reboot moved the port 3977 -> 3978; a conflict on the new port is real, not a release delay` (correct branch: a moved port must not retry) → **serving on 3978 in ~316 ms, old port released, process alive**. No EADDRINUSE on Linux — the port frees immediately, so the patient-retry code stays dormant here and only matters on Windows, as designed.

**C.** Scan completed to all **30,000 tracks** after both reboots; process alive at the end.

Cross-platform status: Windows outage reproduced on master / fixed here (same-bind); Linux both paths verified on real bits; macOS covered by the green integration test on both runtimes in CI (POSIX socket semantics identical to Linux). Nothing left that I'd hold the merge for.

PR #836, opened 2026-08-18
build: one Windows VersionInfo contract for every PE the win-x64 bundle ships

IrosTheBeggar:
Piece #2 of the SignPath plan — useful with or without signing, and needs no account decisions.

## What was wrong (read from the shipped 6.20.0 zip)

| PE | before |
|---|---|
| `mStream.exe` (launcher) | strings `6.20.0` ✓ but fixed-info **`0.1.0.0`** (winresource's default from Cargo.toml), no CompanyName |
| `mstream-server.exe` (Bun) | ProductName **"mStream Server"**, FileDescription "music streaming server" |
| `bin/rust-parser/…exe` | **no VersionInfo at all** |
| `bin/rust-server-audio/…exe` | **no VersionInfo at all** |
| `bin/iroh/….node` | none — but this is `@number0/iroh`'s own prebuilt (upstream), left alone |

Uniform metadata across shipped binaries is a precondition of the SignPath Foundation route (artifact-configuration restrictions enforce product name/version per signed file) and is what users see in Explorer's Details tab / Task Manager.

## The contract (identical on every PE we build)

ProductName `mStream` · ProductVersion = FileVersion = package version as **four** numeric parts, in the string table **and** the fixed-info numbers · CompanyName = package author · LegalCopyright `<author> (<license>)` · FileDescription per file (Task Manager name).

Three sources, each where it belongs:
- **server** — Bun `--windows-*` flags, now driven by the same `canonicalFields()` the stamper uses ([build-bun.mjs](scripts/build-bun.mjs)).
- **launcher** — [rust-launcher/build.rs](rust-launcher/build.rs): 4-part strings, real fixed-info via `set_version_info`, CompanyName, `OriginalFilename mStream.exe`, en-US block. Its baked version is already asserted against the tag by #835.
- **rust sidecars** — stamped **at bundle time** onto the staged copies by the new [scripts/win-versioninfo.mjs](scripts/win-versioninfo.mjs) (resedit, pure JS — the approach Bun uses internally). Deliberately *not* a `build.rs` per crate: they're CI-committed, version-agnostic tools, and a build-time bake would put them in the exact staleness class #835 just closed for the launcher, plus force four sidecar workflows to rebuild+commit five targets each per release — for a Windows string. `bin/` itself is untouched.

## The assert

[scripts/check-win-versioninfo.ps1](scripts/check-win-versioninfo.ps1) reads every PE **the way Windows does** (`Get-Item ….VersionInfo` = Win32 version API — the view Explorer/Task Manager/signing tools get; resedit's own read-back can agree with itself while Windows shows a different string table, which I hit while prototyping). Strict on tags; on PR builds a version-only mismatch on the committed launcher is a warning (a version-bump PR builds against the launcher baked for the previous version by design). **Any PE the script doesn't know fails the build** — nothing ships blank (and later unsigned) by omission. The win-x64 leg runs it right after bundling, then starts the stamped `rust-server-audio` and GETs `/status` — the one shipped binary no other smoke ever executed.

`resedit` is a devDependency; the prod-only CI install adds just it under `scripts/node_modules` (`npm install --prefix scripts --no-save resedit@3.0.2` — a plain `--no-save` install reconciles the whole tree and pulls every devDependency in; `--omit=dev` skips the named package itself; both verified).

## Verified locally (Windows)

- Launcher rebuilt with the new build.rs → `mStream / 6.20.2.0 / rawFV 6.20.2.0 / Paul Sori / en-US`.
- Full win-x64 bundle: both sidecars stamped; **strict check passes on all four PEs**; the same check **fails on the shipped 6.20.0 bundle** for exactly the defects above (negative control, PS 5.1).
- Stamped `rust-server-audio` serves `/status`, stamped `rust-parser` runs, `mstream-server -V` → 6.20.2, staged launcher `--autostart=status` answers; resedit output verified executable under both node and bun.

CI on this PR: the win-x64 leg builds the launcher fresh (PR touches `rust-launcher/`), stamps, asserts (lenient), and every existing smoke then runs on the stamped/baked bytes.

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -225,6 +225,18 @@ jobs:
           "BUN_COMPILE_EXEC_PATH=$exe" | Add-Content -Path $env:GITHUB_ENV
           Write-Host "runtime: $exe"
 
+      # The bundler stamps Windows VersionInfo onto the staged rust sidecars
+      # with resedit (scripts/win-versioninfo.mjs). It's a devDependency, and
+      # this job installs prod-only, so add JUST that package under
+      # scripts/node_modules — a plain `npm install --no-save resedit` would
+      # reconcile the whole tree and drag every devDependency (eslint,
+      # redocly, ...) in with it; `--omit=dev` would then skip resedit itself.
+      # The bundler resolves it from scripts/ (its own directory) first.
+      # Version pinned to match package.json's devDependency — bump both.
+      - name: Install the VersionInfo stamper (win-x64)
+        if: matrix.key == 'win-x64'
+        run: npm install --prefix scripts --no-save --no-audit --no-fund resedit@3.0.2
+
       # Desktop tray launcher: the bundler stages bin/rust-launcher/<name> as
       # the bundle's double-click face (mStream.exe / .app executable /
       # mstream-desktop). Those binaries are CI-committed on master pushes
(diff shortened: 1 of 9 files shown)


IrosTheBeggar:
Stacked on #837 (merged its commit into this branch): the first two CI runs here hit the overlapping-reboot `200 / 000` race — once on darwin-arm64, once on win-x64 — which #837 root-causes (step 5's address save was a bind CHANGE, so #833 recycled the listener by design) and fixes. The VersionInfo step itself was green on both runs: `VersionInfo check OK: 4 PE(s) match the contract, 0 warning(s)`, and the stamped `rust-server-audio` exec check now accepts the headless runner's audio-device probe as proof the PE runs. Merge order: #837 first, then this.

IrosTheBeggar:
Adversarial-review batch applied in 49492abd (review items 7-12): stamper install pinned (resedit + pe-library) with `--ignore-scripts` and **bound to package-lock.json** (version+integrity+resolved compared post-install; red on drift), PR paths filter now covers the stamper/checker/package-lock.json, launcher leniency widened to every package.json-derived field (kills the post-merge red window) but **strict whenever the launcher was rebuilt from the PR** (`LAUNCHER_REBUILT`), CLI read-back survives the Bun exe, `powershell -File` works under 5.1, .npmignore mirrored, ffmpeg runtime downloads recognized, list-tying comment. All verified locally (positive + negative on the lock check; lenient/strict against the committed pre-PR launcher; -File under 5.1).

PR #839, opened 2026-08-18
reboot: keep Quick Connect, federation, the separate-port listeners and every admin save alive across a soft reboot

IrosTheBeggar:
Batch B of the 2026-08-18 adversarial review (findings 1–6, all confirmed 3/3 with reproductions). Server code only; every fix has a test on **node and bun**, plus a smoke on the compiled Windows Bun bundle with a live scanner child.

## 1 · HIGH — a soft reboot silently killed Quick Connect (and federation)
`reboot()` fired `iroh.stop()` / `federation.stop()` fire-and-forget while the #833 kept-socket re-serve now runs ~70 ms later; `endpoint.close()` takes ~1 s, so `onListening → start()` found `endpoint` still set, no-oped, and the late stop nulled it. After **any** reboot-requiring admin save, the tunnel was dead until the next process restart, no error logged (reproduced end-to-end by the review, and by the new test before the fix). Now every endpoint/sidecar teardown is sequenced into the re-serve gate (5 s latency bound), and `iroh.js`/`federation.js` keep an in-flight `stopping` promise, clear their public state up front, and make `start()` wait on it — the pattern `discovery-p2p-stack.js` already used.

## 2 · MED — concurrent admin saves could exit the process, and lost each other's changes
`saveFile` was a plain `fs.writeFile` racing the reboot's config re-read (2.7–4 % of concurrent save pairs killed the server) — and every editor is `loadFile → mutate → saveFile`, so the second save overwrote the file with a document loaded before the first (a `trustProxy` flip lost under a `maxRequestSize` save, both answered 200 — the new test caught this before the fix). New [`util/atomic-json.js`](src/util/atomic-json.js): temp+rename writes, serialized per file, `updateJsonAtomic()` for read-modify-write as one step. `loadFile` snapshots what it hands out; `saveFile` merges only the caller's changed paths onto the current file when another write landed in between. 72 call sites, zero changes to them.

## 3 · MED — separate-port Subsonic/DLNA died on the Windows Bun bundle mid-scan
They still `close()+listen()`ed on every reboot; the re-listen hit EADDRINUSE against the scanner child's inherited handle and the module ref was nulled — Subsonic dead for every mobile client while the tray said Running. New [`util/kept-listener.js`](src/util/kept-listener.js): keeps the socket when port/address are unchanged, recycles with the main listener's same-port patience otherwise; `reboot()` no longer stops them, `onListening` stops them when no longer wanted.

## 4 · LOW — coalesced reboot dropped a late save
Remembered and re-run once — only if a config write actually completed after the in-flight reboot's read, so an already-applied save doesn't cost a second bounce.

## 5 · LOW — unservable bind change exited the process (and every later start)
`serveIt` now **probes** the new bind before giving up the working socket (exact bind for a port move, address-only for a same-port move; `node:net`, because Bun's `node:http` reports a non-local address as EADDRINUSE while its `node:net` says EADDRNOTAVAIL like Node). On failure the change is refused: previous port/address restored in the config file, kept socket keeps serving, zero downtime. Residual races roll back to the previous bind instead of exiting or spinning.

## 6 · LOW — the +1 s sweep cut new-app responses on kept keep-alive sockets
Requests tag their socket with the app generation; the sweep destroys only sockets still busy with an old-generation response.

## Verification
- New unit tests: atomic-json (torn reads/ordering/tmp cleanup), kept-listener (keep/recycle + deterministic retry logic via a fake server), loadFile/saveFile merge semantics.
- New [`reboot-keeps-services.test.mjs`](test/integration/reboot-keeps-services.test.mjs) on **node + bun**: tunnel up again with the same endpoint id; Subsonic port never falters and logs `kept`; two overlapping saves both applied on disk and live; unservable address refused before the recycle, file reverted. Run 3× under concurrent load, stable.
- Existing `reboot-keeps-listener` suite green on both runtimes; full suite 2285/2292 (rest skipped / pre-existing).
- **Compiled Windows Bun bundle**, 30k-file scan child alive, Subsonic separate port + QC on, admin save mid-scan: main `200×37 503×3`, Subsonic `200×40` (`Separate server kept across reboot`), iroh back with the same endpoint id, process alive.

### DIFF
--- a/src/dlna/dlna-server.js
+++ b/src/dlna/dlna-server.js
@@ -1,17 +1,17 @@
 import express from 'express';
-import http from 'node:http';
-import winston from 'winston';
 import * as config from '../state/config.js';
 import * as dlnaApi from '../api/dlna.js';
 import { serveAlbumArtFile } from '../api/album-art.js';
 import { timeSeekMiddleware } from './time-seek.js';
 import { resolveLibraryMediaPath } from './media-path.js';
+import { createKeptListener } from '../util/kept-listener.js';
 
-let dlnaServer = null;
+// Kept listener (util/kept-listener.js): start() after a soft reboot keeps
+// the socket when port/address are unchanged and recycles it (with same-port
+// EADDRINUSE patience) when they changed — the main listener's rule.
+const listener = createKeptListener('dlna');
 
-export function start() {
-  if (dlnaServer) { return; }
-
+function buildApp() {
   const app = express();
 
   // Time-seek (TimeSeekRange.dlna.org) handler runs first; it calls next()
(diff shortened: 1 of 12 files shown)


IrosTheBeggar:
Follow-up in `d48f2bf3`: dedicated test for fix 6 (keep-alive socket reused by the new app survives the +1 s sweep; negative control against the old sweep fails as expected), plus a **Linux Docker smoke** of the compiled linux-x64 server with a 30k-file scan child: 22/22 — Subsonic separate port 40/40 across the reboot and kept, both concurrent saves applied (and the pending re-run mechanism fired for real: `A settings change landed after this reboot read the config — rebooting once more`), unservable address refused with EADDRNOTAVAIL before any recycle + config reverted, genuine port move still recycles, scan completes, process alive. Windows compiled-bundle smoke earlier: same shape (main 200×37/503×3, Subsonic 200×40 kept, iroh back with the same id).

PR #840, opened 2026-08-18
ci: make the launcher binaries commit job race-proof and honest (review 13, 14, 18)

IrosTheBeggar:
Batch C1 of the adversarial review — the findings that could strand or mis-stamp a **release**. Workflow-only (+ `docs/deploy.md`) → merge via git.

## 13 · MED — queued launcher runs conflicted on rebase → master stranded
Runs are serialized, but a queued run still checks out its **older** trigger sha; `git pull --rebase` of its binaries commit onto a tip that already carried a newer run's binaries conflicted on every file (binaries don't merge), the job went red, and master kept whatever it had — exactly the "merge a launcher fix, then push the bump" shape the release doc prescribes. **Now the commit job never rebases:** fetch the tip → **stand down** if launcher inputs newer than the trigger sha already landed there (that push's own run owns `bin/`; its build is a superset) → else commit our set on top of the tip and push → on rejection drop the commit and retry from the new tip.

## 14 · MED — the stamp didn't know whether all four binaries were downloaded
`download-artifact` with no name silently returns whatever still exists; the `git add` glob then staged the checked-out (previous) binaries for missing legs under a fresh stamp — a next-day "Re-run failed jobs" past the 1-day retention did exactly that. Now: scratch download dir + inventory assert (all four, non-empty, else red with "re-run ALL jobs"), retention 7 days.

## 18 · LOW — three loose ends
- Windows build leg now reads the built exe's VersionInfo back via the Win32 API and asserts the contract (`build.rs` only warns on a failed resource compile).
- On **tags**, a missing committed launcher is a red run, not an in-job rebuild (that would ship the ubuntu glibc-2.34 floor).
- The two icons the launcher embeds live outside `rust-launcher/` — now push/PR triggers **and** stamp lines 3–4, asserted at tag time; the workflow file is a push trigger too so a stamp-format change re-lands the set. `docs/deploy.md`: tag the binaries commit **by SHA**, re-run all jobs.

## Verification
- The commit step and the tag assert were **extracted verbatim** from the workflows and run in local git sandboxes: commit job **15/15** across six scenarios (normal · no-op re-run · queued-older-run stands down then the bump's run lands 6.20.3 · tip moved by an unrelated commit, no rebase · push rejected once, retry lands on the sneaked-in commit · incomplete artifact set refused); tag assert: 4-line match ok, legacy 2-line fails as `<unstamped>`, icon mismatch fails, missing binary fails.
- yaml-lint clean. The Windows VersionInfo assert can only run on the real leg — it's the same read-back `scripts/check-win-versioninfo.ps1` already proves in build-bun CI.

## After merging
The workflow-file trigger fires `Build Rust Launcher` on the merge and it re-lands the binaries with the **4-line** stamp right away (no waiting for an unrelated launcher push); until that commit lands, a tag would fail on `<unstamped>` icons by design.


IrosTheBeggar:
**Smoked on the real Actions plumbing** by dispatching this branch's `Build Rust Launcher` (dry_run=false) on the branch itself: all four legs built, the new Windows VersionInfo assert printed `ProductName='mStream' ProductVersion='6.20.2.0' FileVersionRaw=6.20.2.0 CompanyName='Paul Sori'` and passed, and the rewritten commit job ran end to end — inventory of all four artifacts, 4-line stamp from the trigger sha, commit on the tip, `pushed c35a7db onto claude/launcher-commit-job`. That bot commit is kept in the PR on purpose: the merge then carries the 4-line stamp itself, so there is no window where master has the new tag assert but a 2-line stamp. (Sandbox verification of the step's six race scenarios is in the description.)

PR #841, opened 2026-08-18
ci: sidecar provenance stamps + tag assert, lockfile-bound iroh cross-fetch, --locked + pinned cross (review 15–17)

IrosTheBeggar:
Batch C2 of the adversarial review — the supply-chain items, i.e. the "built from source in a verifiable way" story the SignPath Foundation terms will hold us to. Workflow files + `scripts/build-bun.mjs` → merge via git.

## 15 · sidecar provenance (rust-parser, rust-server-audio)
The launcher has had a provenance stamp + tag assert since #835; the sidecars shipped in every tag with **nothing** tying them to the tag's source (a PR touching `rust-parser/` merged just before the bump shipped the previous scanner green; a hand-committed binary would pass). Each sidecar commit job now writes its trigger sha's source-tree hash next to the binaries — `bin/rust-parser/.source-tree` (glibc workflow), `bin/rust-parser/.source-tree-musl` (musl workflow, its own file so the two never rebase-clash on one path), `bin/rust-server-audio/.source-tree` — and build-bun's tag build asserts all three against HEAD on **every** leg. The rust-parser and rust-server-audio workflows now also trigger on their own workflow file, so merging this re-lands the sets with stamps immediately.

## 16 · `stageIroh` cross-fetch bound to the lockfile
For linux-arm64/musl bundles the bundler `npm pack`s `@number0/iroh-<triple>` and stages its `.node` — pinned only by the registry's own packument. Version **and** sha512 integrity now come from `package-lock.json` (the same pin `npm ci` enforces for the host triple); mismatch = fatal in CI, warn+skip locally. Reproduced: real linux-arm64 cross-bundle → `matches package-lock.json integrity (sha512)`; tampered lock → local WARN + no `.node`, `CI=1` → FATAL exit 1.

## 17 · `--locked` everywhere + pinned, hashed cross
All four sidecar workflows pulled cross-rs from `releases/latest` straight into `~/.cargo/bin` and built without `--locked`. Now `cross v0.2.5` with its sha256 (downloaded twice, identical) and `--locked` on every cargo/cross build (`cargo metadata --locked` passes for all three crates today, so it's free). p2p-sidecar isn't in bundles but is CI-committed and shipped via npm → gets both too (no stamp: not in the tag path).

## Verification
- yaml-lint on all five workflows; eslint on the bundler.
- Tag assert extracted verbatim and exercised: all match → ok; missing musl stamp → fails naming the workflow to dispatch; stale server-audio stamp → fails.
- stageIroh path exercised for real on this host (positive + two negative controls, above).

## After merging
The workflow-file triggers fire the sidecar builds on the merge; their `ci: update pre-built …` commits carry the first stamps. A tag before those land fails on "no provenance stamp … dispatch build-rust-parser" by design. Left for a follow-up: the sidecar commit jobs still use the rebase-retry pattern that #840 replaced for the launcher (they only clash when two overlapping pushes hit the same family).


IrosTheBeggar:
**Smoked by dispatching all four sidecar workflows on this branch.** First pass: every build passed (`--locked` + the pinned/hashed cross: `cross.tar.gz: OK`), but the three stamp-writing commit jobs failed at `git add` — `.gitignore`'s `bin/rust-parser/*` / `bin/rust-server-audio/*` rules swallowed the new stamp files (the launcher stamp only worked because it had its own negation). Fixed in `089418b4` (negations in `.gitignore`, mirrored in `.npmignore`); re-dispatched → all three green. The branch now carries the bot commits: `.source-tree` / `.source-tree-musl` / `rust-server-audio/.source-tree` (verified equal to the branch's own `rust-parser` / `rust-server-audio` tree hashes), the p2p-sidecar rebuild, and rebuilt sidecar binaries where the newer toolchain produced different bytes — the same rebuild the post-merge self-trigger would do, landing atomically with the merge instead. Three commit jobs raced on the branch and all converged. Also confirmed the `stageIroh` lockfile-integrity line fires on the PR's own CI cross legs (`matches package-lock.json integrity (sha512)` on linux-arm64 and linux-arm64-musl).

PR #842, opened 2026-08-19
fix(launcher): status tick on Linux, View-logs path quoting, unverified-server phase, idempotent disable (review E1–E4)

IrosTheBeggar:
Batch E of the adversarial review — the four launcher findings (all LOW), Rust only.

## E1 — the minute status tick never fired on its own on Linux
`ControlFlow::WaitUntil` schedules nothing on tao's gtk backend (`run_return` blocks in `gtk::main_iteration_do` until some *other* event wakes the loop), so on Wayland the menu's `up 3h 12m` line went stale for hours; X11 only looked fine because its device thread woke the loop on mouse motion. The tick now comes from a per-generation thread that sleeps to each minute boundary and nudges the loop with a generation-stamped `Tick` **user event** — the same `proxy.send_event` path that already carries `ServerUp` and menu clicks (which #830's Linux Docker smoke proved on gtk). `ControlFlow` stays plain `Wait`; the old `ResumeTimeReached` arm and `next_tick()` are gone.

## E2 — Windows "View logs" broke on legitimate paths
The logs dir was interpolated into a single-quoted PowerShell string (only ASCII apostrophes escaped — a curly apostrophe in a user name like `O'Brien` ended the literal early) and passed to the wildcard `-Path` of Set-Location/Get-Content (`[`/`]` made the `cd` fail, and the reads then ran against the launcher's cwd). Now the dir travels in an env var and every file is an absolute `-LiteralPath` via `Join-Path`, plus `-NoProfile`. **Verified:** the old shape errors on `…\O'Brien [work]\logs`; the new one tails that path and a plain one.

## E3 — a server the probe couldn't verify said "Starting…" forever
The identity probe is a plaintext GET; an SSL-terminated config never answers it, so after `BOOT_TIMEOUT` the tray said "Starting…" for the whole session, inviting a needless Restart. The prober now emits `ProbeGaveUp` when it times out with the child still alive; the loop shows a distinct `Phase::Unverified` (`Running (unverified) · up …`) and marks the server as having been up. **Verified on Windows** with an SSL config: tooltip read `(up 1m, not verified by the launcher)` and ticked 1m→2m→3m.

## E4 — `--autostart=disable` was not idempotent on Windows
auto-launch's `disable()` is a bare `delete_value` that fails `os error 2` when the Run value is absent (Linux/macOS return Ok), so an uninstaller/script got exit 1 for a state that already held. `set_enabled` now treats "not enabled after the attempt" as success for a disable. **Verified:** the committed launcher exits 1 disabling-when-absent, the fixed one exits 0; a present value is still removed.

## Verification
`cargo build --locked`, `clippy -D warnings`, and the crate's unit tests (incl. the new Unverified phase texts) all pass. E2/E3/E4 smoked on Windows as above (E3 read via MSAA). A Linux runtime tick observation is running; E1's `Tick` rides the identical proxy path #830 already exercised on gtk for `ServerUp`/menu events.

Source-only (CI owns the binaries).

### DIFF
--- a/rust-launcher/src/autostart.rs
+++ b/rust-launcher/src/autostart.rs
@@ -46,12 +46,15 @@ pub fn set_enabled(on: bool) -> Result<(), String> {
         Ok(()) => Ok(()),
         // Disable is idempotent: on Windows auto-launch's disable() is a bare
         // `delete_value` and fails with "cannot find the file specified (os
-        // error 2)" when the Run value (or, on a fresh profile, the Run key)
-        // doesn't exist — while Linux/macOS return Ok for a missing file. An
-        // uninstaller or script asking for a state that already holds must
-        // not get exit 1. Judge by the outcome: if the item is not enabled
-        // after the attempt, the request is satisfied. (An unreadable state
-        // stays an error — nothing was proven.)
+        // error 2)" when the Run value doesn't exist — while Linux/macOS
+        // return Ok for a missing file. An uninstaller or script asking for
+        // a state that already holds must not get exit 1. Judge by the
+        // outcome: if the item is not enabled after the attempt, the request
+        // is satisfied. (An unreadable state stays an error — nothing was
+        // proven. That includes a profile with no Run KEY at all: the
+        // crate's is_enabled() ?-propagates that open_subkey failure rather
+        // than reporting "disabled", so this guard deliberately does not
+        // vouch for it.)
         Err(_) if !on && !a.is_enabled().unwrap_or(true) => Ok(()),
         Err(e) => Err(e.to_string()),
     }
--- a/rust-launcher/src/autostart.rs
+++ b/rust-launcher/src/autostart.rs
@@ -42,8 +42,19 @@ pub fn set_enabled(on: bool) -> Result<(), String> {
     if on {
         ensure_autostart_parent_dir();
     }
-    let r = if on { a.enable() } else { a.disable() };
-    r.map_err(|e| e.to_string())
+    match if on { a.enable() } else { a.disable() } {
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
**E1 confirmed on gtk (Linux runtime trace)**

The one finding whose real risk is Linux-specific: does a timer-thread `proxy.send_event(Tick)` actually wake tao's gtk loop, or does it sit blocked in `gtk::main_iteration_do` the way `ControlFlow::WaitUntil` did? Traced **both** sides in a container-local copy of the crate (worktree untouched), ticker accelerated to 5 s for density (the minute-boundary math is separate and unit-tested; the wakeup is what's under test). `control_flow` is always `Wait`, so nothing but an event can advance the loop:

```
TICKTRACE sent gen=1   <- ticker thread -> proxy.send_event(AppEvent::Tick)
TICKTRACE recv gen=1   <- gtk loop dispatches the AppEvent::Tick arm
TICKTRACE sent gen=1
TICKTRACE recv gen=1
... 6 send->recv pairs in 35 s, perfectly interleaved
counts: sent=6  recv=6
```

Every send was received — the cross-thread proxy wakes gtk on each tick. (`rust:latest`, `libgtk-3`, `Xvfb`, `dbus`.)

Independently, the same run's `server is up` line is emitted only by the `AppEvent::ServerUp` arm — i.e. the probe thread's `proxy.send_event(ServerUp)` also woke the gtk loop, the identical mechanism reached via a different event.

Windows side was already verified live (status line ticked 1m -> 2m -> 3m on an SSL config, plus the give-up line in `launcher.log`). E1 is now directly confirmed on both backends.

PR #845, opened 2026-08-19
feat(installer): Windows Inno Setup installer (P3.4a — native installers)

IrosTheBeggar:
First leg of P3.4 (native installers): a per-user Windows setup.exe built by the win-x64 leg from the staged bundle, attached to releases next to the zips.

## Design (details in `build/windows-installer.iss`)
- **Per-user, no UAC** — installs to `%LOCALAPPDATA%\Programs\mStream`, the exact location `install.ps1` uses, so the two install methods never fork. Start Menu shortcut always; desktop shortcut opt-in.
- **Payload = the staged bundle verbatim**, compiled **after** the VersionInfo assert and every smoke — the setup.exe wraps exactly the bytes that were proven. Empty scaffold dirs are skipped on purpose (the server self-creates missing dirs at boot; an installed copy keeps data in `%LOCALAPPDATA%\mStream`, not the install dir).
- **Autostart stays the launcher's job** (self-enables on first tray run, user toggle respected). The **uninstaller removes the HKCU Run value only when it points into the install dir** — the login item is one per-user registration by app name, and uninstalling one copy must never disable another's (the exact footgun that bit an install-script smoke on 2026-08-18).
- **Upgrades/uninstalls stop only processes running from the install dir** (WMI `ExecutablePath` prefix match — never a name-wide taskkill). Launcher first, so the supervision pipe takes the server down with it; a second sweep catches stragglers.
- **User data never touched** by uninstall.
- **VersionInfo contract**: the setup.exe carries the same ProductName/4-part version/Company as every shipped PE; the CI step asserts it via Win32 truth.
- **SignPath-ready**: compile is a discrete step, so PE signing slots in before ISCC and setup.exe signing after.
- `manifest.json` stays zips-only by design (its consumers — install scripts, future auto-updater — eat zips; the generator's glob already ignores the exe). The release job uploads the setup.exe alongside.

## Verification so far
- YAML valid; `windowsVersion` helper produces the 4-part string (`6.20.2.0`).
- This PR's own CI compiles the installer on the win-x64 leg (ISCC ships on the runner image) and asserts its VersionInfo.
- **Next (tonight): local silent-install E2E** — download the PR artifact, then on real Windows: silent install to a scratch dir, assert layout + Start Menu shortcut + uninstall registration, run the installed launcher's CLI face, then silent uninstall asserting clean removal and that a foreign HKCU Run value (pointing outside the install dir) survives while an in-dir one is removed. Results will be posted here.

Deferred within P3.4: Linux deb/rpm (next PR), macOS .pkg (waits for notarization secrets — unsigned pkgs get brutal Gatekeeper treatment).

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1295,10 +1295,13 @@ jobs:
           Write-Host "installer: $($setup.Name) ($([math]::Round($setup.Length/1MB,1)) MB)"
           # The installer exe carries the same 4-part VersionInfo contract as
           # every shipped PE (Win32 truth, like check-win-versioninfo.ps1).
+          # Inno space-pads the VersionInfo string fields to a fixed width —
+          # trim before comparing.
           $vi = $setup.VersionInfo
-          Write-Host "VersionInfo: FileVersion=$($vi.FileVersion) Product=$($vi.ProductName) Company=$($vi.CompanyName)"
-          if ($vi.FileVersion -ne $ver4) { Write-Host "::error::installer FileVersion '$($vi.FileVersion)' is not '$ver4'"; exit 1 }
-          if ($vi.ProductName -ne 'mStream') { Write-Host "::error::installer ProductName '$($vi.ProductName)' is not 'mStream'"; exit 1 }
+          $fv = "$($vi.FileVersion)".Trim(); $pn = "$($vi.ProductName)".Trim()
+          Write-Host "VersionInfo: FileVersion='$fv' Product='$pn' Company='$("$($vi.CompanyName)".Trim())'"
+          if ($fv -ne $ver4) { Write-Host "::error::installer FileVersion '$fv' is not '$ver4'"; exit 1 }
+          if ($pn -ne 'mStream') { Write-Host "::error::installer ProductName '$pn' is not 'mStream'"; exit 1 }
 
       - name: Upload bundle artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1271,11 +1271,48 @@ jobs:
           echo "$OUT" | grep -q IROH_OK || { echo "FAIL: iroh native binding did not load in the standalone binary"; exit 1; }
           echo "OK: iroh native binding loaded from the bundled .node"
 
+      # ── Windows installer (Inno Setup) ──────────────────────────────────
+      # Compiled LAST — after the VersionInfo assert and every smoke — so the
+      # setup.exe wraps exactly the bytes that were proven. ISCC ships on the
+      # windows-latest image. Per-user install to {localappdata}\Programs\
+      # mStream (same location install.ps1 uses); design notes live in
+      # build/windows-installer.iss. When SignPath lands, PE signing slots in
+      # before this step and setup.exe signing after it.
+      - name: Build Windows installer (win-x64)
+        if: matrix.key == 'win-x64'
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $iscc = "${env:ProgramFiles(x86)}\Inno Setup 6\ISCC.exe"
+          if (-not (Test-Path $iscc)) { Write-Host '::error::Inno Setup 6 not found on this runner'; exit 1 }
+          $ver = node -pe "require('./package.json').version"
+          $ver4 = node --input-type=module -e "import {windowsVersion} from './scripts/win-versioninfo.mjs'; import {readFileSync} from 'node:fs'; console.log(windowsVersion(JSON.parse(readFileSync('package.json','utf8')).version))"
+          $stage = (Get-Item dist/stage/mStream-*-win-x64).FullName
+          $out = Join-Path $PWD 'dist/installer'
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
**Local silent-install E2E on real Windows: 18/18**

Ran the CI-built artifact (`mStream-6.20.2-win-x64-setup.exe`, 144 MB, compiled in 21s with lzma2/fast) through the full battery on a real Windows 10 box:

```
ok: install exit 0                       (/VERYSILENT /DIR=<scratch>)
ok: installed: mStream.exe / mstream-server.exe / README.txt /
    webapp\index.html / bin\rust-parser\...exe / unins000.exe
ok: Start Menu shortcut
ok: uninstall registration (DisplayName=mStream, Version=6.20.2)
ok: installed launcher FileVersion=6.20.2.0
ok: installed launcher executes (--autostart=status exit 0)
ok: uninstall exit 0 -> files, shortcut, uninstall key all removed
ok: in-dir process was stopped by uninstall      <- WMI sweep scoped to the
                                                    install dir terminated a
                                                    launcher parked from it
ok: in-dir Run value was removed by uninstall    <- autostart guard, side A
ok: foreign Run value survived uninstall         <- autostart guard, side B:
                                                    a Run value pointing at
                                                    C:\SomewhereElse stayed
                                                    (the one-registration-
                                                    per-user footgun)
```

The live machine's real Run value was backed up before and restored byte-exact after (`restore exact: True`).

CI-side, the win-x64 leg now compiles the installer after the VersionInfo assert + all smokes and asserts the setup.exe's own VersionInfo (trim-aware — Inno space-pads the string fields). Three iteration fixes along the way, all Inno dialect: Pascal brace comments don't nest (`{app}` inside a comment), PascalScript has no `VarToStr` (plain Variant assignment coerces), and the padded VersionInfo strings.

IrosTheBeggar:
**Final verification round: 34/34 on the current artifact** (all three batteries re-run against the CI-built setup.exe from the latest commit)

**New since the first E2E:**

1. **Upgrade over a RUNNING install — 10/10** (the previously untested path):
```
ok: installed launcher booted the installed server (200 on :3998 after ~1s)
ok: both processes live from the install dir (mStream.exe + mstream-server.exe)
ok: install 2 (over running pair) exit 0
ok: old pair terminated by the pre-install sweep
ok: sweep log: launcher terminated by path match
ok: no RestartReplace fallback (no locked files at copy time)
ok: upgraded install boots and serves (200 on :3998)
```
This also closes the "installed copy never observed serving" gap: the installed launcher spawned its sibling server GUI-face, before and after the upgrade.

2. **Electron-collision guard — 6/6, new commit.** Found live on the dev machine: the legacy Electron-era desktop app (mStream Express 5.x) used electron-builder's default dir — the *exact* `%LOCALAPPDATA%\Programs\mStream` path this installer defaults to. Anyone upgrading from the old desktop app would get a silent tree-mix. `PrepareToInstall` now refuses (clear message, silent installs exit 7) when Chromium markers are present without our `mstream-server.exe`; a prior copy of THIS app upgrades normally (regression-tested).

3. **Install/uninstall battery re-run on the final bytes — 18/18** (unchanged from the first round; both autostart-guard directions, scoped sweep, byte-exact Run-value restore).

Remaining untested surface, for the record: the interactive wizard pages (silent mode covers the engine), and the tag-time release-attach line (first exercised at the next tag; fails loudly if the glob misses).

IrosTheBeggar:
**New: the guard now offers to uninstall the legacy Electron app and continue** (follow-up to the collision finding).

Behavior:
- **Interactive**: at Install time, a Yes/No prompt (default **No**): "This folder contains the old mStream desktop app (Electron). Setup can uninstall it now using its own uninstaller, then continue installing the new mStream. Your music files are not affected." Yes -> runs the old app's registered uninstaller, waits for completion, continues into the normal install. No -> the previous refusal page (Back still live for picking another folder).
- **Silent**: still refuses by default (Setup never removes another product silently); the refusal log now names the explicit opt-in: `/REMOVEOLDAPP=1`.

Mechanics, matched to the REAL legacy registration (verified live on a machine that has it): electron-builder registered with an **empty InstallLocation**, so the registration is matched to the target folder via the **uninstaller exe's quoted path** inside `QuietUninstallString`/`UninstallString`. NSIS uninstallers self-copy to TEMP and return early, so completion is detected by the Chromium marker disappearing (60s budget), not by exit code. All processes running from the old app's dir are stopped first (path-scoped enumeration). Failure shapes (declined / no registration / launch failure / still present after a minute) each abort with specific instructions.

Verification on the CI-built artifact:
- **Silent paths, 11/11**: refuse-without-switch (exit 7, switch hint in log, dir untouched) -> remove-and-proceed with `/REMOVEOLDAPP=1` (registration matched by exe path in the log, marker-poll caught the *delayed* deletion from a fake NSIS-style uninstaller that hands off and returns early, new app installed, clean uninstall after) -> markers-without-registration (exit 7, specific message).
- **Interactive**: prompt appears at Install with No focused; **No** -> refusal page (verified click-through); the Yes path's machinery is byte-identical to the `/REMOVEOLDAPP=1` path (the prompt only sets the same flag).
- Fixture modeled on the real registration (empty InstallLocation, quoted exe with spaces, `/currentuser /S`); the real legacy app on the dev machine was never touched (registration re-verified intact after the tests).

PR #846, opened 2026-08-19
feat(packages): Linux deb + rpm from the glibc bundles (P3.4b — native installers)

IrosTheBeggar:
Second leg of P3.4, stacked on #845 (GitHub retargets to master when it merges; only the packages commit is this PR's diff).

## What ships
Four packages per release, built with **pinned nfpm** (sha256-verified, like every fetched build tool here) from the exact zips the build job produced:
- `mstream_<ver>_amd64.deb` / `mstream-<ver>-1.x86_64.rpm` — **full desktop package**: `/opt/mstream` tree, `/usr/bin/mstream-server` + `/usr/bin/mstream-desktop` symlinks, system `.desktop` entry + 512×512 hicolor icon, desktop-db/icon-cache post scripts.
- `mstream_<ver>_arm64.deb` / `mstream-<ver>-1.aarch64.rpm` — **server-only**, matching their bundle (linux-arm64 ships no launcher by design).

Dependency posture: hard deps stay minimal (`libc6`/`glibc` — the server is pure glibc per the bare-Debian-11 audit); GTK3/appindicator/ALSA ride **Recommends**, so a headless `--no-install-recommends` install never drags a desktop stack in. musl targets are deliberately excluded (Alpine uses apk).

`manifest.json` stays zips-only (its consumers eat zips; the generator's `mStream-*.zip` glob never sees the packages). The release job gains `needs: linux-packages` — without it a tag's attach step races the packaging job — and uploads the deb/rpm set next to the zips.

## CI proof on every run
- amd64 deb: installed on the runner, `desktop-file-validate`, **booted through the `/usr/bin` symlink to a 200 ping**, clean `apt remove`.
- x86_64 rpm: installed + exec'd in a `fedora:41` container.
- arm64 packages: content asserts (server present; deb must NOT ship a launcher).

## Local E2E already done (real v6.20.2 zips, ubuntu:24.04 + fedora:41 containers)
- All 4 packages build; amd64 deb installs, validates, **boots as a non-root user and serves 200 in ~2s**, removes cleanly (`/opt/mstream` gone); fedora rpm installs/execs/removes cleanly; arm64 content asserts pass.
- Trap fixes proven along the way, baked into the workflow: `dpkg-deb -c | grep -q` under pipefail SIGPIPE-fails **because** it matches (list to a file first); desktop-entry Categories with two main categories (now `AudioVideo;Audio;Player;`).

## Flagged for follow-up (not this PR)
Running the packaged server **as root** parks runtime data inside `/opt/mstream` (the dataRoot ladder treats any writable appRoot as portable) — leftovers then survive an `apt remove`, which is dpkg-correct but untidy. Non-root runs (the normal case, and what CI smokes) use `~/.local/share/mstream` properly. A task chip proposes teaching boot-config to skip the appRoot rung under system prefixes (`/opt`, `/usr`).

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1318,9 +1318,228 @@ jobs:
           path: dist/installer/mStream-*-win-x64-setup.exe
           if-no-files-found: error
 
+  # ── Linux native packages (P3.4b) ─────────────────────────────────────────
+  # deb + rpm for the glibc bundles, built with pinned nfpm from the exact
+  # zips the build job produced (musl targets use apk-based distros — no
+  # deb/rpm for them). amd64 is the full desktop package: /opt/mstream tree,
+  # /usr/bin symlinks for both faces, a system .desktop entry + hicolor icon.
+  # arm64 bundles are server-only BY DESIGN (no launcher), so their packages
+  # are too. Server hard-deps stay minimal (the server is pure glibc — the
+  # bare-Debian-11 dep audit); the desktop/audio libraries ride Recommends,
+  # so a headless `--no-install-recommends` install never drags GTK in.
+  # The amd64 deb is installed and BOOTED on this runner; the rpm proves
+  # itself in a Fedora container; arm64 can't exec here, so its packages get
+  # content asserts.
+  linux-packages:
+    name: Linux deb/rpm packages
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Download linux-x64 bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: mStream-linux-x64
+          path: zips
+      - name: Download linux-arm64 bundle
+        uses: actions/download-artifact@v4
+        with:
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
**Two more proofs on the built deb** (ubuntu:24.04 container, real v6.20.2 bundle bytes):

- `apt-get install --no-install-recommends ./mstream_6.20.2_amd64.deb`: **libgtk-3-0 confirmed NOT installed**, and the headless server boots and serves 200 through the /usr/bin symlink — the minimal-hard-deps/Recommends posture works as designed for headless boxes.
- **`gtk-launch mstream`** (the installed menu entry, as a non-root user under Xvfb): Exec/TryExec resolves, the launcher runs from /opt, and it boots the server to 200 on :3000 — the desktop-user click path end-to-end, short of a full DE.

Also ran `scripts/release-manifest.sh` locally against a directory containing all 7 zips **plus** the setup.exe and the four deb/rpm packages: exactly the 7 zips in the manifest, no refusal — the tag-time generator provably never sees the new asset types.

PR #850, opened 2026-08-19
p2p-sidecar: stop committing binaries — release assets + committed sha256 manifests, fetch on first use

IrosTheBeggar:
Stops committing p2p-sidecar binaries to git. The nine iroh-linked binaries (172 MB/set) become **GitHub release assets under immutable content-addressed names**, and the tree keeps only a small **committed manifest** the server verifies against when it **fetches its one platform binary on first discovery-network use**. Full design doc: `docs/p2p-sidecar-distribution.md` (in this PR) — phase 1 of the plan to stop regrowing binary history before the separate filter-repo rewrite.

## Why now, with numbers

- Each CI refresh of `bin/p2p-sidecar` added **~172 MB of undeltifiable blobs to history forever**; measured today: **1,118 MB of unique sidecar blobs already reachable** from 15 refresh commits (24 refresh events across the two workflows). rust-parser is at ~1.6 GB the same way (phase 2, outlined below).
- The npm tarball had **no `.npmignore` entry for the sidecar**: every `npm i mstream` shipped **179 MB unpacked (63% of 284 MB) of binaries for all nine platforms**, for a feature that's **off by default**. After this PR: **125 MB → 47 MB compressed, 284 MB → 100 MB unpacked**, and only enabled servers download only their own ~20 MB, sha256-verified.

## The model (decisions made — please veto anything here)

1. **Where assets live**: one rolling machine-managed **prerelease** `p2p-sidecar-assets` (prerelease ⇒ can never become `releases/latest`, which install.sh resolves). **Append-only**: asset names are `<key>-<tree12>-<sha8>[.exe]` — same name ⟹ same bytes — so publish *skips* existing names and *never replaces*, and a manifest committed at any point in history keeps resolving. Rejected: replaced-in-place assets on a rolling tag (breaks every older checkout's pins), per-run version tags (tag sprawl), workflow artifacts (expire), LFS (quotas, forks).
2. **Committed manifests as the trust anchor**: `bin/p2p-sidecar/manifest.json` + `manifest-musl.json` (two files so the two workflows never write the same path — the `.source-tree`/`.source-tree-musl` split carried forward). Each pins `{file, sha256, size, url}` per platform plus `sourceTree`/`builtFrom`/`workflowRun`. This **replaces and strengthens the ledger-only stamps** (nothing ever asserted them; now the pins are verified at every fetch). build-bun's tag asserts still deliberately do not cover p2p (not staged into bundles — the staging comment in build-bun.mjs is updated to say why).
3. **Fetch timing**: inside `discovery-p2p.start()` — i.e. at boot when `discoveryP2p.enabled`, or during the admin enable request; both funnel through the same stack start that already has graceful-degradation + flag-rollback error paths. Because start() can now spend seconds in a download before `proc` exists, a `stop()` landing in that window used to be able to no-op and let the child spawn afterwards into a stack that believes it's stopped (the #839/B1 bug class) — start chains now carry a **generation gate** (`startGen`): stop() bumps it and detaches the in-flight chain, and a chain re-checks its generation right before spawning, aborting instead of orphaning. The successor start joins the same in-flight download via the bootstrap's single-flight. *This gate is the one piece verified by reasoning + the existing stack-race suite rather than a dedicated new test — please eyeball it (~15 lines in `discovery-p2p.js`).* Resolution order unchanged in spirit: **dev cargo build → operator-placed binary in `bin/p2p-sidecar/` → managed (fetched) install** in `dataRoot/bin/p2p-sidecar` (= the same dir for a plain checkout; a writable one when appRoot isn't — the ffmpeg precedent). A `.fetched.json` receipt records what the fetcher installed: receipted installs auto-refresh when the manifest moves; anything else is operator property, never touched.
4. **Offline/airgap/Docker**: `npm run fetch-p2p-sidecar` pre-fetch CLI (bake into images), `MSTREAM_SIDECAR_BASE` mirror override (pins still enforced; https-or-loopback only — the `MSTREAM_FFMPEG_MIRROR`/`MSTREAM_RELEASE_BASE` precedent), or hand-place the binary.
5. **Migration**: the 9 binaries + 2 stamps leave the tip (history cleanup stays with the planned filter-repo pass); `.gitignore`/`.npmignore` get sidecar sections (manifests + README negated in).

## Workflow changes

Both p2p workflows keep their build/self-test/real-run-smoke legs untouched; the commit job becomes a **publish job**: inventory assert → build manifest + immutable-named copies (`scripts/p2p-sidecar-manifest.mjs`) → ensure release exists (create race handled) → upload only missing names → commit the **manifest** through the exact #840/#844 no-rebase / stand-down / retry-from-tip loop. **Assets upload before the manifest commits**, so a committed manifest can never point at a missing asset (a stand-down after upload leaves unreferenced immutable assets — harmless).

Also in this PR:
- musl workflow finally gets the **pinned+sha256'd `cross` install and `--locked`** the glibc workflow already had (it was the last family floating `releases/latest` — review finding C5's remainder).
- `npm-publish.yml` + `deploy-demo.yml` get tag-prefix guards (`p2p-sidecar-*`, `discovery-models-*`). The workflows' own release creation can't trigger them (GITHUB_TOKEN events never do — relied on by every bot commit already), but a **human** editing an asset-store release in the UI could re-fire `release: published`; today that would `npm publish` + redeploy the demo off `discovery-models-1` too. Post-6.20.3-safe: nothing here needs to land before the release, and the guards only *narrow* the triggers.

## First publish after merge (deliberate, self-healing gap)

This PR ships the manifests as **empty stubs** — it cannot upload assets itself (no release mutations pre-merge, by agreement). On merge, the workflow-file changes **self-trigger both workflows on master**: they create the release, upload all nine assets, and commit real manifests within one CI cycle (~20 min). In that window a fresh checkout with the feature enabled logs `no prebuilt binary is published for this platform` and degrades exactly like today's missing-binary path; existing checkouts keep their on-disk binaries (untracked → deleted only on pull, refetched on next start).

**Optional zero-gap alternative**: dispatch both workflows on *this branch* before merging (`workflow_dispatch`, dry_run=false). The branch versions run, create the real release, upload assets, and push real manifests **onto this PR branch** — the merge then lands with populated manifests and no gap. Left for you to pull the trigger since it mutates the repo's releases.

## Verification (all local — no releases touched)

1. **Unit tests** — `test/unit/p2p-sidecar-fetch.test.mjs`, 11 tests mirroring the ffmpeg-bootstrap conventions (loopback server, hermetic): happy install + receipt; checksum-mismatch refused with staging cleaned; oversize body aborted mid-flight; failed execution probe refused; empty/malformed manifest ⇒ unfetchable without network I/O; operator binary never re-fetched; receipt-gated refresh on manifest change; single-flight concurrency; mirror override with pins still enforced; cleartext non-loopback URL refused. Full unit dir: **475 pass / 0 fail**.
2. **Git-sandbox simulation** of both publish-step bodies, extracted **verbatim** from the YAML (short-path sandbox, file:// remotes, PATH-shimmed `gh` modeling create-once/append-only/upload-fails-on-duplicate, PATH-shimmed `git` for the race): **35/35 assertions** across 7 cases — glibc happy (pins byte-match staging, tree/urls/sizes correct), musl publish reusing the release without clobbering the glibc manifest, identical re-run (uploads skipped, manifest no-op), stand-down on newer source at tip (manifest untouched; pre-uploaded assets orphaned harmlessly), sibling push inside the fetch→push window (reject → retry from new tip, both commits preserved), partial staging refused before any `gh` call, and the release-create race (`|| true` + re-view).
3. **Docker end-to-end**, real binaries, fresh `git archive` checkouts, `npm ci`, loopback asset store via `MSTREAM_SIDECAR_BASE`, CI-shaped manifests generated by the real `scripts/p2p-sidecar-manifest.mjs`:
   - **debian (glibc)**: boot with `discoveryP2p.enabled` → `downloading p2p-sidecar-linux-x64-… (21.2 MB)` → `checksum verified — installed` → **real sidecar `ready — endpointId=…` through the server's own spawn path** → admin status `running:true`, receipt written.
   - **alpine (musl)**: same, with the musl binary (the linuxserver.io story).
   - **debian + alpine tampered**: same-size corrupted assets → `checksum mismatch … refusing the download` → `catalog unavailable — feature disabled this boot` → nothing installed, server stays healthy (`/ping` 200, status `running:false`).
   - Seed-guards from the test-helper template were replicated (dead seedListUrl, `useCommunitySeeds:false`, `MSTREAM_TEST_BAKED_SEEDS='[]'`) so the E2E never touched the real discovery network.
4. **eslint** on every touched file: clean (the one `require-await` warning in ffmpeg-bootstrap predates this PR — verified against master's copy).
5. **Full `npm test`**: results in the first comment below (run after the E2E to keep the known load-flakes out of play).

## Phase 2/3 outline (not implemented — rust-parser, rust-server-audio)

Same store pattern (per-family rolling prerelease + committed manifests), but both families **are staged into release bundles and tag-asserted**, so their PRs touch the release path: `build-bun.mjs` fetches manifest-pinned assets at bundle time instead of copying `bin/` (VersionInfo stamping unaffected — it already stamps the staged copy); build-bun's tag asserts read the manifests' `sourceTree` instead of `.source-tree` stamps *plus* gain a per-binary hash check the stamps never had; runtime fallbacks (JS scanner, CLI players) already provide the degradation net. Sequence: let p2p-sidecar publish cycles prove the store, then rust-parser (the ~1.6 GB whale, 92 refreshes), then rust-server-audio. At that point the publish step body should become a shared composite action (rule of three).

## Needs your sign-off specifically

- The rolling **prerelease + append-only immutable assets** shape (vs. versioned releases per source change).
- The `npm-publish.yml`/`deploy-demo.yml` guard edits (post-release-safe, but they touch release-triggered workflows).
- Empty-stub manifests + self-triggered first publish (or the branch-dispatch zero-gap option above).
- Asset-store hygiene rule going forward: **never delete or replace assets** on `p2p-sidecar-assets`.


IrosTheBeggar:
## Full `npm test` results (local, Windows)

`npm test` on the branch: **2311 tests / 510 suites — 2304 pass, 0 fail, 6 skipped, 1 cancelled** (516s).

The one cancellation was `test/subsonic/subsonic-ui.test.mjs` — its `before()` boot hit the 90s `server not ready` ceiling. That's the known load-flake shape on this machine, and the timing lines up exactly: that suite ran while three Docker `npm ci` containers (the E2E legs) were saturating the box. **Solo rerun: 16/16 pass.** The file touches nothing in this diff.

Because the full run started a few minutes before the last commit (the `startGen` stop-race gate, which touches only `src/state/discovery-p2p.js`), the suites that exercise that file were rerun on the final commit: `test/integration/discovery-p2p.test.mjs` + `test/unit/discovery-p2p-stack-race.test.mjs` — **48/48 pass**, including the real-sidecar blob/gossip loops and enable/disable/respawn cycles through the restructured `start()` (the sidecar exe is still on disk untracked in my worktree, so the binary-dependent layers ran rather than skipped).

One CI-coverage note for after the merge: on fresh CI checkouts those binary-dependent integration layers will skip (their existing `describe.skip`-when-no-binary design) until either a fetch step lands in `test.yml` or a local build is present. Worth a small follow-up once the first publish run has populated the manifests — the test could then fetch via the committed pins like production does.

IrosTheBeggar:
Closing as superseded under the revised (approved) plan — the rolling `p2p-sidecar-assets` prerelease design this PR implemented is dead; nothing was ever published under that tag, and no successor references it.

What replaced it:

- **[IrosTheBeggar/mstream-p2p-sidecar](https://github.com/IrosTheBeggar/mstream-p2p-sidecar)** — the crate extracted to its own repo; normal v-tagged releases carry the binaries (CI attaches them + manifest fragments to a **draft**; humans publish). v1.0.0's draft is already built and waiting.
- **#852** — the split-out keepers that merge regardless: npm-publish/deploy-demo tag guards + the musl workflow's `cross` pin/`--locked`.
- **#854** — the mStream side: crate/workflows/binaries leave the tip; committed manifests pin `{repo, tag, file, sha256}` (v1.0.0); fetch-on-first-use, receipt semantics, `startGen` gate, pre-fetch CLI, volume acceptance smoke, and the updated design doc all carried forward from here and re-verified.

The branch (`claude/p2p-sidecar-release-assets`) can be deleted whenever.

PR #854, opened 2026-08-19
p2p-sidecar: own repo + versioned releases; mStream pins {repo, tag, sha256} and fetches on first use

IrosTheBeggar:
Supersedes #850 under the revised plan: the sidecar now lives in its own repo — **[IrosTheBeggar/mstream-p2p-sidecar](https://github.com/IrosTheBeggar/mstream-p2p-sidecar)** — with normal v-tagged releases carrying the binaries, and mStream keeps only **committed text manifests pinning `{repo, tag}` + `{file, sha256, size}`** per platform, fetching on first discovery-network use. No rolling prerelease in this repo, no binaries in git anywhere. Design doc: `docs/p2p-sidecar-distribution.md` (in this PR).

## What already happened in the sidecar repo (created per the approved plan)

- Crate extracted from this repo at `61eca273` (path `p2p-sidecar/`, tree `8080df40efa8` — recorded in the extraction commit + README). The Phase A size branch wasn't on the remote at extraction time, so v1.0.0 was cut from the pre-optimization profile; **when #853 landed, its size-tuned profile was ported verbatim and tagged `v1.0.1`** (win 20.5→6.7 MB, linux 21.2→8.1 MB) — the version this PR pins. v1.0.0's draft is superseded without ever being published (fix-forward; Paul can delete that draft+tag or just ignore it).
- Version policy: `MAJOR.MINOR` mirrors the pinned iroh line (iroh 1.0.3 → **v1.0.x**), PATCH is the sidecar's own. README documents the release flow, the immutability rule for published assets, the *unsigned-assets* stance (bundle-time signing in mStream covers the SAC-critical path), and the capability-probe compatibility contract (probe new RPCs; `unknown command` ⇒ degrade — the exact pattern the N3 integration suite already uses).
- Workflows ported and reshaped: CI gate (PR/main: `cargo test --locked` + build + identity self-test) plus two tag-driven release workflows (glibc/darwin/win + musl) that keep everything Paul asked to keep — pinned+sha256'd `cross`, `--locked`, per-leg self-tests, the real-run socket smoke, qemu musl self-tests — and attach the nine binaries plus two `manifest-fragment*.json` files to a **DRAFT release**. Draft→publish is the documented choice: CI never publishes; fragments upload LAST as each family's completion marker; a published release is refused by re-runs (immutable — fix forward with a new tag).
- **Two full release cycles have run green**: v1.0.0 (first cycle, all three workflows green first try) and **v1.0.1** (the #853-profile port) — each a draft fully populated by CI (9 binaries + 2 fragments), left unpublished for Paul.

## This PR (the mStream side)

- **Deleted from the tip**: the crate (`p2p-sidecar/`), its two build workflows, the 9 binaries + 2 ledger stamps (172 MB). History cleanup remains the separate filter-repo project.
- **Committed pins**: `bin/p2p-sidecar/manifest{,-musl}.json` pin **v1.0.1's real fragments** (generated by the new ritual helper against the draft's assets). They go live the moment the draft is published; until then a fetch 404s with a clear log line and the feature degrades exactly like the missing-binary case — merge order with the publish doesn't matter.
- **Fetch on first use** (unchanged in spirit from #850, reworked pins): URL *derived* from token-validated `{repo, tag, file}` (never stored); sha256 + in-flight size cap + `--print-id` execution probe; receipt-gated refresh; operator binaries and nested-clone dev builds always win; `startGen` stop-race gate; single-flight; `MSTREAM_SIDECAR_BASE` mirror override (https-or-loopback only).
- **The release ritual**: `node scripts/update-p2p-sidecar-manifest.mjs <tag>` regenerates both manifests from a *published* release's fragments → small text PR. Documented in the design doc; the helper honors the mirror override so it's smoke-testable (and air-gap-usable).
- **Docker/image guidance** (docs + `bin/p2p-sidecar/README.md`): bake with `RUN npm run fetch-p2p-sidecar` for zero runtime egress (the linuxserver.io-class recommendation), or run with a persistent volume over the data root.
- **New acceptance smoke** (`test/smoke/docker/p2p-sidecar-volume-smoke.sh`, `npm run test:smoke:sidecar-volume`): volume-over-dataRoot semantics — one fetch onto the volume, container recreate ⇒ **zero** re-fetches, manifest bump ⇒ **exactly one** refresh.
- **Test-helper guard**: dead-loopback `MSTREAM_SIDECAR_BASE` in `test/helpers/server.mjs` (same family as the dead `seedListUrl`) so suites on binary-less checkouts can never fetch from the real release mid-test.
- Includes #852's commit (guards + musl `cross` pin) so this branch stands alone; whichever merges first, the other's diff shrinks to nothing/deletions.
- Dev loop note: `p2p-sidecar/` is gitignored (**root-anchored** — an unanchored pattern also swallows `bin/p2p-sidecar/`'s manifests; caught by a post-commit `ls-tree` check and now a comment in both ignore files) so `git clone <sidecar-repo> p2p-sidecar && cargo build --release` restores the old build-and-run-in-place flow.

## Verification

1. **Unit**: `test/unit/p2p-sidecar-fetch.test.mjs` reworked to schema 2 — 13 tests (adds: non-token repo/tag/file pins are unfetchable and never interpolated into a URL; exact derived-URL shape via the exported `deriveAssetUrl`). Full unit dir: **477 pass / 0 fail**.
2. **Release-workflow simulation** (new repo's publish steps, extracted verbatim, `gh` shimmed with real semantics — create-fails-if-exists, upload-fails-on-duplicate-unless-clobber, draft vs published): **27/27** across 8 cases — happy draft+6+fragment (pins byte-match staging), musl joins the same draft without a second create, idempotent re-run (fragment present ⇒ zero uploads), partial staging refused before any `gh` call, re-run after partial upload (draft+fragment-absent ⇒ clobber+complete), published+complete ⇒ graceful no-op, published+incomplete ⇒ hard error, create race (`|| true` + re-view).
3. **The real thing, twice**: v1.0.0 and v1.0.1 tag pushes → all workflows green both cycles; each draft populated with 9 binaries + 2 fragments; fragment sha256s **verified against the downloaded CI bytes** both times.
4. **Ritual smoke**: `update-p2p-sidecar-manifest.mjs` run against loopbacks serving the real fragments for **both** releases → committed manifests match the fragments exactly (the v1.0.1 output is what this PR pins) — the new-release ritual has been exercised end-to-end twice before its first production use.
5. **Docker E2E, 4/4 — run twice** (once per release; final run on the pinned **v1.0.1** bytes): debian glibc + alpine musl × good/tampered — fresh `git archive` checkouts, `npm ci`, loopback store, committed pins — download → `checksum verified` → real sidecar `ready — endpointId=…` through the server's spawn path → status `running:true` + receipt; tampered same-size bytes → refused, degraded, server healthy, nothing installed. Seed guards replicated so nothing touched the real network.
6. **Volume acceptance smoke: 12/12 — run twice** (v1.0.0 and v1.0.1 binaries) — phases A/B/C as described above, with dataRoot genuinely diverged from appRoot (checkout at `/opt/mstream`, `XDG_DATA_HOME` on the volume).
7. **eslint** on every touched file: clean (the one `require-await` warning in ffmpeg-bootstrap predates this work).
8. **Full `npm test`**: results in the first comment below.

## Paul's clicks (in any order except where noted)

1. **Publish the v1.0.1 draft** in mstream-p2p-sidecar (this PR's pins point at it). The v1.0.0 draft is superseded — delete its draft+tag or ignore it.
2. Merge #852 (tiny, independent), then this PR.
3. #850 is closed with a pointer; delete `claude/p2p-sidecar-release-assets` when convenient.
4. Phase D (bundle staging + signing) proceeds against the store/manifest shape exactly as it is in this PR — nothing in `build-bun.yml`'s staging was touched here.


IrosTheBeggar:
## Full `npm test` results (local, Windows, final branch head)

**2313 tests / 510 suites — 2307 pass, 0 fail, 0 cancelled, 6 skipped, exit 0.** No flakes this run (unlike the #850 run, nothing else was loading the machine).

Notably green with real meaning for this PR: the whole `discovery-p2p` integration file including the **N3 seeders + swarm** and **zero-touch auto-publish** suites — real sidecar processes spawned through the reworked `start()` (this machine still has a binary on disk, so the binary-dependent layers ran rather than skipped; on binary-less checkouts they skip exactly as before, and the new dead-loopback `MSTREAM_SIDECAR_BASE` guard in the test helper guarantees no test can ever reach the real release).

PR #851, opened 2026-08-19
feat(signing): Windows Authenticode via SignPath (P3.5 phase 2)

IrosTheBeggar:
Wires the win-x64 leg to SignPath (org `075a387f…`, project `mStream`). Design per the signing plan, updated for the installer's existence:

## Flow
1. **After the VersionInfo assert** (bytes final, metadata proven): the 6-PE set uploads as an artifact → **request #1** against `win-bundle-pes` (5 PEs restricted to `product-name=mStream` + 4-part `${version}`; the upstream iroh .node signed unrestricted) → signed files verified (`Get-AuthenticodeSignature` = Valid) and swapped into the stage → **zip recut with System32 bsdtar** (mirrors the bundler; Compress-Archive's backslash entries and GNU tar's tar-in-zip-clothing both dodged).
2. **Everything downstream runs on signed bytes** — the full smoke battery (boot, scan, reboots, bind change) plus ISCC, which compiles the installer **from signed PEs**.
3. **Request #2** signs the setup.exe against `windows-installer` (no product-version restriction yet — Inno space-pads VersionInfo strings; tighten after the first release read).

## Posture (mirrors the darwin signing step)
| Context | Behavior |
|---|---|
| Tag | `release-signing` — Foundation cert, **manual approval**, 2h wait budget (approve promptly; a paused Windows runner burns minutes) |
| `workflow_dispatch` | `test-signing` — auto-approved test cert; every dispatch co-proves this pipeline |
| PR / fork / no token | Skip, ship unsigned (quota + fork-secret safety) |
| **Tag with token missing** | **Fail loud** — never silently ship an unsigned release |

## Repo-versioned artifact configs
`build/signpath/win-bundle-pes.xml` + `windows-installer.xml` — lockstep copies of the UI configs, with the sidecar `pe-file` line documented as arriving together with bundle staging (Phase D of the sidecar plan).

## Verification
- YAML + structural asserts (step ordering: sign → smoke → ISCC → sign installer → uploads; both requests wired to the org/project/config slugs).
- This PR's own CI exercises the **skip path** (PR context, no signing).
- The **test-signing dispatch proof** runs the moment the remaining Phase-3 clicks exist: project `mStream`, both artifact configs pasted, both policies, GitHub App on the repo, CI-user token as `SIGNPATH_API_TOKEN`. I'll fire and verify it then.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #855, opened 2026-08-19
feat(bundle): bake the p2p-sidecar into every release bundle (Phase D)

IrosTheBeggar:
The plan's R1: bundles ship the sidecar built in, so discovery P2P works on binary installs — with the Windows copy positioned to be Authenticode-signed alongside the PE set.

## How it works
- `build-bun.mjs` stages the target's sidecar from the **pinned release assets** (the same committed manifests the runtime fetch uses; URL derivation imported from `p2p-sidecar-bootstrap` — one implementation, no drift): fetch → **sha256+size verify** → stage → **VersionInfo stamp** on Windows (`mStream P2P Sidecar`, full product contract). Cache under `dist/sidecar-cache/` keyed by pin.
- The runtime resolver's existing `appRoot/bin/p2p-sidecar/<name>` rung picks it up — **bundle installs never fetch**; npm/source/Docker behavior unchanged.
- musl bundles stage the `-musl` assets from `manifest-musl.json`.
- **Posture** (launcher precedent): CI missing-asset = loud failure (`MSTREAM_ALLOW_MISSING_SIDECAR=1` = deliberate); local/offline = warn + ship without (runtime fetch fallback); **hash mismatch = fatal everywhere, never skippable, never cached**. `MSTREAM_SIDECAR_BASE` mirrors honored, pins still enforced.
- `check-win-versioninfo.ps1` $known gains the sidecar as a **stamped, full-contract** entry (optional-when-absent covers deliberate/offline builds; CI absence is already fatal upstream).
- **Smoke (1b)** on every smoking leg: the staged binary must exist AND execute (`--print-id`) — on darwin that exercises the **signed** copy (the codesign walk covers it as a generic Mach-O automatically; no sign-step changes needed).

## Verified locally (win-x64, real assets)
Fetched v1.0.1 from the actual public release URL → sha verified (6.7 MB) → staged → stamped → **stamped PE executes** (`--print-id`) → zip carries it → **strict VersionInfo contract 5/5** including the new row.

## Coordination
- The SignPath branch (#851) gets the matching update (6th `pe-file` in `win-bundle-pes`, **with** product restrictions now that it's stamped, + the collect/apply steps) — pushed separately; the UI config to paste is the 6-entry version.
- Bundle size cost: ~+3 MB compressed per bundle (the #853-optimized builds).

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -959,6 +959,19 @@ jobs:
           echo "$OUT" | grep -q dbPath || { echo "FAIL: worker self-dispatch"; echo "$OUT"; exit 1; }
           echo "ok: worker self-dispatch"
 
+          # (1b) the baked-in p2p-sidecar: staged from the pinned release
+          # assets (sha-verified at bundle time) and actually EXECUTABLE on
+          # this leg — --print-id is the sidecar's own identity self-test.
+          # Bundle installs' discovery P2P depends on exactly this file (the
+          # resolver's appRoot rung); on darwin this also proves the SIGNED
+          # copy still runs (the sign walk covers it as a generic Mach-O).
+          SC=$(ls "$RUN"/bin/p2p-sidecar/p2p-sidecar-* 2>/dev/null | head -1)
+          [ -n "$SC" ] || { echo "FAIL: no p2p-sidecar staged in the bundle (bin/p2p-sidecar empty)"; exit 1; }
+          SCOUT=$("$SC" --print-id --data-dir "$RUN/sc-selftest" 2>/dev/null)
+          echo "$SCOUT" | grep -Eq '^[0-9a-f]{64}$' || { echo "FAIL: staged p2p-sidecar --print-id produced no/invalid output"; echo "$SCOUT"; exit 1; }
+          rm -rf "$RUN/sc-selftest"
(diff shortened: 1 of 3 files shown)


IrosTheBeggar:
**End-to-end boot smoke (local win-x64, staged bundle): PASS** — the user-facing claim proven directly.

Booted the staged bundle's `mstream-server.exe` with `discoveryP2p.enabled: true`, an isolated data root (`LOCALAPPDATA` redirect), and `MSTREAM_SIDECAR_BASE` pointed at a **dead loopback port** so any fetch attempt would fail loudly:

- **(a) zero-fetch boot**: `[p2p-sidecar] ready` logged with the dead fetch base — the stack came up without touching the network.
- **(b) it's the baked-in copy**: the live sidecar child's `ExecutablePath` (via WMI, parent-pid-scoped) = `<stage>/bin/p2p-sidecar/p2p-sidecar-win32-x64.exe` — the resolver's appRoot rung, exactly as designed.
- **(c) no managed install created**: the isolated data root contains no fetched binary and no `.fetched.json` receipt.

Together with the earlier proofs (real-release fetch + sha verify at bundle time, stamped PE executes, zip carries it, strict VersionInfo 5/5), Phase D is smoke-complete locally; CI is the cross-platform wave.

PR #856, opened 2026-08-19
feat(ffmpeg): pin downloads to tagged releases via a committed manifest

IrosTheBeggar:
Retires the rolling-`latest` ffmpeg download in favor of the p2p-sidecar trust model: **committed, reviewed pins; fail-closed verification; updates as manifest-bump PRs**.

## Why
- Every weekly rolling refresh handed Windows users a brand-new **unsigned** binary hash with zero SmartScreen / **Smart App Control** cloud reputation — and SAC has no per-app override. A stable pinned hash accrues reputation instead of resetting weekly.
- Supply chain: we shipped whatever BtbN published that week (trusting their live checksum file at fetch time). Now the digest is committed text, reviewed like any dependency bump.

## What
- **`bin/ffmpeg/manifest.json`** (committed; gitignore negation): dated BtbN autobuild tag + per-platform `{file, sha256, size}` for linux-x64/arm64 + win32-x64; versioned martin-riedl paths + digests for both mac archs. URLs are **derived from validated pins** (token/segment checks — no traversal, no scheme smuggling; the mac host is a code constant, not manifest data).
- **Fail closed**: size cap during transfer, exact-size + sha256 after — any miss deletes the file and refuses. No fetch-the-checksum fallback, no silent downgrade to `latest`. (System-PATH rung unchanged as the graceful degrade.)
- **Local update check**: `.fetched.json` receipt records the installed pins; `checkForUpdate()` is now a receipt-vs-manifest compare — **zero network** unless a refresh is due. New builds arrive by updating the manifest (`node scripts/update-ffmpeg-manifest.mjs [tag]` — refuses the rolling tag).
- **Provenance preserved**: no marker in a custom `ffmpegDirectory` = operator's binaries, hands off (legacy `.checksum` still honored as the marker); default-dir installs converge onto the pins once, receipt replaces `.checksum`. `transcode.autoUpdate=false` still freezes everything.
- Preserved wholesale: single-flight per dir, staging + atomic pair swap, execution probes, `MIN_FFMPEG_MAJOR`, `MSTREAM_FFMPEG_MIRROR` (flat namespace; pins enforced regardless of source).

## Verified
- **12/12 unit**: new pure `pinnedRelease()` suite (URL derivation, malformed-entry + traversal refusals, committed-manifest self-check across all 5 platforms) + the existing single-flight suite, which now doubles as the mismatch-refuses-install proof (fake mirror bytes vs real pins).
- **Real-network E2E (win32-x64)**: pinned 170 MB archive fetched from the actual BtbN release URL → sha+size verified → installed → receipt pins correct (18 s).
- **No-op locality**: dead mirror + matching receipt → `checkForUpdate()` silent, zero fetch attempts.
- **Legacy migration**: dead mirror + `.checksum`-era install → refresh triggered, failed closed (`ECONNREFUSED`), live binaries untouched.
- Lint: zero added problems (the one warning pre-exists on master).

## Operational notes
- BtbN retains dated releases in bulk but not forever — refresh the manifest a few times a year so **fresh installs** never chase a pruned tag (installed servers never re-fetch, so they're unaffected). The helper prints a review-ready diff summary.
- Existing installs converge onto the pinned build on their first post-update check (one re-download), then go quiet.


IrosTheBeggar:
**Added: monthly auto-refresh** (`.github/workflows/update-ffmpeg-manifest.yml`, per Paul's request).

- Schedule (3rd of month) + `workflow_dispatch` (optional tag input).
- Re-runs the helper, then `--verify` **downloads all 7 pinned assets (~500 MB) and hashes them against the pins** — that run is the CI for the auto-PR, since GITHUB_TOKEN-created PRs trigger no workflows. Any miss fails before a PR exists.
- Pins moved → opens/force-updates `auto/ffmpeg-manifest-update` with a link to the generating run. Pins current → quiet exit. **A human merges** — pins stay reviewed.
- Helper gained 3-attempt backoff retries; a real transient 404 from martin-riedl's edge hit during local testing and was absorbed on retry, so this is proven, not speculative.

Local proof: full `--verify` green (7/7 assets sha-verified, ~533 MB), stand-down path exercised ("no changes — pins already current").

IrosTheBeggar:
**Linux E2E (Docker, node:24 glibc): PASS** — closes the one genuine correctness gap.

`extractTarXz` derives the in-archive prefix from the asset file name, and dated BtbN assets have *versioned* names the old rolling-`latest` code never used — untested until now (the win-x64 E2E couldn't prove it: zip extraction is name-based). Fresh container install: pinned 128 MB `ffmpeg-N-126217-…-linux64-gpl.tar.xz` fetched from the real release URL → sha verified against the pin → **extracted (prefix matched)** → both binaries execute (`ffmpeg/ffprobe version N-126217-ge1e325235e-20260819`) → receipt correct. 22 s.

**Smoke ledger for this PR:**
| Path | Proof |
|---|---|
| win32-x64 install | Real-network E2E (170 MB, sha, receipt, 18 s) |
| linux-x64 install + tar extraction | Docker E2E above |
| darwin install | *Accepted low residual*: shares `downloadPinned` with the two proven paths; `extractZipUnix`/swap unchanged code; pinned mac URLs downloaded+hashed 3× by `--verify` (not extracted/executed on darwin — no mac in this loop) |
| checkForUpdate locality / legacy migration / fail-closed | Dead-mirror phases (earlier comment) |
| Pin validation | 12/12 unit incl. committed-manifest self-check |
| Helper `--verify` | 7/7 assets ×2 runs; transient-404 retry exercised live |
| Monthly workflow bash | **Pre-merge dispatch impossible** (GitHub 404s workflows not on the default branch) — post-merge rehearsal planned: dispatch with the older `autobuild-2026-08-18-15-03` tag → verify the auto-PR opens correctly → close it; then a no-tag dispatch → expect the quiet "pins already current" exit |

Note: CI's bundle boot smokes never download ffmpeg (leg times prove it), so CI green here does **not** cover these paths — the ledger above is the coverage.

PR #858, opened 2026-08-19
build: measured compression for the installer and linux packages

IrosTheBeggar:
Paul flagged the Windows installer's size. Root cause: the compression settings were speed presets whose cost was asserted, never measured. This PR sets them from measurements.

## Windows installer: `lzma2/fast` → `lzma2/ultra64` + 2 block threads
- The old comment claimed max "squeezes a few MB" — a same-payload proxy (7z lzma2 over the real win-x64 stage, 162 MB dominated by the 119 MB Bun server exe) measures **51 MB (fast) vs 38 MB (max-class): −25%**. Expected setup.exe: **56 MB → low-to-mid 40s** (smaller than the darwin pkgs).
- `LZMANumBlockThreads=2` halves the compile wall time for a 1–2% ratio give-back and bounds memory (~1.4 GB) on the 4-core runner. ISCC runs on every win-x64 leg, so **this PR's own CI shows the real size and time cost** — if the leg time is unacceptable, dialing back is a one-line revert.

## deb/rpm: nfpm gzip default → `xz` (both arches)
- Measured on the **exact released 6.21.0 linux-x64 bytes** with the pinned nfpm 2.47.0 in Docker: **deb 74 → 61 MB, rpm 74 → 61 MB (−18%)**; the xz deb apt-installs cleanly on bookworm (and the job's existing deb install smoke re-proves it in CI).
- xz over zstd: compatibility — dpkg has read xz since 2012 and rpm since 4.7 (EL6); zstd would exclude EL7 and older dpkg.

## What didn't make the cut (measured)
- `--minify` on the Bun build: 119 → 116 MB uncompressed, ~1 MB net after compression — not worth minified-server behavior risk.
- Velvet UI exclusion: **correction** — velvet DOES ship in every bundle (an earlier truncated listing said otherwise): ~5 MB uncompressed / ~1.5 MB compressed per bundle. Left alone here since the whole velvet UI is slated for removal, which collects this automatically.
- UPX: never — packed exes are SmartScreen/Smart App Control poison, the opposite of the signing-reputation strategy.
- The floor is the Bun runtime itself (~73% of the payload); only a runtime change moves it.

Expected next-release asset sizes: setup.exe ~42–45 MB, deb ~63/49 MB, rpm ~63/48 MB. The zips are untouched (deflate for universal unzip compatibility).

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1610,6 +1610,14 @@ jobs:
           scripts:
             postinstall: ./pkg/postinstall.sh
             postremove: ./pkg/postremove.sh
+          # xz over nfpm's gzip default: ~25-30% smaller on this binary-heavy
+          # payload (measured against the released bundles), and xz is the
+          # compatibility-safe strong choice (dpkg since 2012, rpm since 4.7 /
+          # EL6 — zstd would exclude EL7 and older dpkg).
+          deb:
+            compression: xz
+          rpm:
+            compression: xz
           overrides:
             deb:
               depends: [libc6]
@@ -1638,6 +1646,10 @@ jobs:
             - src: /opt/mstream/mstream-server
               dst: /usr/bin/mstream-server
               type: symlink
+          deb:
+            compression: xz
+          rpm:
+            compression: xz
           overrides:
             deb:
               depends: [libc6]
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
**Measured on this PR's own CI run** — better than predicted:

| Asset | v6.21.0 (released) | This PR | Δ |
|---|---|---|---|
| `win-x64-setup.exe` | 56 MB | **37.3 MB** | **−33%** |
| rpm x86_64 | 76 MB | 63.5 MB | −16% |
| rpm aarch64 | 59 MB | 45.3 MB | −23% |

The setup.exe is now the **smallest major asset in the release** (darwin pkgs are 40/47 MB).

**Time cost, measured:** win-x64 leg 2m46s vs ~2m25s baseline (ISCC compile: 55.7s total); packages job 3m07s vs 1m54s. ~90 seconds of CI per run, total, for ~50 MB off the downloads.

PR #865, opened 2026-08-20
Auto-update: launcher notify + one-click apply (takeover relaunch)

IrosTheBeggar:
Part 3 of the auto-update feature (after #862 installer fixes and #864 server brain). The tray learns about updates and performs the one thing only it can do: restart mStream into the staged version.

## What's here

**Notify — from the status file, never HTTP.** The launcher reads `update-status.json` (written by #864's checker into the shared data home, the same dir as `launcher.lock`) at startup and on the existing minute tick. A new menu line under the status line renders: disabled *"Up to date (vX)"*, *"Downloading update…"*, *"Update available (vX)"* (opens the hardcoded releases page — no URL is ever taken from the file), or the enabled action — **"Restart to update to vX"** (managed) / **"Install update vX"** (Inno). The status line itself now shows the running version: *"mStream 6.21.2 · up 3h 12m"* (server-reported; `build.rs` stamps the bundle version into every target as the pre-first-boot fallback).

**Apply — the Quit path plus one spawn, never exec.** `stop_current()` gracefully ends the child first (stdin drop → supervision exit, 8s grace) so the port is free and Inno's process sweep finds nothing; then the handoff spawns and the loop exits:
- **Managed**: relaunch the launcher behind `$ROOT/current` — target derived from our own canonicalized exe path (walk up to the root holding `current`), never trusted from the file. `~/Applications` copies relaunch themselves (stable path, contents already refreshed) via `open -n` so LaunchServices identity survives. Original server args (`-j`, `--portable`) ride along.
- **Inno**: spawn the sha256-verified `setup.exe` #864 downloaded — accepted only from our own `updates/` dir with the exact asset-name shape — `/VERYSILENT /NORESTART /MSTREAMRELAUNCH=1`; a param-gated `[Run]` entry added to the `.iss` relaunches the tray after a silent update (the existing entry is `skipifsilent`).
- `applyRequested` in the file (auto mode, or a webapp *"restart to update"* click) triggers the same path on the next tick, exactly once; a failed handoff recovers by respawning the old server.

**`--takeover`** on the relaunched instance: retries the single-instance lock ~12s (fslock releases at old-process death), skips the browser announce, and **forces the tray face**. All status-file content is treated as data: versions shape-checked before display, methods charset-checked, installer paths validated, nothing executable or navigable ever read from it.

## Two bugs the live smoke caught (fixed in this PR)

1. A relaunch inheriting tty-shaped stdio routed into the **console pass-through face** — the update handoff became a bare server and the tray vanished. `--takeover` now forces tray.
2. The inherited pty **died with the old session and SIGHUP'd the update** mid-handoff. The spawn now uses null stdio in its own `setsid` session (Windows: `DETACHED_PROCESS`).

## Verified

- **21 cargo tests** (status parse tolerance incl. hostile inputs — non-triple versions, `"Managed; rm -rf /"` methods, out-of-dir installer paths all render inert; the update-line view matrix; relaunch-target derivation through `current`; `--takeover` never leaks into server argv while `-j --takeover` still forwards verbatim); clippy clean.
- **Full live loop on this Mac** (fake HOME, scratch port): fresh-built launcher + the real #864 server binary installed managed → server staged 9.9.9 from a local feed in `auto` mode and flagged `applyRequested` → the launcher tick applied it: graceful server stop, takeover relaunch through `current`, lock handoff, and the new session serving the same port **from the 9.9.9 tree**, clean teardown (supervision reaped everything).

## Notes

- Depends on #864 at runtime only (it reads the file #864 writes; without it the menu shows the build-stamped "Up to date" line and stays inert). Merge #864 first for coherent history.
- The launcher binaries in `bin/rust-launcher/` regenerate via the usual Build Rust Launcher workflow after merge.

### DIFF
--- a/build/windows-installer.iss
+++ b/build/windows-installer.iss
@@ -96,7 +96,10 @@ Filename: "{app}\mStream.exe"; Description: "Launch mStream"; Flags: nowait post
 ; /MSTREAMRELAUNCH=1 after gracefully stopping mStream and exiting itself.
 ; skipifsilent above would leave the user trayless at the end of a silent
 ; update, so this param-gated twin relaunches exactly (and only) then.
-Filename: "{app}\mStream.exe"; Flags: nowait; Check: WantsRelaunch
+; --takeover: a relaunch mid-update is never a first run — without it the
+; tray would announce (pop the user's browser open) after every silent
+; update, at whatever moment the update happened to land.
+Filename: "{app}\mStream.exe"; Parameters: "--takeover"; Flags: nowait; Check: WantsRelaunch
 
 [Code]
 const
--- a/build/windows-installer.iss
+++ b/build/windows-installer.iss
@@ -92,10 +92,22 @@ Name: "{userdesktop}\mStream"; Filename: "{app}\mStream.exe"; Tasks: desktopicon
 
 [Run]
 Filename: "{app}\mStream.exe"; Description: "Launch mStream"; Flags: nowait postinstall skipifsilent
+; The tray launcher's one-click update runs this installer /VERYSILENT with
+; /MSTREAMRELAUNCH=1 after gracefully stopping mStream and exiting itself.
+; skipifsilent above would leave the user trayless at the end of a silent
+; update, so this param-gated twin relaunches exactly (and only) then.
+Filename: "{app}\mStream.exe"; Flags: nowait; Check: WantsRelaunch
 
 [Code]
 const
   RunKey = 'Software\Microsoft\Windows\CurrentVersion\Run';
+
+{ True only for the launcher-driven silent update (/MSTREAMRELAUNCH=1) —
+  see the param-gated [Run] entry above. }
+function WantsRelaunch(): Boolean;
+begin
+  Result := ExpandConstant('{param:MSTREAMRELAUNCH|0}') = '1';
+end;
 
 function AppPrefix(): string;
 begin
--- a/rust-launcher/build.rs
+++ b/rust-launcher/build.rs
@@ -22,16 +22,23 @@
 // winresource for non-windows targets built on a Windows host. No-op for
 // every non-Windows TARGET.
 fn main() {
-    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() != Ok("windows") {
-        return;
-    }
-    println!("cargo:rerun-if-changed=../build/mstream-logo-cut.ico");
     println!("cargo:rerun-if-changed=../package.json");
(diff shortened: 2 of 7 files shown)


IrosTheBeggar:
**Windows live loop: COMPLETE SUCCESS** — the full Inno auto-update path, on a real Windows 10 desktop, with the real tray.

Harness (mirrors the Mac loop, Windows-flavored): built the branch (+ a local master merge — clean, only build-bun.yml) twice — a 6.21.2 bundle and a 9.9.9 bundle, both carrying THIS PR's launcher — compiled both through the real `.iss` with fresh ISCC, served `manifest.json` (via `release-manifest.sh`) + the 9.9.9 setup.exe from a loopback feed (`MSTREAM_RELEASE_BASE`), installed 6.21.2 silently into an isolated tree (`/DIR` + `LOCALAPPDATA` redirect; HKCU Run + Inno uninstall key backed up and byte-exact restored after).

The loop, all green:
1. **Tray face via wscript launch** → launcher (v6.21.2.0) + supervised server up on the scratch port.
2. Server check hit the feed → validated the 9.9.9 manifest → **downloaded + sha256-verified** `mStream-9.9.9-win-x64-setup.exe` into `userDataHome()/updates/` → `update-status.json` written: `staged: true, method: inno, installerPath` set.
3. `POST /api/v1/admin/update/apply` → `applyRequested` → on the launcher tick: graceful server stop, **silent Inno upgrade** (`/VERYSILENT /MSTREAMRELAUNCH=1`), the param-gated `[Run]` relaunch with `--takeover` → **new launcher + server, both v9.9.9.0, same port**, fresh pids, no strays, status self-corrected to `current: 9.9.9` on the new boot.
4. Silent uninstall + full teardown clean.

Two harness-side notes that CONFIRM design decisions rather than contradict them:
- Launching `mStream.exe` from a console produced the **console pass-through face** (bare server) — and that unsupervised server correctly **never wrote** `update-status.json` (`supervisedByLauncher()` gate working exactly as documented: an npm/console instance can't clobber the tray's channel). The wscript (no-console) launch is the double-click-equivalent.
- The scratch launcher's first-run autostart self-enable rewrote HKCU Run to the scratch path — expected per-user-registration semantics; the harness restored the original byte-exact.

The branch is behind master (#866/#867); my test ran on a local merge that was clean — worth merging master in before this lands.

IrosTheBeggar:
(Post-merge note: this PR merged moments before — the master merge that the Windows live loop was tested against is equivalent to what the merge to master produced, both parents identical. The PR branch is deleted; nothing further needed.)

PR #878, opened 2026-08-21
fix(webapp): local search rewrote row filepaths outside the file explorer

IrosTheBeggar:
## The bug

The default UI's local search bar doesn't hide rows when you type — it **re-renders** the list from `currentBrowsingList`. That only works if each browse panel pushed enough data to redraw its own rows, and the DB-driven panels pushed `{ type: 'file', name }` and nothing else. So `runLocalSearch` fell back to reconstructing the path:

```js
const fileLocation = x.path || getFileExplorerPath() + x.name;
```

That rule only ever held for the file explorer. Inside an album view it rewrote every row's `data-file_location` from the real path to a stale explorer dir plus the song title:

```
music/The Wall/CD1/01 In the Flesh.flac   →   music/In the Flesh
```

Clicking or queueing a filtered row then 404'd and played nothing. Back-navigation replays the saved filter (`previousSearch` + a synthetic `keyup`), so it also corrupted rows the user never typed over.

## The fix

Invert the contract. Each panel pushes the exact arguments it passed to the renderer, and `runLocalSearch` replays them verbatim:

```js
filelist += createMusicFileHtml(x.path, x.title, x.aa, x.rating, x.subtitle);
```

The filtered render now calls the same function with the same values as the original render, so the two can't drift — this is a structural fix rather than a per-panel patch.

## What the audit turned up beyond the reported bug

Auditing every panel that pushes `{ type: 'file' }` found the missing path was one symptom of a broader omission:

| Panel | Was losing on filter |
|---|---|
| `getAlbumSongs`, `getGenreSongs` | path (the reported bug) |
| `getRatedSongs` | path, **album art, star rating**, artist subtitle |
| `redoRecentlyPlayed` / `redoMostPlayed` / `redoRecentlyAdded` | path, **album art**, artist subtitle |
| `printdir` (file explorer) | dropped the server's authoritative `file.path`; never pushed `artist`/`title`, so `Artist - Title` labels collapsed to the filename |
| `getAllGenres` | **no branch existed** — genres fell through to the file-row `else` and rendered as playable rows pointing at the genre name |
| `getAllAlbums`, `getArtistsAlbums` | `data-year`, which `getAlbumsOnClick` reads back — a filtered album could open a *different* album of the same name |

`renderGenre()` was extracted so the genre list has one renderer, matching `renderArtist`/`renderAlbum`.

## Testing

**Automated — 12 new tests.** `webapp/` has no browser harness, so these follow the existing `webapp-waveform-cache.test.mjs` pattern: slice the real functions out of `m.js` and run them against stub globals, rather than re-typing the implementation where it could drift from what ships. They pin a stronger invariant than "the path survives" — **a row that survives the filter must be byte-identical to the row the panel drew** — which also catches the art/rating/subtitle losses. All 12 fail on `master`.

Reverting *only* the `runLocalSearch` fallback (keeping `renderGenre` and the fixed producers) reproduces the exact reported corruption, confirming the tests bite for the right reason:

```
+ actual - expected
+   'music&#x2F;In the Flesh?'
-   'music&#x2F;The Wall&#x2F;CD1&#x2F;01 In the Flesh.flac'
```

**Live end-to-end with a negative control.** Real `keyup`, real `.click()` on the filtered row, real range request against the queued URL:

| | filtered row's path | server response |
|---|---|---|
| Old code | `music/Here Comes the Sun` | **404** `text/html` |
| Fixed | `music/The Beatles/Abbey Road/07 - Here Comes the Sun.flac` | **206** `audio/x-flac` |

Also driven live: genre filter stays a genre row; back-navigation replay (drilled into a genre, hit back → search bar restored, rows still genre rows); recently-added filter with album art and subtitle intact.

**Suite:** 536/539 unit tests pass, 0 failures. New test file lints clean (`webapp/` is eslint-ignored).

## Reviewer notes

- **Not covered:** 5 of 8 `runLocalSearch` branches (`directory`, `playlist`, `artist`, playlist-state rows, `m3u`) — the fixture has `directories: []` and no m3u entries. None of those five were changed behaviorally.
- **Not run:** full `npm test`. Unit suite only; the change is webapp-only and touches no server code. Worth the `full-ci` label.
- Starred/rated, most-played, file-explorer and album-grid filters rest on the unit tests — the preview library has zero rated songs, so that panel wasn't exercised live.
- No coverage for null album names (the `SINGLES` path), missing `metadata.title`, or filepaths containing `#`/`?`/`%`.
- **Velvet is deliberately untouched:** `webapp/velvet/alpha/m.js:1588` and `webapp/velvet/assets/js/mstream.js:1198` still carry the old fallback, but velvet's live UI is `index.html` → `/app.js`, which has no `runLocalSearch` or `currentBrowsingList` at all. Those copies are reachable only via `old.html` — dead code, and velvet is slated for removal.

### DIFF
--- a/test/unit/webapp-local-search.test.mjs
+++ b/test/unit/webapp-local-search.test.mjs
@@ -1,0 +1,422 @@
+/**
+ * runLocalSearch() in webapp/alpha/m.js — the local filter bar above the
+ * browse column.
+ *
+ * The filter does not hide rows, it RE-RENDERS them from
+ * `currentBrowsingList`. That only works if the entries each panel pushes
+ * carry everything the row needs. They didn't: the DB-driven panels pushed
+ * `{ type: 'file', name }` and nothing else, so the re-render fell back to
+ * `getFileExplorerPath() + x.name` — a rule that only ever held for the file
+ * explorer. Inside an album view a row went from
+ *
+ *     music/The Wall/CD1/01 In the Flesh.flac      (unfiltered)
+ *     music/In the Flesh                           (after typing one char)
+ *
+ * and clicking or queueing it played nothing. Back-navigation replays the
+ * saved filter (`previousSearch` + a synthetic keyup), so the corruption also
+ * hit rows you never typed over.
+ *
+ * The invariant pinned here: for every panel, filtering is *lossless* — a row
+ * that survives the filter is byte-identical to the row the panel drew. That
+ * covers the reported data-file_location bug and the quieter losses in the
+ * same code path (album art, star rating, artist subtitle, album year).
+ *
+ * webapp/ has no browser test harness, so rather than re-typing the
+ * implementation this slices the real functions out of m.js and runs them
+ * against stub globals. If they are renamed or restructured the slice fails
+ * loudly rather than silently testing nothing.
+ */
+
+import { describe, test } from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const M_PATH = path.resolve(__dirname, '..', '..', 'webapp', 'alpha', 'm.js');
+const SRC = fs.readFileSync(M_PATH, 'utf8');
+
+// ── source slicing ──────────────────────────────────────────────────
+// Brace-match from a `function NAME(` / `async function NAME(` header to
+// its closing brace. Naive on braces inside strings, which is fine for
+// these functions: every one of them ends at column 0.
+function sliceFn(name) {
+  const lines = SRC.split(/\r?\n/);
+  // m.js writes both `function name (` and `function name(`.
+  const header = (l) => l.split(' (').join('(');
+  const start = lines.findIndex((l) =>
+    header(l).startsWith(`function ${name}(`) ||
+    header(l).startsWith(`async function ${name}(`));
+  assert.ok(start >= 0, `m.js no longer defines ${name}() at top level — update this test`);
+
+  for (let i = start; i < lines.length; i++) {
+    if (lines[i] === '}') return lines.slice(start, i + 1).join('\n');
+  }
+  throw new Error(`could not find the end of ${name}()`);
+}
+
+function sliceConst(name) {
+  const lines = SRC.split(/\r?\n/);
+  const start = lines.findIndex((l) => l.startsWith(`const ${name} = `));
+  assert.ok(start >= 0, `m.js no longer defines const ${name} — update this test`);
+  for (let i = start; i < lines.length; i++) {
+    if (lines[i] === '};') return lines.slice(start, i + 1).join('\n');
+  }
+  throw new Error(`could not find the end of const ${name}`);
+}
+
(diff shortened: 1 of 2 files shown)


IrosTheBeggar:
Rebased onto master (2026-09-04) over the federation peer-browse work (#927), which had reshaped the same panels — six conflict hunks. Two adaptations beyond picking sides, both covered by new tests:

- **`renderGenre()` now carries `${peerAttr()}`**, like the inline genre row it replaced. Without it a peer's genre rows lose `data-peer`, and `getGenreSongsList` → `adoptPeer(el)` would open the *local* genre.
- **The hoisted `aa` values use master's peer-aware `artImgAttr(…, 's')`** instead of the pre-federation local-server URL. Without it, filtered rows on a peer would point their album art at this server.

The test sandbox now slices the peer helpers (`peerAttr`, `browseApi`, `localIgnoreVPaths`, `artUrl`, `artImgAttr`) and declares `peerContext` / `browseGeneration`; two new tests set a peer context and require filtered song and genre rows to keep `data-peer` and the peer art proxy URL, byte-identical to the panel's own render. 14/14 in the file, plus the other m.js-slicing test still green.

Review notes (nothing blocking):
- Every `{ type: 'file' }` push on the rebased tree now carries `path` + `title` (file explorer, genre/album songs, starred, recently/most played, recently added); the playlist view's rows go through the `state === 'playlist'` branch with `filepath`/`lokiId`/`metadata` as before.
- `getRatedSongs` still pushes `metadata: value.metadata` alongside the explicit args — harmless, just redundant now.
- `printdir` computes `title` for `.m3u` rows too — unused, harmless.
- The slicer's brace-matching relies on each sliced function ending with `}` at column 0; it fails loudly (assert) rather than silently, which is the right failure mode for a refactor.

IrosTheBeggar:
Smoke run on the rebased head (02d7126b, 2026-09-04): the default UI served by `node cli-boot-wrapper.js` as a fresh user-less instance over the 9-track `test/fixtures/music` library, driven in a real browser with real `keyup` events on `#localSearchBar` and `.click()` on the filtered rows; row attributes read back from the DOM.

| Flow | Result |
|---|---|
| Albums → Night Drive → type `neon` | 1 row, `data-file_location` = `music/Vosto/Night Drive/02 - Neon.mp3`, byte-identical to the unfiltered row; clicking it loaded the track (tab title → "Neon - Vosto"); `GET /media/<that path>` with a Range header → **206** `audio/mpeg` |
| Genres → type `ambi` | only the Ambient genre row survives, still a genre row (no `data-file_location`), identical |
| Back-navigation replay: drill into Ambient (3 songs) → back | search bar restored to `ambi`, rows are the same filtered genre rows |
| Recent Added → type `rise` | 1 row with the real path, album-art `src` and artist subtitle intact, identical |
| File Explorer → Vosto → Night Drive → type `neon` / `static` | 1 row with the server's own path, identical; clicked row's path serves **206** |
| **Negative control**: same album flow with master's `m.js` swapped in | filtered row rewritten to `music/Neon`, `GET /media/music/Neon` → **404** `text/html` — the reported bug, reproduced on the old code and gone on this branch |

Not smoked here (no fixture data): Starred/Rated, Most Played, and the peer views — those rest on the unit tests (14/14 in `test/unit/webapp-local-search.test.mjs`, including the two peer-context cases).

PR #879, opened 2026-08-23
feat(api): report discovery readiness on GET /api/

IrosTheBeggar:
## The gap

The ping's `discovery` flag says the feature is switched **on**. It cannot say whether the scan has produced anything yet — and those are different questions with the same symptom for a client.

A server with discovery enabled and an unfinished scan looks fully capable. The client offers sonic similarity, sends `similarTo` / `minSimilarity` on `random-songs`, and every pick 400s against an empty pool. In the mobile app that means Auto DJ stops advancing the queue with nothing on screen to explain it — the feature was offered, accepted, and then silently did nothing.

`features.discoveryReady` closes that in the call clients already make to learn the server version:

```json
{
  "server": "6.22.0",
  "apiVersions": ["1"],
  "features": { "subsonic": false, "discoveryReady": false }
}
```

Paired with the ping flag it separates the three states cleanly:

| ping `discovery` | `discoveryReady` | Meaning |
|---|---|---|
| `false` | — | Feature off |
| `true` | `false` | **On, nothing analysed — offer nothing yet** |
| `true` | `true` | Usable |

## Two choices driven by this endpoint being public

**A boolean, not a count.** How many tracks are analysed is library-size information, and no client needs it to decide whether to offer the feature.

**A one-row existence check, not `getIndex()`.** `getIndex()` rebuilds the in-memory index whenever its cache is stale, so calling it from an unauthenticated route would let anyone force repeated index builds on a large library. `hasEmbeddings()` short-circuits on `collectDiscoveryData` before touching the database at all, then runs a single `LIMIT 1` lookup, and swallows any error — a capability probe must never be able to 500 the endpoint that reports it.

It also filters on `model_id`: vectors in another model's space cannot answer queries in the configured one, so "has rows" is not the same as "ready".

## Testing

Six tests in `test/api-discovery-readiness.test.mjs` cover the predicate: off-in-config returns false **without opening a database** (the short-circuit is the security-relevant part), empty table, NULL embedding, wrong model, one good row, and an unusable store not throwing.

Verified live against 6.22 with discovery enabled and an unscanned library — ping reports `discovery: true`, `/api/` reports `discoveryReady: false`, which is exactly the case that has been breaking clients.

OpenAPI spec updated.

## Note

I considered `/api/v1/federation/health`, which already exposes `analyzedCount`. Two reasons against it: it only exists from 6.17.2, and it is the one federation route that does **not** check `federation.enabled` while every sibling does — which looks like an oversight rather than a contract, so it is a fragile thing for clients to depend on. Worth a separate look either way.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -957,9 +957,31 @@ paths:
                   apiVersions:
                     type: array
                     items: { type: string }
+                  features:
+                    type: object
+                    description: >
+                      Capability flags a client can gate UI on without an extra
+                      round-trip.
+                    properties:
+                      subsonic:
+                        type: boolean
+                        description: Whether the Subsonic API surface is mounted.
+                      discoveryReady:
+                        type: boolean
+                        description: >
+                          Whether a sonic-similarity query would find anything
+                          right now. Distinct from the ping's `discovery` flag,
+                          which reports only that the feature is enabled: a
+                          server can have discovery on with an unfinished scan,
+                          in which case similarity queries fail against an empty
+                          pool. A boolean rather than a count because this
+                          endpoint is public.
               example:
-                server: "6.4.2"
+                server: "6.22.0"
                 apiVersions: ["1"]
+                features:
+                  subsonic: false
+                  discoveryReady: true
 
   /api/v1/ping:
     get:
--- a/src/db/discovery-similarity.js
+++ b/src/db/discovery-similarity.js
@@ -77,6 +77,34 @@ function l2normalize(v) {
  *   artists: Map<artistName, { vec, analyzedCount, topTags }>,
  * }
  */
+/// Cheapest possible "would a similarity query find anything?" — one indexed
+/// existence check, no vector decode, no index build.
+///
+/// Deliberately NOT getIndex(): that rebuilds the in-memory index whenever the
+/// cache is stale, and the only caller of this is the PUBLIC /api/ endpoint.
+/// Letting an unauthenticated request force an index build would be a free
+/// amplification vector on a large library.
+///
+/// Returns false when discovery is switched off, the DB has never been
+/// created, or nothing has been embedded yet for the configured model — which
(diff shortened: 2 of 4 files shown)


IrosTheBeggar:
Client-side follow-up, noted here so it doesn't get lost after this merges.

The app (mstream_music) currently learns this the hard way: it sends `similarTo`/`minSimilarity`, gets back `no songs are within the similarity range` from a server whose scan hasn't reached the library yet, and only then drops the constraint for the session and keeps playing on a widened pool. That works, but it costs one visibly degraded pick and a toast the user didn't need to see.

Once `features.discoveryReady` is live, the app should read it from the `GET /api/` call it already makes for the version banner, and use it as a **pre-filter** rather than a source of truth:

- `discovery: true` + `discoveryReady: false` → don't offer the constraint at all this session; the switch reads as unavailable-yet rather than failing on first use.
- `discoveryReady: true` → send it, and keep the existing learn-from-rejection path as the backstop. The endpoint answers "are there any vectors", not "is this specific track analyzed", so a partial scan can still miss an individual pick — the fallback stays load-bearing and shouldn't be removed.

Deliberately not building against it yet: the app shouldn't ship a branch on a field no released server sends. Worth doing after this lands and a release carries it, at which point the version floor for the field is known and can go in the app's feature table alongside the others.

Two things about the shape here that I'd rather have questioned now than after release, since `GET /api/` is unauthenticated and public:

1. **Boolean, not a count.** `analyzedCount` would leak library size to anyone who can reach the port. No client needs the number to decide whether to offer the feature.
2. **`LIMIT 1` existence check, not `getIndex()`.** `getIndex()` rebuilds the in-memory index when its cache is stale — reachable from an unauthenticated route, that's a way to make a large library do repeated index builds on demand. `hasEmbeddings()` short-circuits on the config flag before touching the DB at all, and swallows errors so a capability probe can't 500 the endpoint that reports it.

Issue #154, opened 2018-11-19
permissions error on /save/db


aroundmyroom:
As the db resides now under mStream I have some issues but maybe it is all user error

git pull is used as user 'root' as things need to be compiled
mStream is running as normal user
When mStream is running as normal user, but mStream is in a folder not owned by the user we get errors. This is obvious, 

I think /save/db /save/log etc. should not be created by the git pull but by the user running mStream.
Otherwise you will get issues between the git pull and user running mStream.

I am not even sure if a sudo npm link etc will help or sudo git pull 

2018-11-19T07:43:35.489Z error: File scan error: error : Error: Error: EACCES: permission denied, open '/home/dennis/mStream/save/db/files.loki-v1.db~'

NB. I solved it manually, but I think other users could get in trouble when there is a difference between the user running mStream and the mStream binary.

IrosTheBeggar:
The `EACCES: permission denied` error was due to to a bug from the v4 merge that has been fixed now.  I forgot to push the /save/logs and /save/db directory causing a fail to happen during write

mStream saves everything to its own directory by default because it's simple and works most of the time.  I also added a wizard to modify the config through a CLI program.  This should help guide new users to the right options to choose where to save their data if the standard config doesn't work for them

aroundmyroom:
I pulled and the folders where there, but as I pull as root the folders are chmodded with root
I do not run mStream as root but as a user, as this user has no root rights it will get an access denied on those folders as mStream want to write its data as that user not as root

I will change this that all files in my folder are owned by that user, but I wonder if than things can be compiled correctly if necessary (I believe that sudo comes around the corner than)

IrosTheBeggar:
Did you run git pull ad sudo? That's the only reason those folders would be under root

aroundmyroom:
thats correct, but I will change that, as I had on my previous VM box issues with compiling node
now it resides on another VM and there I had no issues with node 10.x so I will chown this afternoon or later today everything

Issue #146, opened 2018-11-06
scanning 


aroundmyroom:
It seems that after the update of the config.json my db is not scanned at the time interval set

this is the log .. 
from start of mStream I only have these log entries:
{"message":"File scan started on /media/music/","level":"info","timestamp":"2018-11-02T13:13:51.640Z"}
{"message":"File scan message: database saved","level":"info","timestamp":"2018-11-02T13:14:31.106Z"}
{"message":"File scan completed with code 0","level":"info","timestamp":"2018-11-02T13:14:31.329Z"}
{"message":"File scan started on /media/music/","level":"info","timestamp":"2018-11-03T13:14:04.013Z"}
{"message":"File scan message: database saved","level":"info","timestamp":"2018-11-03T13:29:24.996Z"}
{"message":"File scan completed with code 0","level":"info","timestamp":"2018-11-03T13:29:25.401Z"}
{"message":"File scan started on /media/music/","level":"info","timestamp":"2018-11-04T13:13:59.430Z"}
{"message":"File scan message: database saved","level":"info","timestamp":"2018-11-04T14:02:37.189Z"}
{"message":"File scan completed with code 0","level":"info","timestamp":"2018-11-04T14:02:38.911Z"}
{"message":"File scan started on /media/music/","level":"info","timestamp":"2018-11-05T13:14:03.939Z"}
{"message":"File scan message: database saved","level":"info","timestamp":"2018-11-05T15:15:25.426Z"}
{"message":"File scan completed with code 0","level":"info","timestamp":"2018-11-05T15:15:27.517Z"}
{"message":"File scan started on /media/music/","level":"info","timestamp":"2018-11-06T13:13:56.365Z"}
{"message":"File scan message: database saved","level":"info","timestamp":"2018-11-06T13:35:10.051Z"}
{"message":"File scan completed with code 0","level":"info","timestamp":"2018-11-06T13:35:16.814Z"}
dennis@docker:~/mStream-data$ more config.json

This is the part of the config .. 
dennis@docker:~/mStream-data$ more config.json
{
  "port": 3000,
  "userinterface":"public",
  "scanOptions":{
    "scaninterval": 2,
    "saveInterval": 250,
    "skipImg": false,
    "pause": 50,
    "bootScanDelay": 60
  },

scaninterval is 2 but what the log shows is something different

IrosTheBeggar:
try `scanInterval` with the i capitalized

aroundmyroom:
[...]

PR #880, opened 2026-08-24
fix(discovery): recover the p2p sidecar when it dies, and rejoin the topic

IrosTheBeggar:
> **Handoff:** full investigation record, verification status, reproduction steps and open items are in [this comment](https://github.com/IrosTheBeggar/mStream/pull/880#issuecomment-5398359448). **CI on this PR is build-only** — `test.yml` runs only on the `full-ci` label or `workflow_dispatch`, so the suite results below are local.

## What happened

A live server's p2p-sidecar was `SIGKILL`ed at `09:08:48Z` and the server spent the next **6h45m off the discovery network** with the config still saying `enabled`. Diagnosed from operator logs, then reproduced locally end to end.

The network itself was healthy throughout — a fresh node joins in ~3s, reaches neighbours, and pulls snapshots normally. This was entirely a local failure to recover.

```
09:08:48Z  warn  [p2p-sidecar] exited (code=null signal=SIGKILL)
09:21:14Z  info  [p2p-sidecar] ready — endpointId=99e4fd0beb32…   ← respawned by an hourly
09:21:18Z  warn  [discovery-peer-dbs] rotation fetch … failed        rotation fetch, 3.7s earlier
```

That respawn is the trap. `publish`/`fetch`/`announce` all lazily `start()` the sidecar **without joining the topic**, so the rotation fetch brought the *process* back as a bare process — `running: true`, `joined: false`, invisible. Strictly worse than staying down: it looks healthier while being just as disconnected. The same pattern had already happened once on 08-16.

## Three defects

**1. Nothing put a dead sidecar back.** `p2p.start()` has exactly one real caller (the stack's boot), and both periodic passes that could have noticed — the mesh-health watch and the catalog prune — return early on `!isRunning()`. The child now reports its own unexpected death through a private hook, and the stack **replays itself** on a 5s/15s/60s/5min ladder that never gives up. A replay, not a respawn: the topic subscription, announce payload and holds beacon all live in the sidecar's memory. Deliberate stops are classified via the existing `startGen` counter and never trigger recovery.

**2. The admin enable route answered `200` and did nothing.** `running` recorded that a start *completed*, never that the sidecar was *alive*, so re-enabling after a crash hit `if (running) return` — the exact button an operator presses to fix an outage, replying success while repairing nothing. Verified against the running server before the fix:

```
POST /api/v1/admin/discovery/p2p/enabled {"enabled":true}
→ {"enabled":true,"collectForced":false}      # success…
→ status: {"running":true,"joined":false,"neighbors":0}   # …and still off the network
```

Same bug class as M3 in ddcf03ca, which fixed the reboot race but not this one.

**3. An RPC racing the kill took down the whole server.** Between the child dying and Node delivering `'exit'`, `proc` is still set and its stdin is a broken pipe. `rpc()` writes, `EPIPE` lands on the stream, and an unhandled `'error'` on a stream is an uncaught exception. **The status route polls exactly there** — so an admin watching their Discovery page during an OOM kill lost their music server, library and all. Found by the new test suite, not by inspection.

The mesh-health watch also stops skipping the un-joined case (`if (!s.joined || …) return`) — the single most broken state a live sidecar can be in, and the one the reported server was stuck in.

## Tests

- `test/unit/discovery-p2p-crash-recovery.test.mjs` — models the flag/timer contract with **negative controls** that reproduce the previous implementation and fail it (the precedent set by `discovery-p2p-stack-race.test.mjs`). Always runs.
- `test/integration/discovery-p2p-crash-recovery.test.mjs` — kills a **real** sidecar and asserts it comes back **joined**, keeps its persisted endpoint identity, leaves no orphan, and that a burst of RPCs racing the kill fails the requests rather than the process. Hermetic (no network, no seeds, no peers: the sidecar reports `joined:true` on an empty bootstrap set). Skips where no sidecar binary exists, like the gossip layers of the existing suite.

Negative-controlled: with the stdin handler removed, all three integration tests fail with the server dead.

```
unit                     542 pass / 0 fail
discovery integration     44 pass  +  3 new pass
lint                      unchanged (80 pre-existing errors on HEAD, 0 added)
```

## Not in this PR

Two things I found and deliberately left out of scope:

- **Hourly rotation churn.** 333 warnings since 08-10, alternating between the same two unreachable peers (`cannot reach the blob's origin: No addressing information available`). The per-peer backoff caps at 60min while rotation runs hourly, so it has always expired by the next pass — it retries forever. Harmless, but it was the *only* p2p line in the logs, so the real failure landed in noise that looks exactly like business as usual.
- **The admin panel renders a dead sidecar as healthy-ish.** `meshSearching` is true whenever `!running || !joined || neighbors === 0`, so the panel shows "not joined yet" over a *"searching for peers"* progress bar — indistinguishable from normal startup, indefinitely.

Happy to take either as a follow-up.

### DIFF
--- a/src/state/discovery-p2p-stack.js
+++ b/src/state/discovery-p2p-stack.js
@@ -18,6 +18,55 @@ let stopping = null;
 
 export function isStackRunning() { return running; }
 
+// Recovery backoff after an unexpected sidecar death. Short first — the
+// common cause is a one-off OOM kill, and seconds off the mesh cost nothing
+// — then widening, so a sidecar that CANNOT stay up (bad binary, wedged data
+// dir, a memory limit it will always hit) settles at ~12 attempts an hour
+// instead of hot-looping. It never gives up: the failure this fixes is a
+// server quietly staying off the network forever, and a recovery loop that
+// exhausts itself would just reintroduce it with extra steps.
+const RECOVERY_DELAYS_MS = [5000, 15000, 60000, 300000];
+let recoveryTimer = null;
+let recoveryAttempts = 0;
+
+function cancelRecovery() {
+  if (recoveryTimer) { clearTimeout(recoveryTimer); recoveryTimer = null; }
+  recoveryAttempts = 0;
+}
+
+// Called by discovery-p2p.js when the child died without us asking. Replays
+// the ENTIRE stack — subscribe, spawn, join, re-announce, holds — because
+// everything the sidecar knew about the mesh died with the process.
+function scheduleRecovery(why) {
+  // `running` is the feature's INTENT. A crash leaves it true (that is the
+  // stale flag this module also repairs below), so this reads "we are
+  // supposed to be on the network"; a deliberate stop clears it first and
+  // therefore silences recovery, which is exactly right.
+  if (!running || recoveryTimer) { return; }
+  const delay = RECOVERY_DELAYS_MS[Math.min(recoveryAttempts, RECOVERY_DELAYS_MS.length - 1)];
+  recoveryAttempts += 1;
+  winston.warn(`[discovery-p2p] sidecar ${why} — replaying the stack in `
+    + `${Math.round(delay / 1000)}s (attempt ${recoveryAttempts})`);
+  recoveryTimer = setTimeout(() => {
+    recoveryTimer = null;
+    if (!running) { return; }
+    // Clear the dead sidecar's stale intent HERE so the start below does the
+    // real work instead of tripping the "already running" guard.
+    running = false;
+    startDiscoveryP2pStack()
+      .then(() => winston.info('[discovery-p2p] sidecar recovered — rejoined the catalog topic'))
+      .catch((err) => {
+        winston.warn(`[discovery-p2p] sidecar recovery failed: ${err.message}`);
+        // Re-arm: startDiscoveryP2pStack() left `running` false on failure,
+        // and scheduleRecovery reads it as "give up". The intent has not
+        // changed — the config still says enabled — so restore it.
+        running = true;
+        scheduleRecovery('recovery failed');
+      });
+  }, delay);
+  if (recoveryTimer.unref) { recoveryTimer.unref(); }
+}
+
 export async function startDiscoveryP2pStack() {
   // Serialize behind an in-flight stop. A soft reboot stops the stack and then
   // re-serves, and stopping the sidecar can take up to ~10s (a shutdown RPC
@@ -30,13 +79,28 @@ export async function startDiscoveryP2pStack() {
   if (stopping) {
     try { await stopping; } catch (_err) { /* stop's own caller logs it */ }
   }
-  if (running) { return; }
+  if (running) {
+    // `running` records that a start COMPLETED — never that the sidecar is
+    // still alive. An OOM kill or a crash leaves it stale, and this is the
+    // route the admin "enable the network" button lands on: it answered 200
+    // {"enabled":true} while repairing precisely nothing, which is the worst
+    // possible reply to an operator trying to fix an outage. Ask the sidecar
+    // itself, and if it is gone treat THIS call as the repair it looks like.
+    // The import is free here: a true `running` means the module is resident.
+    const p2p = await import('./discovery-p2p.js');
+    if (p2p.isRunning()) { return; }
+    winston.warn('[discovery-p2p] the stack is marked running but the sidecar is gone — replaying the start');
+    running = false;
+  }
   if (starting) { return starting; }
   starting = (async () => {
     const p2p = await import('./discovery-p2p.js');
     const catalog = await import('./discovery-catalog.js');
     const seeds = await import('./discovery-seeds.js');
     catalog.subscribe();
+    // Armed BEFORE the spawn, so a sidecar that dies seconds into its life
+    // is covered too. Idempotent — re-registering just replaces the slot.
+    p2p.setUnexpectedExitHandler(scheduleRecovery);
     await p2p.start();
     // Two-phase bootstrap. Phase one joins the topic IMMEDIATELY with
     // what's known locally (baked seeds + cached list + the operator's
@@ -71,6 +135,10 @@ export async function startDiscoveryP2pStack() {
     // whose snapshot we hold stays listed however long it's been silent.
     catalog.startPruning(() => new Set(peerDbs.list().map((e) => e.endpointId)));
     running = true;
+    // We are on the network again: retire any pending retry and reset the
+    // backoff, so the NEXT crash starts from 5s rather than wherever this
+    // one's ladder left off.
+    cancelRecovery();
   })();
   try { await starting; } finally { starting = null; }
 }
@@ -86,6 +154,10 @@ export async function stopDiscoveryP2pStack() {
   // resolved was the bug — the window is seconds wide, and a reboot lands
   // squarely inside it.
   running = false;
+  // A pending retry outlives the flag it reads, so drop it here as well —
+  // otherwise a crash landing just before an operator disables the feature
+  // would resurrect the sidecar seconds after they turned it off.
+  cancelRecovery();
   stopping = (async () => {
     const seeds = await import('./discovery-seeds.js');
     const peerDbs = await import('./discovery-peer-dbs.js');
@@ -94,6 +166,10 @@ export async function stopDiscoveryP2pStack() {
     peerDbs.stopAutoFetch();
     catalog.stopPruning();
     const p2p = await import('./discovery-p2p.js');
+    // Nothing to recover from a shutdown we asked for. (stop() also bumps
+    // its generation counter, so the exit is classified as expected anyway —
+    // this is the belt to that suspenders.)
+    p2p.setUnexpectedExitHandler(null);
     await p2p.stop();
   })();
   try { await stopping; } finally { stopping = null; }
--- a/src/state/discovery-p2p.js
+++ b/src/state/discovery-p2p.js
@@ -40,6 +40,29 @@ import * as config from './config.js';
 // The catalog module (discovery-catalog.js) is the main subscriber.
 export const events = new EventEmitter();
 
+// Unexpected-death hook, set by the stack (discovery-p2p-stack.js).
+//
+// The sidecar is a child process, so an OOM kill, a panic, or an operator's
+// `kill` takes it down with nobody asking — and NOTHING used to put it back:
+// start() has exactly one real caller (the stack's boot), and the two
+// periodic passes that could have noticed (the mesh-health watch, the
+// catalog prune) both return early on `!isRunning()`. A server therefore sat
+// off the network, config still saying "enabled", until someone restarted it
+// by hand.
+//
+// Deliberately NOT routed through `events` above: that emitter re-emits
+// whatever the child names in its `event` field, so a buggy sidecar could
+// synthesize its own resurrection. This is a private channel the child
+// cannot reach.
+//
+// The handler gets the whole stack replayed, not a bare respawn — the topic
+// subscription, the announce payload and the holds beacon all live in the
+// sidecar's memory and die with it. A respawn without them yields the worst
+// state of all: a process that reports `running: true` while being just as
+// invisible to the network as no process at all.
+let unexpectedExitHandler = null;
+export function setUnexpectedExitHandler(fn) { unexpectedExitHandler = fn; }
+
 const ext = process.platform === 'win32' ? '.exe' : '';
 // musl detection mirrors task-queue.js's rust-parser resolution.
 const isMusl = process.platform === 'linux' && !process.report?.getReport()?.header?.glibcVersionRuntime;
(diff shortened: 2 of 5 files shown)


IrosTheBeggar:
## Handoff notes — full investigation record

Everything below is what I verified, how I verified it, and what I deliberately left open. Written for whoever picks this up next.

---

### 1. Verification status — read this first

**What is proven:**

| Check | Result | Where |
|---|---|---|
| Unit suite | 542 pass / 0 fail | local, darwin-arm64 |
| Existing p2p integration | 44 pass / 0 fail | local, darwin-arm64, real sidecar |
| New crash-recovery integration | 3 pass / 0 fail | local, darwin-arm64, real sidecar |
| Other discovery integration (similarity, federation, export) | pass | local |
| Lint | 145 problems vs **141 on HEAD** — 0 new errors, +4 `require-await` warnings | local |
| Build matrix, 9 platforms + installer contracts | pass | CI run 32750086361 |

The 4 added warnings are `async` model functions with no `await` in the new unit test. That is the same idiom as the sibling `discovery-p2p-stack-race.test.mjs`, which contributes one identical warning today — the functions model an async API and the tests `await` them.

**What is NOT proven, and why:**

- **The test suite did not run in CI for this PR.** `.github/workflows/test.yml` triggers only on the `full-ci` label or `workflow_dispatch`. The 11 green checks on this PR are `build-bun.yml` (9-platform build matrix) plus installer contracts — none of them run `npm test`. To get a real CI signal:
  ```bash
  gh workflow run test.yml --ref claude/p2p-discovery-outage-75c7ae
  ```
  I did not dispatch it: the workflow comment says the heavy matrix is for "once you mark it ready to merge", which is the maintainer's call, not mine.
- **The new integration suite skips in CI regardless.** It needs a real sidecar binary, and none exists in CI since the sidecar moved to its own repo (fetch-on-first-use, and `startServer` pins `MSTREAM_SIDECAR_BASE` to a dead port so no suite can pull one mid-test). Same situation as the existing gossip layers. That is precisely why the unit coverage exists in parallel — it runs everywhere and carries the negative controls.
- **Cross-platform behaviour of the new code is unverified beyond compiling.** The stream `'error'` handlers and the recovery timers are the parts I would want a Linux and Windows run to confirm.

**To run the new tests locally you need a sidecar binary** — either a dev build at `p2p-sidecar/target/release/p2p-sidecar` (clone `IrosTheBeggar/mstream-p2p-sidecar`, `cargo build --release`) or an operator prebuilt in `bin/p2p-sidecar/`. Without one the integration suite reports as skipped, not failed.

```bash
node --test --test-reporter=spec test/unit/discovery-p2p-crash-recovery.test.mjs
node --test --test-reporter=spec test/integration/discovery-p2p-crash-recovery.test.mjs
```

---

### 2. The diagnosis, in full

Reconstructed from 337 hourly log files spanning 2026-08-10 → 2026-08-24. All times UTC.

```
08-20 01:21:11  Database initialized / Access mStream locally    ← last real boot
08-20 01:21:13  [p2p-sidecar] ready                              ← boot stack: spawn AND join. healthy.
...
08-24 09:08:48  warn  [p2p-sidecar] exited (code=null signal=SIGKILL)
08-24 09:21:14  info  [p2p-sidecar] ready — endpointId=99e4fd0beb32…
08-24 09:21:18  warn  [discovery-peer-dbs] rotation fetch … failed
                      "cannot reach the blob's origin: No addressing information available"
```

The 09:21:14 respawn was **accidental**, and the 3.7-second gap to the rotation warning is the tell. The hourly rotation pass called `fetchPeer` → `discoveryP2p.fetch()` → `await start()`. `start()` spawns; it does not join. From that moment the server had a live sidecar that was not on the catalog topic, and the mesh-health watch skipped it because its guard read `if (!s.joined || s.neighbors > 0) return`.

Supporting details:

- **The identity key is fine.** Every `ready` line across two weeks reports the same `endpointId`, including across the 08-16 and 08-24 crashes and the 08-20 binary upgrade. Persistence at `{dbDirectory}/discovery-p2p/identity.key` works as designed — I initially suspected identity churn and it was a dead end.
- **This had happened before.** `08-16 08:36:45` SIGKILL → `08:46:02` bare respawn, same shape. It was masked by an unrelated full reboot at `16:59:51` that same day.
- **Zero `[discovery-seeds]` lines exist in all 337 files.** That is *consistent with* the guard bailing, but it is not independent proof — a healthy server logs nothing from that module either. The real proof is the reproduction below.
- **Both kills were `signal=SIGKILL, code=null`** — uncatchable, from outside the process, on a Docker install (`/app/mstream`). The container memory limit is the prime suspect; snapshot fetches are the memory-hungry moments. **This root cause is still uninvestigated** and is not addressed by this PR — the fix makes the server *survive* it, not avoid it.

### 3. Reproducing the original bug

Against the pre-fix code, on a server with `discoveryP2p.enabled`:

1. `curl :PORT/api/v1/admin/discovery/p2p/status` → confirm `running:true, joined:true`.
2. `ps -axo pid=,args= | grep p2p-sidecar` → `kill -9 <pid>`.
3. Status now reports `running:false, joined:false`, config still `enabled:true`. Nothing recovers it — I watched ~10 minutes, spanning two mesh-health ticks and one reconcile pass.
4. Trigger any lazy-start path, e.g. `POST /api/v1/admin/discovery/p2p/peer-dbs/fetch {"endpointId":"<a catalog peer>"}`. This is what rotation did.
5. Status now reports **`running:true, joined:false, neighbors:0`** — the exact state the reported server sat in for 6h45m.
6. `POST /api/v1/admin/discovery/p2p/enabled {"enabled":true}` returns `200 {"enabled":true,"collectForced":false}` and changes nothing.

### 4. Operator workaround for an already-affected server

**Restart the server.** It is the only action that replays the full stack.

A live repair without restart is `POST /api/v1/admin/discovery/p2p/join` with a community seed ticket — I measured that restoring `joined:true` with 3 neighbours in 8 seconds. But it does **not** restore the announce payload, which died with the old process, so the server would receive peers while remaining invisible to them until it re-announces. Half a fix; prefer the restart.

---

### 5. Open items and judgement calls for the next agent

**Things I chose not to do:**

1. **A boot-time start failure still disables the feature for the session.** `server.js` catches it, logs `catalog unavailable — feature disabled this boot`, and leaves it off. Recovery does not cover this: the handler is registered inside the start body, but `running` never becomes true, so `scheduleRecovery` no-ops by design. Pre-existing behaviour, deliberately unchanged — but arguably a boot that failed on a transient (a slow binary download, a busy data dir) should retry too. Worth a decision.
2. **`isStackRunning()` is exported and never used** anywhere in `src/`, `test/`, or `webapp/`. Dead export, noticed in passing. Left alone to keep the diff honest.
3. **Recovery logs one `warn` per attempt, forever.** At the 5-minute ceiling that is ~12 lines/hour indefinitely for a permanently broken sidecar. I judged a loud, bounded loop better than a silent one — that silence is the whole bug — but it is a deliberate tradeoff, not an oversight.
4. **`setUnexpectedExitHandler` is a single slot, not an emitter.** Fine for one consumer (the stack). If a second consumer ever needs the signal, it needs revisiting. I chose a private slot over the existing `events` emitter on purpose: that emitter re-emits whatever the child names in its `event` field, so routing recovery through it would let a buggy or hostile sidecar synthesize its own resurrection.

**Two follow-ups already scoped, out of scope here:**

5. **Hourly rotation churn.** 333 warnings since 08-10, alternating between the same two unreachable peers. `FETCH_BACKOFF_MAX_MS` is 60 min and `ROTATE_INTERVAL_MS` is 60 min, so the per-peer backoff has always just expired when rotation runs — the same dead candidates get picked forever. Mostly noise, but it was the *only* discovery-p2p line these logs produced, which is why a 6h45m outage looked like business as usual.
6. **The admin panel renders a dead sidecar as healthy-ish.** `meshSearching` in `webapp/admin/index.js` is true whenever `!running || !joined || neighbors === 0`, so a crashed sidecar shows "not joined yet" under a *"searching for peers"* progress bar — indistinguishable from normal startup, indefinitely. With this PR's recovery loop there is now real state worth surfacing ("reconnecting, attempt N"), which would need a small addition to the status route.

**Note on the third defect.** The EPIPE crash was found *by writing the test*, not by reading the code — the first integration run killed the server outright. If you touch `rpc()` or the spawn path, keep the negative control in mind: remove the `child.stdin.on('error')` handler and all three integration tests fail with the server dead. That is the fastest way to confirm the handler still matters after a refactor.

IrosTheBeggar:
## Root cause found: the SIGKILLs are OOM kills from a connection leak in iroh-gossip 0.101.0

Follow-up to the [handoff](https://github.com/IrosTheBeggar/mStream/pull/880#issuecomment-5398359448)'s open item — *"Both kills were `signal=SIGKILL, code=null` … This root cause is still uninvestigated."* It is now investigated, reproduced, and fixed-in-validation.

### The tell in the logs

Both kills landed at almost exactly the same sidecar **uptime**: 101.8 h (08-12 boot → 08-16 kill) and 103.8 h (08-20 boot → 08-24 kill). The one sidecar never killed was replaced by an operator reboot at 80.4 h — it just never reached the line. A constant-rate leak hitting a fixed ceiling produces exactly this: the sidecar's RSS climbs ~2–3 MB/h until it out-grows the Node server, and the container's OOM killer picks the fattest process. The 08-16 kill following the 5×snapshot-fetch burst by 8 minutes (and arriving 2 h *earlier* on the uptime clock) fits the same line plus a step.

### The mechanism (iroh-gossip 0.101.0 — the pinned version, and the newest released)

1. Gossip housekeeping constantly drops peer links **between two live processes**: hyparview answers any message from a non-active-view peer with `DisconnectPeer` (shuffles run every 60 s; the line carries an upstream `TODO … Maybe remove?`), and `accept_conn` retires the older of two duplicate connections (simultaneous dials, reconnects).
2. Dropping the link only drops the connection's send channel. The QUIC connection is closed later, by cleanup that runs **when the connection task finishes**.
3. The bug: `SendLoop::run`'s `select!` arm `Some(msg) = self.send_rx.recv()` silently **disables** on channel-close instead of breaking, so the send half hangs forever on `closed`; `connection_loop` uses `tokio::join!`, so the task never finishes and the close-cleanup never runs.
4. iroh 1.0.3 enables QUIC keep-alive by default → between two live endpoints the zombie never idles out. (Dead peers *are* reaped by the idle timeout — which is why the leak never shows in a stable or kill-a-peer test, only under real-world link churn.)

Upstream knows: [iroh-gossip#145](https://github.com/n0-computer/iroh-gossip/issues/145) describes it, [PR #154](https://github.com/n0-computer/iroh-gossip/pull/154) fixes it (respin of #146 with maintainer feedback), nothing merged, no release since 0.101.0.

### Reproduction (A/B, local, stock binary vs stock+PR154)

10-node mesh, announces every 5 s, instrumented actor reporting its live connection-task count (a healthy node needs ≈1 per neighbor):

| minute | stock 0.101.0 (mean tasks/node) | + PR #154 |
|---|---|---|
| 1 | 21.6 | 5.8 |
| 7 | 27.0 | 5.2 |
| 25 | still climbing | flat, = neighbor count |

Per-zombie cost measured at ~100–200 KB from the RSS gap between the runs. Ruled out with dedicated soaks: the hourly failed rotation fetches (86 reps, flat RSS), idle endpoint/relay/store-GC (flat), and raw message volume (plumtree caches saturate at their retention windows and hold — the unique-bytes heartbeat design is fine).

**mStream's full p2p integration surface passes against a PR154-patched sidecar build** — the existing `discovery-p2p.test.mjs` suite plus this PR's crash-recovery suite, real sidecars included.

### Plan

- **Phase 0 — merge this PR as-is.** It's the severity fix and nothing above changes a line of it. Until Phase 1 ships, each burn-down becomes a seconds-long self-healed blip; a restart resets the ~4-day clock.
- **Phase 1 — sidecar v1.0.2:** fork iroh-gossip at 0.101.0 + cherry-pick PR #154, pin by tag, tag v1.0.2, roll the sha256 manifests via `scripts/update-p2p-sidecar-manifest.mjs`. Comment the reproduction upstream; re-pin to crates.io when a fixed release exists.
- **Phase 2 — belt and suspenders (mStream, independent of Phase 1):** sidecar RSS watchdog in the 5-minute mesh-health watch (`/proc/<pid>/statm`, `ps` fallback; config `discoveryP2p.sidecarMaxRssMb` ~256 MB) → planned stack replay instead of a container OOM that could just as easily pick *Node* next time; surface sidecar RSS + recovery state in the status route/admin panel.
- **Phase 3 — hygiene:** rotation backoff cap vs the 60-min rotate interval (the 333-warning noise that buried this outage), the sidecar's unbounded per-origin `last_accepted` map, and the `/api/v1/ping` 500s that started with the 08-20 boot (stale-token client → should be 401, not 500).

### Pre-merge once-over of this PR's code

Reviewed all five changed files plus both test suites before merging. The recovery ladder's intent semantics hold up (a deliberate stop cancels recovery; the timer pre-clears `running` so a bare lazy respawn gets *upgraded* to a full replay rather than blocking it; timers are unref'd), the admin-route repair and the three pipe `'error'` handlers are correct, and the tests are the strongest part — negative controls that reproduce the legacy behavior, and a real-binary integration suite covering orphans, identity persistence, and the EPIPE burst. Verified locally on this branch: new unit suites exit 0, discovery p2p integration 65/65 (twice, real sidecars), all 11 CI build checks green.

One narrow race, noted rather than blocking: if an operator disables the feature while a recovery *replay attempt is already in flight* (the timer has fired, the start hasn't finished), the attempt fails against the teardown and its catch restores `running = true` and re-arms — which can resurrect the sidecar seconds after the disable. It needs a crash + an in-flight recovery + a concurrent disable, and a second disable cleanly stops it, so it's a hardening follow-up: gate the re-arm (and/or the timer body) on `config.program.discoveryP2p.enabled === true` so recovery can never outvote the config.

Full write-up with timeline, code anchors, chart, and the repro harness: *The 102-Hour Fuse* (artifact link on request).

PR #883, opened 2026-08-24
harden: sanitize child_process call in index.js...

anupamme:
## Summary
Harden input handling in `src/api/cli-audio/index.js` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | javascript.lang.security.detect-child-process.detect-child-process |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `javascript.lang.security.detect-child-process.detect-child-process` |
| **File** | `src/api/cli-audio/index.js:41` |
| **Assessment** | Defensive hardening |

**Description**: Detected calls to child_process from a function argument `binary`. This could lead to a command injection if the input is user controllable. Try to avoid calls to child_process, and if it is needed ensure user input is correctly sanitized or sandboxed. 

## Threat Model Context

This API endpoint appears to be publicly accessible. This is a web service - vulnerabilities in request handlers are directly exploitable by remote attackers.

## Changes
- `src/api/cli-audio/index.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/src/api/cli-audio/index.js
+++ b/src/api/cli-audio/index.js
@@ -31,7 +31,10 @@ export const PLAYERS = [
   { name: 'mplayer', kind: 'spawn',  binary: 'mplayer', probeArgs: ['-v'],        AdapterClass: MplayerAdapter, label: 'MPlayer' },
 ];
 
+const ALLOWED_BINARIES = new Set(PLAYERS.filter((p) => p.kind === 'spawn').map((p) => p.binary));
+
 function probeBinary(binary, args) {
+  if (!ALLOWED_BINARIES.has(binary)) { return Promise.resolve(false); }
   // Async probe: we used to call spawnSync here, but that blocks the Node
   // event loop, which starves the MPD TCP probe's 'data' callback and makes
   // its short timeout misfire. Using spawn lets every probe run concurrently.


IrosTheBeggar:
Closing: the semgrep hit is a false positive here. probeBinary() is module-private and its only caller passes player.binary from the hardcoded PLAYERS constant (mpv/vlc/mplayer) — nothing on that path comes from a request. The proposed allowlist is derived from that same constant, so the guard can never trip; it would only suggest to future readers that the argument might be attacker-controlled. Thanks for the report, but no change needed.

anupamme:
Thanks for taking the time to look into this. You’re right. I missed the fact that probeBinary() is only reached with binary values originating from the static PLAYERS configuration, so the value isn’t attacker-controlled along the current call path.

The allowlist I added doesn’t provide a meaningful security boundary given that data flow, so I agree that this scanner finding is a false positive rather than an exploitable command-injection issue.

PR #895, opened 2026-08-25
feat(admin): inline two-pane folder picker on the Add Folder screen

IrosTheBeggar:
Implements **Option B** from the [Add Directory design exploration](https://claude.ai/code/artifact/317af1c3-5daa-470f-8560-b93fedf4fffb), with the requested adjustments: no audiobooks checkbox, no "scans in the background" caption, the button reads **ADD FOLDER**, and the left pane is wider than the right (1.5 : 1).

## What changed

The Add Folder card becomes a two-pane picker:

- **Left — browse.** Clickable breadcrumbs (separator-preserving on Windows, with the drive selector), `.. up one level`, one row per subfolder. **Click selects** a folder; **open** (or double-click) descends into it; **[select this folder]** takes the directory currently being viewed. Backed by the same `POST /api/v1/admin/file-explorer` the old modal used — the modal component itself is untouched and keeps serving its other caller.
- **Right — the live summary.** The exact path being added (monospace), the shelf name auto-derived by the existing `makeVPath` rule shown as a chip with inline **rename**, the give-access-to-all checkbox (now `v-model` state instead of `getElementById`), and the ADD FOLDER button — disabled until a folder is picked and the name is a valid, unused slug, with inline error lines ("Letters, numbers and dashes only" / "That name is already in use") instead of a rejected submit.
- Selection is a path, not a row — it survives browsing elsewhere in the tree.
- The audiobooks checkbox is removed as dead UI; `isAudioBooks` stays optional server-side (Joi default false) and is simply omitted from the payload.

**No backend changes.** Same `PUT /api/v1/admin/directory`, same duplicate-vpath guard, same post-add bookkeeping (users' vpaths, folders map). The directories table below the card is untouched. The vPath jargon disappears from the first-run encounter without renaming anything server-side.

## Verified (live browser against a real server)

- Browse from home, jump to a deep path, descend via **open**, climb via `..` — breadcrumb wraps gracefully on absurdly deep paths.
- Click-select highlights the row, fills ADDING, and auto-derives the chip; rename to a custom slug via the inline input.
- ADD FOLDER: disabled state before selection (grey), enabled after; the add landed server-side — after a **full page reload** the table lists the renamed vpath pointing at the click-selected folder.
- Windows: drive selector + backslash-preserving breadcrumbs are code-pathed per the old modal's conventions (not smoke-tested on a real Windows box — flagging for the full-ci windows shards, though this is client-side rendering CI can't exercise either).

Deferred by design: the mock's per-library scan-status column — it needs a scan-progress endpoint that doesn't exist yet, same bucket as the design canvas notes.

### DIFF
--- a/webapp/admin/index.js
+++ b/webapp/admin/index.js
@@ -7738,7 +7738,7 @@ const discoveryView = Vue.component('discovery-view', {
                       <div v-on:click="headerTab = 'config'" :style="headerTabStyle('config')">Config</div>
                     </div>
                     <div v-if="headerTab === 'mesh'" style="display: flex; gap: 20px; align-items: center; min-height: 200px; flex-wrap: wrap;">
-                      <svg viewBox="0 0 300 210" width="300" height="210" style="flex-shrink: 0;">
+                      <svg viewBox="0 0 300 210" width="300" height="210" style="flex-shrink: 0; max-width: 100%; height: auto;">
                         <line v-for="n in meshMap.neighbors" :key="'l-' + n.id" x1="150" y1="105"
                           :x2="n.x" :y2="n.y" stroke="#4a7c59" stroke-width="2"></line>
                         <circle v-for="n in meshMap.outer" :key="'o-' + n.id" :cx="n.x" :cy="n.y" r="7"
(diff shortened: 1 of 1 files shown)


IrosTheBeggar:
Responsive smoke round (desktop 1280 / tablet 768 / mobile 375, mobile with touch emulation):

- **Found and fixed**: no wrap breakpoint — a 375px phone squeezed the two panes to ~139px each (truncated names, clipped footer link, one-word-per-line labels, button crushed to "ADD"). Fixed in c4cb833a with inline flex-basis: panes wrap and stack full-width on phones, stay side-by-side (left larger) from tablet up. No media query needed.
- **Verified after the fix**: zero horizontal overflow at all three sizes; tablet side-by-side; mobile stacked with the full touch path exercised — tap selects, tap *open* descends, tap *.. up* climbs, selection survives navigation, ADD FOLDER renders full width. Desktop unchanged at 1.5:1.
- Materialize's existing responsive shell (sidebar collapse, mobile nav button) behaves as before — untouched.

IrosTheBeggar:
Second responsive round — the **Discovery panel** this time (real server, two live peer sidecars, 375/768/1280 + touch):

Four mobile overflows found and fixed in 693cf143, all nastier than they looked because an ancestor clips — no scrollbar appeared, the content was just **unreachable** on a phone:

- Mesh map SVG fixed at 300px → scales down via max-width (viewBox keeps aspect), unchanged from tablet up
- Config grid's `minmax(280px, 1fr)` forced a 280px track in a ~272px card → `minmax(min(280px, 100%), 1fr)`; same for the peer-cards grid's 300px minimum
- List-mode peer table protruded **192px** — Downloaded + action columns were cut off entirely → wrapped in `overflow-x: auto` (swipes on phones, invisible on wider screens)

Re-verified at 375px: all five header tabs and both list modes render inside the card with zero overflow, the tab row wraps into tappable pills, and the table scrolls in its wrapper. Tablet/desktop measured unchanged (SVG 300, stats 3-up, no table scrollbar). Mobile screenshots show the mesh map + cards genuinely working at phone size.

PR #897, opened 2026-08-25
Boot watchdog: roll back updates that crash at boot, hold them, self-heal on the next release

IrosTheBeggar:
## Why

Groundwork for making **apply-updates-automatically the default**. The audit of the auto-update path found one systemic gap: recovery from a bad release depends on the *new* server being healthy enough to run the update checker. A release that passes the installers' exec probe (`-V` answers) but crashes during a real boot — config/db/module-load regression — was a dead install:

- `current`, the login item, and the macOS `~/Applications` copy already point at the broken version (the stage-time flip), so every restart and reboot lands on it again.
- The launcher never restarts a crashed server, and with the server dead there is no checker — **a fixed follow-up release could never arrive on its own.**

## What

**Launcher boot watchdog** (`rust-launcher/src/rollback.rs`, wired in `tray_app.rs`): a crash-before-serving server gets one retry, then the watchdog

1. confirms the managed layout is committed to this launcher's own version (build stamp vs `current`), with no `--server-bin`/`MSTREAM_SERVER_BIN` override in play;
2. picks the newest lower same-key version on disk (prune always keeps one), exec-probes it, and skips known-bad versions;
3. records the failed version in **`update-hold.json`** (beside the status file), deletes the stale armed `update-status.json` (a stale `applyRequested` read from a Stopped phase is how a crash used to turn into a relaunch loop), re-points `current` (atomic symlink swap on unix, junction re-create on Windows), restores `~/Applications` on macOS via ditto from the versioned root copy;
4. hands off to the previous launcher through the existing verified takeover relaunch. The login item re-points for free (`ensure_default_on` re-asserts with the new exe path). A server that cannot even *spawn* takes the same path.

**Server side** (`src/util/update-check.js`): a held version is never staged or applied; a `current` link still aiming at one is re-pointed at the running version (`enforceHold` — belt to the launcher's suspenders); holds are pruned once a version ≥ them actually boots. So the headline flow needs no human: **bad release boots → watchdog rolls back → hold blocks re-staging → the fixed release stages and applies normally.** `clearHold: true` on the update-settings endpoint is the operator override, surfaced on the About card as "clear hold & retry".

**Guard rails**: rollback chains are bounded (targets are strictly lower versions, held versions are skipped, at most the 2–3 versions prune keeps), and the launcher's auto-apply poll now only acts from a live-server phase and never into a held version.

## Testing

- 6 new launcher unit tests (candidate selection matrix, hold dedupe/cap/tolerant parsing, marker-root `~/Applications` shape, execute effects) — 31/31 pass, clippy clean.
- 2 new integration tests (hold honored + superseded by a newer release + `clearHold`; `enforceHold` re-pointing) — full unit (563) and integration (688) suites pass.
- New live smoke `npm run test:smoke:update-watchdog` (`test/smoke/update-watchdog-smoke.sh`): fabricates a managed root whose `current` crashes at boot with the **real launcher binary** in both bundles, and asserts the retry, the re-point, the recorded hold, the stale-status deletion, and the takeover into the previous version. Scratch HOME — never touches real data or login items.

## Out of scope / follow-ups before flipping the default

- R2: headless guards (no exit-0 in auto without a detectable supervisor; boot-attempt counter in `cli-boot-wrapper.js`).
- R3: deeper pre-flip probe (`--boot-probe`).
- R4: CI coverage of the apply leg (auto-arming, takeover cycle).
- Inno/pkg installs have no versioned rollback material; the watchdog rightly declines there (unchanged behavior).

### DIFF
--- a/.github/workflows/build-rust-launcher.yml
+++ b/.github/workflows/build-rust-launcher.yml
@@ -1,8 +1,8 @@
 name: Build Rust Launcher
 
 # Two duties, mirroring build-rust-parser.yml:
-#   - PR check (build + clippy on all three desktop OSes) so the crate can't
-#     bit-rot between phases.
+#   - PR check (build + test + clippy on all three desktop OSes) so the crate
+#     can't bit-rot between phases.
 #   - On master pushes touching the crate: build the four shipped launcher
 #     binaries, self-test each, and COMMIT them to bin/rust-launcher/ — the
 #     bundler (scripts/build-bun.mjs) stages the launcher from there as the
@@ -98,6 +98,18 @@ jobs:
       - name: Build (release)
         working-directory: rust-launcher
         run: cargo build --release --locked
+
+      # Per-OS on purpose, and --release so the build step's artifacts are
+      # reused: the boot-watchdog rollback module (rollback.rs) is cfg'd per
+      # target — junction re-point via mklink on Windows, atomic symlink swap
+      # on unix, the ~/Applications shape on macOS — so a test run on one
+      # platform proves little about the others. Before this step NOTHING in
+      # CI ran the launcher's tests at all: the windows junction arm had
+      # never executed anywhere, and a linux-only dead-code error shipped to
+      # a PR unseen by a mac-side cargo test.
+      - name: Test
+        working-directory: rust-launcher
+        run: cargo test --release --locked
 
       - name: Clippy
         working-directory: rust-launcher
--- a/docs/install.md
+++ b/docs/install.md
@@ -72,6 +72,18 @@ stage into layouts the installer owns: package-manager installs (deb/rpm,
 the macOS `.pkg`), Docker, npm, and hand-extracted copies are told about
 updates but never touched.
 
+If an applied update crashes before it ever serves, the desktop launcher's
+**boot watchdog** rolls it back on its own: after a failed retry it
+re-points `current` at the previous version (kept on disk for exactly this),
+restores the `~/Applications` copy on macOS, relaunches, and records the
+failed version in `update-hold.json` so the daily check doesn't re-stage it.
+The admin panel shows the held version; the hold clears automatically the
+moment a release newer than it ships (or by hand via the panel's
+"clear hold & retry"). Headless installs get the server-side half of this —
+a held version is never staged or applied, and a `current` link left on one
+is re-pointed at the running version — but with no launcher watching the
+boot, the rollback itself needs the supervisor or a re-run of the installer.
+
 Rolling back? Re-run the installer with `MSTREAM_VERSION=<old tag>`, restart,
 and then **skip the bad release** (the admin panel's skip link, or
 `updates.skipVersion` in the config) — otherwise the next daily check
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "test:smoke:docker": "node test/smoke/docker/run-docker-stack-smoke.mjs",
     "test:smoke:windows": "node test/smoke/windows-native-daemons.mjs",
     "test:smoke:update-watchdog": "sh test/smoke/update-watchdog-smoke.sh",
+    "test:smoke:update-watchdog:win": "powershell -NoProfile -ExecutionPolicy Bypass -File test/smoke/update-watchdog-smoke.ps1",
     "fetch-p2p-sidecar": "node scripts/fetch-p2p-sidecar.mjs",
     "fetch-mstream-player": "node scripts/fetch-mstream-player.mjs",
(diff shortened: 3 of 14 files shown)


IrosTheBeggar:
## Windows validation status (handoff notes)

**Verified so far:**

| Surface | macOS | Linux | Windows |
| --- | --- | --- | --- |
| Launcher unit tests (31, incl. 6 rollback) | ✅ local | ✅ container (glibc, flat layout) | ⏳ CI — this PR adds a `cargo test` step to the `check` matrix; the windows-latest leg is its first-ever run, and it executes the junction re-point (`remove_dir` + `mklink /J` + canonicalize verify) via the `execute_repoints_current...` test |
| Process-level watchdog smoke (crash → retry → rollback → hold → takeover) | ✅ `update-watchdog-smoke.sh`, 5/5 | covered by CI unit tests only | ❌ **not yet run — this is the ask below** |
| Full loop with the real server (bad 9.9.9 staged+auto-applied → crash → rollback → hold refuses re-stage → fixed 9.9.10 staged+applied → healthy) | ✅ real bundle, exit 0 | — | — (not required: same server-side code, platform-neutral, and the launcher legs are what differ) |
| JS suites (563 unit / 688 integration, incl. new hold tests) | ✅ | ✅ CI | n/a — the update integration suite self-skips on win32 (posix staging harness) |

**What still needs a real Windows machine:** one process-level run of the watchdog on the install.ps1 managed layout (junction `current`, flat bundles). The Windows-specific code is the junction re-point and the `DETACHED_PROCESS` takeover under a live lock handoff — unit-covered on windows-latest as of this PR, but never run as real processes.

**For the Windows agent:**

1. Check out this branch. Confirm the `check (windows-latest)` leg of *Build Rust Launcher* is green on the latest commit (that's the junction unit coverage).
2. Build the launcher: `cd rust-launcher; cargo build --release`
3. Run the smoke (needs `rustc` on PATH — it compiles two stub servers on the spot; prefer an interactive session):
   ```powershell
   npm run test:smoke:update-watchdog:win
   ```
   or directly: `powershell -NoProfile -ExecutionPolicy Bypass -File test\smoke\update-watchdog-smoke.ps1`
4. Expected: 5 `PASS` lines and exit 0 —
   - `PASS current re-pointed at 0.0.1`
   - `PASS hold recorded for 6.23.0`
   - `PASS stale update-status.json deleted`
   - `PASS a takeover launcher is running from the rolled-back layout`
   - `PASS the failed-version launcher exited`
5. On any `FAIL`: the script prints `launcher.log`; please attach it plus `update-hold.json` from the scratch dir it names, and note whether a message box appeared (a box = the watchdog declined or the rollback errored — the log line right before it says which).

Heads-up: `update-watchdog-smoke.ps1` was written on macOS against install.ps1's conventions (PS 5.1, ASCII, junction via `New-Item -ItemType Junction`) and has **not itself been executed on Windows** — if the harness rather than the launcher misbehaves (e.g. the `\\?\`-prefix comparison or process matching), fix the script in place; the launcher behavior under test is the 5 assertions above.

IrosTheBeggar:
**Windows smoke: 5/5 PASS** (real machine, Windows 10, interactive session — per the handoff checklist).

1. `check (windows-latest)` on the latest commit: **success** (all three OSes green).
2. Launcher built from the branch (`cargo build --release`, 1.94.1).
3. `update-watchdog-smoke.ps1` run as real processes:

```
PASS current re-pointed at 0.0.1
PASS hold recorded for 6.23.0
PASS stale update-status.json deleted
PASS a takeover launcher is running from the rolled-back layout
PASS the failed-version launcher exited
```

`launcher.log` narrates the exact designed sequence: spawn → crash → `(attempt 1/2) - retrying` → crash → *"mStream 6.23.0 cannot boot here - rolling back to 0.0.1"* → `current re-pointed at \?\...mStream-0.0.1-win-x64` (the junction remove_dir + mklink + canonicalize path, executed for real) → lock-handoff wait → *"handed off to the previous version"*. **No message box** — the decline/error paths never fired.

**Two harness-only fixes pushed to the branch** (as invited — the launcher needed zero changes):
- `rustc` derives the crate name from the file name, and `stub-6.23.0.rs` → `stub_6.23.0` is an invalid crate name (dots). Now passes `--crate-name` explicitly.
- `2>&1 | Out-Null` on a native command under `ErrorActionPreference='Stop'` turns any stderr line into a terminating NativeCommandError on PS 5.1 — stderr now goes to a file (kept as the failure hint).

The Windows column of the validation table is complete: junction re-point and DETACHED_PROCESS takeover under a live lock handoff, proven as real processes.

PR #899, opened 2026-08-25
chore(admin): remove dead user modals; Discovery header to light mode

IrosTheBeggar:
Follow-up cleanup to #896, same shape as #891: the master-detail Users page edits libraries and permissions inline, so the old per-user modals are dead code — nothing sets `currentViewModal` to `user-vpaths-modal` or `user-access-modal` anymore.

- Deletes both component definitions (`usersVpathsView`, `userAccessView`) and their `modVM` registrations — **161 lines, pure removal**.
- The `user-password-modal` and `lastfm-modal` stay; both are still referenced.

Verified by grep (zero remaining references to either component or name) and a live browser smoke: the Users detail panel renders, the library editor unlocks, and the password modal opens with its field — no console errors.


IrosTheBeggar:
Second commit (17f67579): the Discovery header card converts to **light mode** — bordered white section, the old panel's state colors (green connected / red reconnecting / orange waiting), default blue links, light tab pills keeping the dark active state, light stat tiles and config rows. The one deliberate exception: **the mesh map keeps a dark rounded inset** — its palette was drawn for a dark ground and now reads like a map tile against the light card, with the legend beside it in light ink.

Verified live against a server with two real mesh peers: all five tabs render on the light card; the dark inset holds the populated map (green linked neighbor, dashed offline dot). Screenshots of mesh + config tabs in the session record.

PR #900, opened 2026-08-25
Deep pre-flip probe: refuse releases that exec but cannot boot

IrosTheBeggar:
R3 of the auto-update audit (after #897's launcher watchdog and #898's headless guards) — the third of the four blockers before flipping `updates.mode` to `auto` by default.

## Why

The installers exec-probe a freshly extracted bundle with `-V` before flipping `$ROOT/current`. That proves the binary *runs*, not that it can *boot* — a broken module graph, a config the new build's schema refuses, or a database from the build's future all slip through and get caught only **after** the flip, by the R1/R2 watchdogs (recover-by-rollback). Stage-time refusal is strictly better where it's possible: `current` never moves, nothing needs recovering, the daily check retries, and the next release heals — the oldest, best-tested path in the feature.

## What

**`mstream-server --boot-probe`** ([boot-probe.js](../blob/claude/preflip-boot-probe-afecf4/src/util/boot-probe.js)), run by both installers right after `-V`:

1. **Module graph** — the same `import('./src/server.js')` the real boot performs: bundling regressions, native deps (sqlite driver, onnx, iroh), top-level module errors.
2. **Config** — the machine's *existing* config through the *new* build's Joi schema, validate-only (never generated, never written; a fresh install has nothing to check).
3. **Database, read-only** — real driver open plus `PRAGMA user_version` vs the build's `SCHEMA_VERSION`: an older build staged over a migrated db (manual rollback across a schema bump) is refused *before* it can leave scans refusing. Environmental problems (locked/unreadable db) are a note, not a failure — the probe judges the build, not the disk.

Zero side effects by contract (no config generation, no attempt counting — same purity rule as `-V`, pinned by test), self-bounded, and safe to run while the old server is live — which is exactly when managed staging runs it.

**Installer contract** (install.sh / install.ps1): exit 0 = pass; nonzero **with** a `boot-probe:` sentinel = refuse when a working install would be displaced (first installs get a diagnostic and proceed, same policy as `-V`); nonzero with **no** sentinel = a bundle that predates the flag → treated as a pass (`-V` vouched), so new installers keep staging old bundles — the documented manual-rollback flow keeps working, and the existing CI installer-contract stubs stay green by construction. install.ps1 captures the probe under a local `ErrorActionPreference='Continue'` window (the PS 5.1 native-stderr trap the Windows validation pass documented).

**Deliberate asymmetry, now documented in rollback.rs:** the watchdogs' rollback-target probe stays shallow (`-V`). A rollback target ran on this machine before, and post-crash the choice is a server-with-a-caveat vs no server — availability-first. Pre-flip, nothing is lost by being strict.

## Testing

- 3 new probe integration tests: pass-and-creates-nothing (purity), config parse/schema refusals with the sentinel, future-schema-db refusal + current-schema pass.
- 1 new staging-refusal test driving the real loop: server → `install.sh` → probe → refusal, `current` untouched — alongside the existing tamper/broken-binary refusal tests.
- Every branch also verified by hand against the **compiled Bun binary** (module graph from `$bunfs`, Bun sqlite adapter's `readOnly` path) — the shape the installers actually probe.
- Full suites green: 576 unit / 700 integration; `sh -n install.sh`; launcher `cargo test` 31/31 (comment-only change there).

## Out of scope

R4 (apply-leg CI smoke) — last audit item before the default flip.

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -218,6 +218,28 @@ jobs:
           try { & .\install.ps1 *>&1 | Out-Null; throw "installed a bundle that cannot exec" } catch { if ("$_" -notmatch 'does not run on this system') { throw } }
           $j2 = Get-Item "$env:RUNNER_TEMP\app\current" -Force
           if (($j2.Target | Select-Object -First 1) -notmatch 'mStream-9\.9\.9-win-x64$') { throw "current moved off 9.9.9: $($j2.Target)" }
+          # the DEEP probe: a bundle that EXECS (-V answers, so the shallow
+          # probe passes) but reports it would not BOOT must be refused on
+          # the "boot-probe:" sentinel with current untouched. Needs a real
+          # exe — rustc is preinstalled on the runner; stderr to a file, not
+          # 2>&1 (under EAP=Stop native stderr becomes a terminating error).
+          $pb = 'mStream-9.9.11-win-x64'
+          Set-Content stub.rs 'fn main(){let a=std::env::args().nth(1);if a.as_deref()==Some("-V"){println!("9.9.11");return;}if a.as_deref()==Some("--boot-probe"){println!("boot-probe: FAIL simulated boot regression");std::process::exit(1);}}'
+          rustc -O --crate-name probestub stub.rs -o probe-stub.exe 2>rustc.err
+          if ($LASTEXITCODE -ne 0) { Get-Content rustc.err; throw "rustc stub build failed" }
+          Push-Location rel
+          New-Item -ItemType Directory -Force $pb | Out-Null
+          Copy-Item ..\probe-stub.exe "$pb\mstream-server.exe"
+          Set-Content "$pb\README.txt" 'probe'
+          python3 -m zipfile -c "$pb.zip" $pb
+          Remove-Item -Recurse -Force $pb
+          Pop-Location
+          Remove-Item rel\mStream-*.zip -Exclude "$pb.zip" -Force
+          Remove-Item rel\manifest.json -Force
+          bash scripts/release-manifest.sh 9.9.11 rel
+          try { & .\install.ps1 *>&1 | Out-Null; throw "installed a bundle that cannot boot" } catch { if ("$_" -notmatch 'would not BOOT') { throw } }
+          $j3 = Get-Item "$env:RUNNER_TEMP\app\current" -Force
+          if (($j3.Target | Select-Object -First 1) -notmatch 'mStream-9\.9\.9-win-x64$') { throw "current moved off 9.9.9 after probe refusal: $($j3.Target)" }
           # uninstall
           $env:MSTREAM_UNINSTALL = '1'
           & .\install.ps1 *>&1 | Tee-Object -Variable un
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -246,6 +246,12 @@ jobs:
           if (Test-Path "$env:RUNNER_TEMP\app") { throw "app root survived uninstall" }
           if (($un | Out-String) -notmatch 'left in place') { throw "no data-home note" }
           Write-Host "install.ps1 contract OK"
+          # The runner appends `exit $LASTEXITCODE` to pwsh steps, and the
+          # deep-probe stub's --boot-probe exit 1 is the last NATIVE exit
+          # code set (script invocations and cmdlets never reset it) — so a
+          # fully-passing run would end red. Every check above is a throw;
+          # reaching here IS the pass.
+          exit 0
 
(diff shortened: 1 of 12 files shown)


IrosTheBeggar:
## Windows validation status (handoff notes)

**What CI now covers on Windows for this PR:**
- The probe's own integration tests run un-skipped on the windows test shards (wrapper child processes under Node): purity, config refusals, db-schema refusal.
- The windows installer-contract leg now includes a **deep-probe refusal case** (latest commit): a rustc-stub `mstream-server.exe` that answers `-V` but fails `--boot-probe` with the sentinel — install.ps1 must refuse with `would not BOOT` and leave `current` on the junction it started on. (Before that commit, the leg's fake exe failed `-V`, so the new probe code never actually executed there.)
- The win-x64 bundle leg proves the compiled binary boots (superset of the probe's module-graph check); the probe's branches against a compiled binary were verified on macOS/Bun.

**What CI structurally cannot cover — the ask for the Windows machine:** GitHub runs these steps under **pwsh 7**, but the real `irm | iex` floor is **Windows PowerShell 5.1**, and the probe capture in install.ps1 uses a local `ErrorActionPreference='Continue'` window specifically for 5.1's native-stderr semantics — the exact failure class the last 5.1 validation pass caught in the smoke harness. Two short checks, both explicitly under `powershell.exe` (5.1), not pwsh:

1. **Refusal branch under 5.1** — rebuild the contract shape by hand (or adapt the snippet from `.github/workflows/build-bun.yml`, "the DEEP probe" section): a working fake install at `current`, then publish a rustc-stub bundle that answers `-V` and prints `boot-probe: FAIL simulated boot regression` / exit 1 on `--boot-probe`. Run `install.ps1` (env: `MSTREAM_RELEASE_BASE` at a local file server, `MSTREAM_INSTALL_DIR` scratch, `MSTREAM_NO_DESKTOP=1`, `MSTREAM_NO_PATH=1`). Expect: a thrown `would not BOOT (...)` including the sentinel text, nonzero exit, and `current` unchanged.
2. **Pass path under 5.1** — same layout, stub exits 0 on `--boot-probe`: the install must proceed normally with no stray error records surfacing from the probe's stderr capture.

On any surprise: the thrown message (or lack of one) plus the raw `$probeText` value are the interesting artifacts; the capture window is at the "DEEP probe" comment block in install.ps1 — fix the harness/installer in place as before if it's a 5.1 quirk rather than probe behavior.

IrosTheBeggar:
**Windows PowerShell 5.1 validation: ALL PASS** — both handoff checks, on real `Desktop 5.1.19041.6456` (not pwsh), real machine.

Harness per the handoff: rebuilt the contract shape by hand — probe-ok 9.9.9 rustc-stub installed as baseline (`current` junction), local feed via `release-manifest.sh` + loopback http, `MSTREAM_INSTALL_DIR` scratch, `MSTREAM_NO_DESKTOP/NO_PATH`.

**1. Refusal branch under 5.1** — published 9.9.11 (`-V` ok, `--boot-probe` → `boot-probe: FAIL simulated boot regression` / exit 1):
- ✅ thrown message matches `would not BOOT` **and carries the sentinel text verbatim**
- ✅ nonzero (terminating throw)
- ✅ `current` junction unchanged (still 9.9.9)

**2. Pass path under 5.1** — published 9.9.12 (`--boot-probe` → ok / exit 0):
- ✅ install proceeded; `current` flipped to 9.9.12
- ✅ **zero stray error records** (`$Error` empty of NativeCommandError/boot-probe entries; none in the combined output) — the local `EAP='Continue'` capture window holds under real 5.1.

**Corroborating war story from building the harness** — the exact failure class the capture window guards against bit my *harness* twice before the probe ever ran: (a) bare `bash` from PowerShell resolves to WSL's `System32\bash.exe`, and (b) `release-manifest.sh` logs its *success* line to stderr, which under `EAP=Stop` 5.1 raised a terminating `NativeCommandError` **even through a `2>` file redirect** — fixed only by the same local `EAP='Continue'` window install.ps1 uses. So the design decision under test here is demonstrably load-bearing, not defensive theory: without that window, any stderr line from the probe (a warning, a deprecation notice) would kill the install on every 5.1 machine.

No installer or probe changes needed — the shipped code behaved exactly as designed under 5.1. Harness was scratch-local (happy to PR it as `test/smoke/boot-probe-51-smoke.ps1` if you want it kept).

IrosTheBeggar:
Harness pushed as requested: `test/smoke/boot-probe-51-smoke.ps1` + `npm run test:smoke:boot-probe:win51` (invokes `powershell.exe` explicitly, so it's always the 5.1 floor; a pwsh run warns that it isn't). Cleaned up from the hand-rolled version: derived paths, git-bash located via git.exe, Store-stub python probe, env save/restore. Verified green through the npm entry point on this machine before pushing — same 4 PASS lines as the validation report above.

PR #909, opened 2026-08-25
Boot hold: serve a live diagnosis and self-heal instead of crash-looping on environmental failures

IrosTheBeggar:
## Why

The 2026-08-25 production incident: the disk holding `/config` filled, and a container restart put mStream into an infinite boot loop — banner + `Error: disk I/O error` (`SQLITE_IOERR_SHMSIZE`, errcode 4874: SQLite couldn't grow the WAL shared-memory file) repeating forever under s6, port dead, `docker ps` claiming "Up". Nothing anywhere said *the disk is full*.

## What

**Layer 1 — say what's actually wrong.** New `src/util/boot-errors.js` classifies boot-time DB failures: SQLite result codes (both runtimes) + plain fs codes, enriched with a `statfs` reading of the db directory. `disk I/O error` becomes *"mStream cannot start: the disk holding its database is full — /config/db, 0 B free of 98 GB"* with a concrete hint.

**Layer 2 — hold instead of crash.** For classified **environmental** failures on a **first boot** (`disk-full`, `db-io`, `db-damaged`, `db-unwritable`, `db-missing-dir`, `db-locked`, `fs-limit`), the server binds its configured port and answers everything **503**: a self-refreshing diagnosis page for browsers, JSON + `X-MStream-Boot-Hold: <reason>` for API callers. The full boot retries every 15s (`MSTREAM_BOOT_HOLD_RETRY_MS` to tune) and a successful retry swaps the real app onto the *same socket* via the existing keep-listener machinery — fix the disk, and the next page refresh is the app.

Deliberate boundaries, documented in place:
- **Unrecognized failures crash exactly as before** (a migration bug must stay loud, and the R2 boot-watchdog rollback still works).
- **Soft reboots keep their exit contract** (the previous life's timers must not outlive their closed db handle); the supervisor's single restart lands in the hold with the same diagnosis.
- A config failure *during a retry* re-enters the hold (config usually shares the volume with the db) instead of exiting.

## Hardening from the adversarial review (8 finder angles, findings verified then fixed)

- **`logger.js`**: a winston transport write error (ENOSPC on the log file) was an unhandled `'error'` on the default logger → process death *while logging about the full disk*. Now listened for (empirically reproduced both ways), and `addFileLogger` is idempotent for same-dir re-runs.
- **`bun-sqlite-adapter.js`**: the constructor is now wrapped and `errno`→`errcode` mirrored, so open-time failures classify under Bun builds too.
- **`manager.close()`**: invalidates memoized user/library caches and tolerates a half-open handle — a damaged db replaced mid-hold is re-read for real, and the released handle means the file is replaceable on Windows.
- Watchdog counter is left alone (an earlier draft cleared it and would have defeated bad-build rollback), the scanner reap is skipped on retries (its synchronous process probes would stall the diagnosis page), hold logging is throttled (a weekend hold can't flood the admin live-log ring), and the hold reuses the listener `serveIt` already built, with a proper fatal error handler for the socket's post-recovery life.
- `test/helpers/server.mjs` `waitForReady` now reports a boot hold as a terminal failure with the server's own diagnosis instead of burning 90s and saying "unknown".

Known limitations (deliberate, noted in review): boot steps *after* `initDB` (lyrics/backup init writes) can still crash-fast if the disk is at the exact boundary where the WAL opens but the next write fails; the desktop tray launcher currently shows a held server as "Running (unverified)" — launcher-side hold awareness is queued as a follow-up.

## Tests

- `test/unit/boot-errors.test.mjs` — 15 classifier/rendering cases (incl. the real incident error shape, Bun's `errno` shape, non-holdable fallthroughs, HTML escaping).
- `test/integration/boot-hold.test.mjs` — end-to-end: damaged `mstream.db` → hold (JSON + HTML 503s, process survives multiple retry windows) → config broken mid-hold → diagnosis flips to `config` → restored → flips back → damaged file **deleted while held** (proves the handle release; would be EPERM on Windows otherwise) → app recovers on the same socket.
- `test/unit/logger-error-resilience.test.mjs` — a transport error must not kill the process.
- Full suite: 2,444 pass / 1 load-flake (dlna ECONNRESET while 8 review agents saturated the CPU; passes 20/20 in isolation — same class as the pre-existing federation-dial flake).

### DIFF
--- a/src/db/bun-sqlite-adapter.js
+++ b/src/db/bun-sqlite-adapter.js
@@ -13,12 +13,19 @@ import { Database } from 'bun:sqlite';
 // name ('SQLITE_ERROR', ...) or undefined (e.g. FTS5 MATCH parse errors).
 // Callers key on the node-style code — notably the FTS5->LIKE search fallback
 // in src/api/search.js and src/api/subsonic/handlers.js — so translate it here
-// so the shim is behaviourally indistinguishable from node:sqlite.
+// so the shim is behaviourally indistinguishable from node:sqlite. The
+// numeric result code matters too: node:sqlite carries it in `errcode`,
+// bun's SQLiteError in `errno` — mirror it across so consumers that map
+// codes (util/boot-errors.js classifying a boot failure as environmental)
+// see one shape on both runtimes.
 function withNodeErrors(fn) {
   try {
     return fn();
   } catch (err) {
-    if (err && err.name === 'SQLiteError') { err.code = 'ERR_SQLITE_ERROR'; }
+    if (err && err.name === 'SQLiteError') {
+      err.code = 'ERR_SQLITE_ERROR';
+      if (err.errcode === undefined && Number.isInteger(err.errno)) { err.errcode = err.errno; }
+    }
     throw err;
   }
 }
@@ -36,9 +43,13 @@ class StatementSync {
 export class DatabaseSync {
   #db;
   constructor(location, options = {}) {
-    this.#db = options.readOnly
+    // The open itself must be wrapped too: a bun:sqlite open failure
+    // (SQLITE_CANTOPEN on a vanished volume, SQLITE_READONLY on a bad
+    // mount) is exactly what the boot-failure classifier needs to see in
+    // node:sqlite's shape — unwrapped, those crashed unclassified.
+    this.#db = withNodeErrors(() => (options.readOnly
       ? new Database(location, { readonly: true })
-      : new Database(location, { create: true });
+      : new Database(location, { create: true })));
   }
   exec(sql) { return withNodeErrors(() => this.#db.exec(sql)); }
   prepare(sql) { return new StatementSync(withNodeErrors(() => this.#db.prepare(sql))); }
--- a/src/db/manager.js
+++ b/src/db/manager.js
@@ -201,8 +201,20 @@ export function transaction(fn) {
 
 export function close() {
   stopSharedCleanup();
+  // The caches memoize rows read through the handle being released; the next
+  // initDB() must re-read them from whatever file it opens. This matters for
+  // the boot hold (server.js): a failed initDB can populate some caches from
+  // a damaged database before throwing, the operator replaces the file, and
+  // the retry's initDB would otherwise serve users/libraries from the old
+  // bytes (loadUsersCache & co. early-return when already populated).
+  invalidateCache();
+  _anonymousUserId = null;
   if (db) {
-    db.close();
+    // A half-configured handle (constructor succeeded, a pragma threw — the
+    // boot-hold entry path) may refuse to close cleanly; that must not turn
+    // the release into a crash. Releasing also matters on Windows: an open
+    // handle blocks the operator from replacing a damaged mstream.db.
+    try { db.close(); } catch (err) { winston.warn(`Could not close the database handle: ${err.message}`); }
     db = null;
   }
 }
--- a/src/logger.js
+++ b/src/logger.js
@@ -171,6 +171,18 @@ winston.configure({
   exitOnError: false
 });
 
+// A transport write failure — ENOSPC on the log file when the disk fills
+// being the case that matters — is re-emitted as 'error' on the default
+// logger (winston Logger._onEvent), and an unlistened 'error' on a stream is
+// an uncaught exception: logging about a full disk would KILL the process,
+// which is exactly the moment the boot hold needs logging to stay soft
+// (server.js). The facade doesn't expose the default logger instance;
+// startTimer()'s Profiler carries a public reference to it. exitOnError
+// above does not cover this path — that flag is for the exception handler.
+winston.startTimer().logger.on('error', (err) => {
+  try { fs.writeSync(2, `[logger] transport error: ${err && err.message ? err.message : err}\n`); } catch { /* stderr gone too */ }
+});
+
 function dateKey() {
   const d = new Date();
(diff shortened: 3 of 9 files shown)


IrosTheBeggar:
## Independent smoke test of the shipped image + a verified matrix of the neighboring "disk I/O error" causes

Ran a full smoke sweep of `lscr.io/linuxserver/mstream:latest` (v6.23.0-ls21, built 2026-08-25T02:12Z) on Docker Desktop's Linux VM, then deliberately broke the storage under `/config` every way a real deployment does. Summary: **the release itself is healthy on normal storage**, and the "disk I/O error" family is reproducible on demand from several *different* environmental causes — all of which land in this PR's classifier. Details below, including two places where the lab results suggest sharpening the diagnosis text.

### What passes (baseline)

- Fresh install, arm64 **and** amd64: full migration chain to v66, web UI HTTP 200.
- Upgrade path: volume initialized by `v6.22.0-ls20`, then `:latest` on the same volume — v65 → v66 migration applies, boots clean.
- Bind-mounted `/config` (virtiofs) and even a **100%-full** `/config` with an up-to-date schema (nothing to write at boot ⇒ survives; matches the incident being migration/first-write-sensitive).

### Verified failure matrix (same image, only the filesystem under `/config` changes)

| Storage under the db dir | Boot error (verbatim) | errcode | Classifier bucket |
|---|---|---|---|
| FUSE without shared mmap (sshfs `-o direct_io`; ≈ unRAID shfs, mergerfs `cache.files=off`, rclone) | `disk I/O error` | **5386** = IOERR_SHMMAP | `db-io` |
| CIFS, default mount (no `nobrl`) | `database is locked` | 5 (plain BUSY) | `db-locked` |
| NFSv3 (no client-side statd) | `database is locked` | 5 | `db-locked` |
| db dir root-owned (PUID mismatch with `lsiown` bypassed) | `attempt to write a readonly database` | 1544 = READONLY_DIRECTORY | `db-unwritable` |

All four crash-loop identically under s6 on current master with `docker ps` showing "Up" — exactly the failure shape this PR replaces with a hold, and all four are first-boot environmental failures, so the hold applies to each. The disk-full incident (SHMSIZE 4874) and the SHMMAP row are siblings: same `SQLITE_IOERR` family, different syscall, and the `LOW_DISK_BYTES` split in `classifyBootError` is what correctly separates them. 👍

### Two refinements the lab results argue for

**1. The `db-locked` hint is misleading for the CIFS/NFS rows.** Both "database is locked" reproductions were **first boots on a fresh, empty volume — no second process existed anywhere**. Broken byte-range locking on those mounts makes SQLite report BUSY spuriously and *permanently*: the 15s retry can never succeed, while the page tells the operator to hunt for a second instance or stuck scanner that isn't there. Suggest the hint also name the real possibility: db directory on a CIFS/NFS mount whose locking doesn't work (CIFS fix: `nobrl` mount option, or move the db to local storage; NFS fix: working lockd/statd, or local storage).

**2. The fs type is already in hand — naming it makes both hints deterministic.** `freeSpace()` calls `fs.statfsSync`, and the returned object also carries `type` (the `f_type` magic): `0x6969` NFS, `0xFE534D42` SMB2, `0xFF534D42` legacy CIFS, `0x65735546` FUSE, `0x794C7630` overlayfs. Mapping those few magics would turn "usually a nearly-full disk, a failing drive, or a network/overlay filesystem" into *"…is on an NFS mount, which cannot support SQLite's WAL mode"* — and would let `db-locked` distinguish the genuinely-locked case from the broken-locking case with zero extra syscalls. (`/proc/mounts` gives the same answer with prettier names, Linux-only.)

### Validated escape hatches (follow-up material, beyond this PR's scope)

- **`journal_mode = DELETE` works where WAL throws SHMMAP**: verified read/write on the sshfs/`direct_io` mount. An automatic WAL→DELETE fallback at `initDB` would turn the FUSE-class permanent hold into a working (degraded) server — but it must propagate to every connection that sets WAL (`scanner.mjs`, rust-parser, `discovery-db.js`) per the existing connection-symmetry rule.
- **DELETE alone does *not* rescue CIFS/NFS** (still "database is locked"), but `node:sqlite` accepts URI filenames, and **`file:…?vfs=unix-dotfile` + DELETE verified working on both**. Opt-in only — a crash leaves a stale dot-lock that reads as "database is locked" until manually removed.

### Repro lab (for tests or future spelunking)

- **CIFS**: alpine+samba container, then `docker volume create --driver local --opt type=cifs --opt device=//$IP/config --opt "o=username=…,password=…,vers=3.0,uid=911,gid=911"` and mount it as `/config`.
- **NFS**: privileged alpine+nfs-utils exporting a **named volume** (overlayfs can't be NFS-exported), `mount -t nfsd nfsd /proc/fs/nfsd`, then a `type=nfs,o=addr=$IP,nfsvers=3` volume. Warning: a wedged kernel-nfsd container becomes unremovable ("did not receive an exit event") until the Docker VM restarts.
- **FUSE/SHMMAP (the exact `disk I/O error`)**: inside the lsio image itself — `apk add sshfs openssh`, enable root login (Alpine's locked root silently refuses key auth), `sshfs -o direct_io,allow_other root@localhost:/realcfg /fusecfg`, point `-j` at a config whose `dbDirectory` lives there.

One last observation, in this PR's favor: with the hold in place, the lsio `s6-notifyoncheck` port probe (`nc -z localhost 3000`) starts succeeding during a hold, so the container finishes init with the diagnosis page reachable on the mapped port — a strict upgrade over today's dead-port crash loop for every row in the table above.

PR #908, opened 2026-08-25
feat(first-run): tray "Set up mStream" item + headless boot invitation

IrosTheBeggar:
Wires the v0.4.0 setup wizard (companion to #907) into both first-run surfaces:

**Tray launcher** — new "Set up mStream" menu item (between Quick Connect and the autostart toggle) opens a real terminal running the bundled `mstream-player setup --server <this server>`. Per-OS seams mirror `open_logs_terminal`:
- macOS: executable `.command` via `open` (no AppleEvents consent prompt), with a CSI 8 resize to the wizard's designed 120×42
- Windows: `wt.exe` first (Windows Terminal renders the wizard's pixel art via sixel), falling back to a fresh conhost (`CREATE_NEW_CONSOLE`; the wizard degrades to half-blocks)
- Linux: the same emulator-candidate walk as View logs

The player binary is resolved once at boot — the copy build-bun stages next to the server binary first, then the server's runtime-fetched managed copy — and the item is greyed out when neither exists (musl installs have no player build).

**Server** — when a boot finds no folders **and** no accounts, the boot log invites setup: the browser line always, plus the exact runnable wizard command when a player binary is already present (`installedPlayerPath()` is a stat, never a download). A deliberately public server has folders, so it never nags after setup.

**Old-server compat note**: the shipped wizard treats scan-progress endpoints as garnish (`Err(_) => {}`) — against a server without `/api/v1/scan/status` it silently misses the final "scan complete" flip and nothing else; no error surfaces.

Verification:
- `cargo test`: 35 green, including new `player_key` (node-name lockstep) and `find_player_bin` (bundled-beats-managed) coverage; clippy clean
- node suite: 2390 tests, 0 failures (discovery-p2p files excluded — they fail identically on a pristine tree in this environment)
- live macOS smoke: fresh sandboxed server printed both invitation lines with the exact runnable command; `open_setup_terminal` drew the wizard in Terminal.app at 120×42 against that server; after configuring folders+admin through the wizard's own API surface and rebooting, neither line printed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #16, opened 2017-02-28
Add last.fm scrobbling


mateossh:

IrosTheBeggar:
I like this idea but  just don't have time to work on it in the foreseeable future.  Got a lot of other projects in the pipeline for now (like an Android App).  

I will start working on documenting mStream in case anyone else is interested in forking it and adding this

IrosTheBeggar:
I'm finally getting around to working on this.  I will hopefully have it done within a week

IrosTheBeggar:
LastFM scrobbling has been added to the master branch

You can set it up in the command line with the `-l` and `-z` flags 

```
mstream -l username -z password
```

PR #915, opened 2026-08-26
feat(tray): terminal-first tray — Quick Connect page, Open Admin Panel, first-run wizard auto-open

IrosTheBeggar:
The tray's **Quick Connect** item opened the web player's modal (`{url}/#quick-connect`). On macOS and Windows it now opens the wizard's standalone pairing page — `mstream-player qr`, the Done screen with the scannable **pixel QR** and app buttons — through the exact machinery Set up mStream uses: the bundled Ghostty console (mStream Dock icon, Terminal.app fallback) on macOS, `wt.exe` (conhost fallback) on Windows.

Mechanically, `open_setup_terminal` generalizes to `open_wizard_terminal(page)`: a `WizardPage` enum carries the subcommand (`setup`/`qr`), the window title ("mStream Setup" / "mStream Quick Connect" — matching the OSC titles the v0.5.0 player sets itself), and a **distinct** `.command` script name so the two tray items never clobber each other's launch files (pinned by test).

**Kept on purpose**: Linux stays on the webapp modal (consistent with the no-bundled-console posture), and every platform falls back to the modal when the install has no player binary or the terminal launch fails — with the reason in the launcher log.

Verification: cargo 37/37 (page mapping, both pages' ghostty configs, never-`-e` still pinned), clippy `-D warnings` clean on host + windows targets, and a live launch through the production code path on this machine — `opened via bundled Ghostty console` with the `qr` page running from the staged bundle against a sandbox server.

One behavior note for review: against a configured server, the `qr` page authenticates from the player's saved session (created when setup ran on that machine); without one it renders the page with a polite fetch-error note rather than a QR — same class of experience as the web modal asking for a login. Fresh servers (open admin) need no auth.

### DIFF
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1689,6 +1689,7 @@ jobs:
         run: |
           D=$(ls -d dist/stage/mStream-*-${{ matrix.key }})
           cp -R "$D" "$RUNNER_TEMP/inv"
+          rm -rf "$RUNNER_TEMP/inv/save"   # the copy must be DB-pristine too: the main smoke's library would trip the setupComplete backfill
           case "${{ matrix.key }}" in
             darwin-*) SRV="$RUNNER_TEMP/inv/mStream.app/Contents/MacOS/mstream-server" ;;
             win-*)    SRV="$RUNNER_TEMP/inv/mstream-server.exe" ;;
--- a/.github/workflows/build-bun.yml
+++ b/.github/workflows/build-bun.yml
@@ -1689,7 +1689,7 @@ jobs:
         run: |
           D=$(ls -d dist/stage/mStream-*-${{ matrix.key }})
           cp -R "$D" "$RUNNER_TEMP/inv"
-          rm -rf "$RUNNER_TEMP/inv/save"   # the copy must be DB-pristine too: the main smoke's library would trip the setupComplete backfill
+          rm -rf "$RUNNER_TEMP/inv/save" "$RUNNER_TEMP/inv/mStream.app/Contents/MacOS/save"   # DB-pristine on BOTH layouts: darwin's dataRoot lives inside the .app
           case "${{ matrix.key }}" in
             darwin-*) SRV="$RUNNER_TEMP/inv/mStream.app/Contents/MacOS/mstream-server" ;;
             win-*)    SRV="$RUNNER_TEMP/inv/mstream-server.exe" ;;
--- a/docs/install.md
+++ b/docs/install.md
@@ -139,7 +139,7 @@ directly, with `manifest.json` holding their sha256s:
 Windows, `mStream.app` on macOS, `mstream-desktop` on Linux — starts the
 server in the background, puts an mStream icon in your tray / menu bar
 (a status line — "Running · up 3h 12m", or Starting… / Stopped — then Open
-mStream · Quick Connect · Set up mStream · Start at login · View logs ·
+Admin Panel · Quick Connect · Set up mStream · Start at login · View logs ·
 Restart server · Quit), and opens your browser at the player. Start-at-login
 is on by default; one click in the tray menu turns it off. **Set up mStream**
 opens a terminal running the guided setup wizard (the bundled
--- a/docs/install.md
+++ b/docs/install.md
@@ -139,19 +139,17 @@ directly, with `manifest.json` holding their sha256s:
 Windows, `mStream.app` on macOS, `mstream-desktop` on Linux — starts the
 server in the background, puts an mStream icon in your tray / menu bar
 (a status line — "Running · up 3h 12m", or Starting… / Stopped — then Open
-Admin Panel · Quick Connect · Set up mStream · Start at login · View logs ·
+Admin Panel · Quick Connect · Start at login · View logs ·
 Restart server · Quit), and opens your browser at the player. Start-at-login
-is on by default; one click in the tray menu turns it off. **Set up mStream**
-opens a terminal running the guided setup wizard (the bundled
-`mstream-player setup`) — music folders, admin account, extras. It is
-one-time onboarding: on a first install it opens by itself, and once the
-server is set up the menu item greys out (the wizard itself declines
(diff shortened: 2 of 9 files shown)


IrosTheBeggar:
Scope grew by two operator requests after hands-on testing, same theme:

- **'Open mStream' → 'Open Admin Panel'** (opens `/admin` directly — the tray is the operator's surface; the post-boot announce keeps its own smart routing).
- **First install auto-opens the setup wizard**: the post-boot announce, when no folders are configured, opens the guided wizard in the console (mac) / terminal (win) instead of the browser admin panel — browser stays the fallback and the linux behavior; `--takeover`/`--autostarted`/`--no-open` suppress it exactly like the browser pop. Live-verified on a fresh sandbox boot.

Also found during that testing, tracked separately in the player repo: `setup --server` never consults the saved session (unlike `qr`), so reopening the wizard against a configured server shows 'could not reach the server' for what is actually a 401. Fix incoming as a player PR; harmless to this one.

IrosTheBeggar:
One more operator decision folded in: **the wizard is one-time onboarding**. The tray's 'Set up mStream' now enables only while the library is unconfigured (re-checked each minute tick, so finishing setup greys the item without a restart); first-install auto-open is unchanged. The server-truth backstop lives player-side — mstream-terminal-player#16 parks the wizard on configured servers with an *'already set up — use the admin panel'* message (which also fixes the reopen bug found in testing: a 401 was being reported as 'could not reach the server'). install.md updated to the one-time framing.

IrosTheBeggar:
Redesigned per the operator's setupComplete proposal — and the pre-flight behavior check it demanded caught two real defects beyond the design itself:

1. **The launcher's folder check was reading a ghost.** Folders moved to SQLite; wizard-configured installs never get config-file folders, so every configured modern install re-ran first-run behavior forever. The real install's own launcher.log confirmed it ('browser target is the admin panel' on every boot of a configured library).
2. **The health probe never marked authed servers as up.** `GET /` answers `302 → /login` once accounts exist; the probe demanded `200 + mStream`, so the tray sat on 'Starting…' — status line, announce, everything — for every password-protected install.

What shipped: the server owns a one-time `setupComplete` config marker (written at the first library or first user, boot-backfilled for pre-flag installs so upgraders never see first-run behavior); the launcher reads it (legacy-folders belt for old servers) for the first-install wizard auto-open and `browse_target` routing; the **'Set up mStream' tray item is removed entirely**; the probe accepts the authed-server shape. The wizard/player needs no knowledge of any of it — player #16 reduces to an optional standalone message fix.

Verified with the smoke that should have existed all along — a **non-wiped two-boot** run against the source server: boot 1 fresh → invitation + wizard auto-open via the bundled console + flag written at the first folder; boot 2 unwiped → up in 1s, no wizard, no invitation, announce at the player. cargo 38, node 2396, clippys clean.

IrosTheBeggar:
@windows-agent one-line workflow fix needed — the three build-leg smokes are red and the cause is the invitation smoke's fixture, not the feature.

**Diagnosis**: the step copies the staged bundle to dodge the main smoke's in-place taint, but the taint is INSIDE the copy: `save/` (the SQLite DB, which the main smoke populated with a library) rides along. The new boot backfill then — correctly, by design — marks a fresh-config-beside-existing-DB install as set up, so the 'not set up yet' line never prints and the assert fails. The local two-boot demo passed because its DB was genuinely fresh.

**Fix** — in `.github/workflows/build-bun.yml`, step `Smoke test first-run invitation (fresh boot)`, directly AFTER the line:

```
          cp -R "$D" "$RUNNER_TEMP/inv"
```

insert this one line (same indentation):

```bash
          rm -rf "$RUNNER_TEMP/inv/save"   # the copy must be DB-pristine too: the main smoke's library would trip the setupComplete backfill
```

Nothing else changes. Commit to `claude/tray-quickconnect-console`; suggested message: `ci(smoke): make the invitation fixture DB-pristine (the backfill is working as designed)`. If the push bounces on workflow scope, report here.

IrosTheBeggar:
Fixture fix implemented and pushed as `2b1b4df6` — one line after the `cp -R`, exactly as specified (the diagnosis is right, and as the author of that step on #908 I can confirm the copy was never meant to carry `save/`; the backfill just made the latent taint observable).

**CI verification is blocked by GitHub, not the branch**: Actions is in a [major outage](https://www.githubstatus.com/) right now — the in-flight run for `b4a970cd` had all jobs cancelled, and neither my fix push nor an empty-commit retrigger (`6fe5ede9`) produced a check-suite at all. I'm watching the incident; the moment Actions recovers I'll retrigger and report the three previously-red build-leg smokes here.

IrosTheBeggar:
@windows-agent one more line — my fixture wipe was linux-minded. The dispatch run proves the fix works (linux-x64 and win-x64 invitation smokes now green) but **darwin still fails**: its bundle's dataRoot is INSIDE the .app, so the tainted DB sits at `inv/mStream.app/Contents/MacOS/save` and survived the root-level wipe (the darwin log still shows `First-run setup marker written` on the 'pristine' boot).

In `.github/workflows/build-bun.yml`, step `Smoke test first-run invitation (fresh boot)`, REPLACE the line:

```
          rm -rf "$RUNNER_TEMP/inv/save"   # the copy must be DB-pristine too: the main smoke's library would trip the setupComplete backfill
```

with:

```
          rm -rf "$RUNNER_TEMP/inv/save" "$RUNNER_TEMP/inv/mStream.app/Contents/MacOS/save"   # DB-pristine on BOTH layouts: darwin's dataRoot lives inside the .app
```

Suggested message: `ci(smoke): wipe darwin's in-app save/ too — its dataRoot lives inside the .app`. Note your pushes currently create no workflow runs (push-event delivery is broken for your credential or a GitHub hiccup) — just push; the Mac session dispatches the run manually and reports.

IrosTheBeggar:
Done — `7149dfb8`: the wipe now clears both layouts (`inv/save` and `inv/mStream.app/Contents/MacOS/save`), suggested message verbatim. Per your note, no dispatch from this side — over to the Mac session to run and report the darwin leg.

PR #928, opened 2026-08-29
feat(discovery): fetch byte ceiling + model-compat gate for auto-fetch

IrosTheBeggar:
Two hardenings from the 2026-08-28 live discovery-network audit.

## 1. Arbitrary-volume download (the size lie)

`fetchPeer`'s storage-cap and disk-free-floor checks ran on the peer-**announced** size, but nothing bounded the transfer itself: a malicious catalog peer could announce `size: 20480` and serve a multi-GB blob — an unbounded write onto the volume (prod has already crash-looped on a full disk once).

- `discoveryP2p.fetch()` now carries `maxBytes` = announced size (byte-exact for honest peers — it's the publisher's own stat), or the storage-cap headroom for size-less announcements. Sidecars **≥ v1.0.4** abort the transfer at the ceiling (companion PR: IrosTheBeggar/mstream-p2p-sidecar#, see below); v1.0.3 tolerates and ignores the extra key.
- `fetchPeer` re-runs the admission checks on the **actual bytes on disk** after download: over-announced ⇒ reject, delete, forget the blob, normal failure backoff. Same-hash refreshes are exempt from the delete (the file IS the held copy). This on-disk re-check is the load-bearing bound until the sidecar pin bumps to v1.0.4.
- The raw admin catalog-fetch route (`POST /p2p/fetch` by endpointId) gets the same announced-size ceiling; its ticket mode stays uncapped (manual hand-off from someone the admin already trusts, no announcement to bound it).

## 2. Block auto-fetch of model-incompatible snapshots (default)

The similarity search reads only `WHERE model_id = ours`, so a snapshot announced under another model — or none at all (a server that never embedded, e.g. the empty 'Cypher' DB prod auto-fetched) — is unsearchable dead weight. `candidateOrder` only *deprioritized* them; reconcile still fetched them whenever the shelf had room.

- New pure predicate `modelCompatible` gates reconcile's top-up **and** rotation's candidate pool once a local model exists. No local model ⇒ no signal ⇒ everyone passes (unchanged).
- Escape hatches: `discoveryP2p.autoFetchIncompatibleModels` (config, default **false**), the manual admin fetch (ungated), and stale-refreshes of snapshots already on the shelf (ungated — what you hold keeps refreshing until removed).

## Tests

- Unit: model-gate + `modelCompatible` cases; full unit dir 607/607.
- New integration suite **“announced-size lie”**: honest byte-exact announce fetches fine; an under-announced blob is refused (500), the held seq-1 copy survives untouched, and the liar's bytes are not left in `discovery-peers/`. Written to pass under both v1.0.3 semantics (post-check) and v1.0.4 (transfer abort).
- Full discovery integration suite 50/50, run locally against a patched v1.0.4 dev sidecar build.

## Not in this PR

- The sidecar-side transfer abort ships separately (mstream-p2p-sidecar `feat/fetch-max-bytes` → v1.0.4); the manifest pin bump follows the usual release ritual after that publishes.
- Prod's existing empty Cypher shelf entry predates the gate — remove it via the admin shelf UI if unwanted (the gate stops *new* auto-fetches, it doesn't evict).

### DIFF
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -1076,7 +1076,10 @@ export function setup(mstream) {
 
     let addressing;
     let label;
+    let maxBytes;
     if (req.body.ticket) {
+      // Ticket = a manual hand-off from someone the admin already trusts;
+      // no announced size exists to bound it, so it stays uncapped.
       addressing = { ticket: req.body.ticket };
       label = `ticket ${req.body.ticket.slice(0, 32)}…`;
     } else {
@@ -1086,11 +1089,15 @@ export function setup(mstream) {
       }
       addressing = { hash: entry.payload.hash, provider: entry.from };
       label = `peer ${req.body.endpointId.slice(0, 12)}…`;
+      // The catalog flow rides the peer's own announcement, so hold the
+      // transfer to the size it claimed — the same ceiling fetchPeer uses
+      // (sidecars ≥ v1.0.4 abort past it; older ones ignore the param).
+      maxBytes = entry.payload.size > 0 ? entry.payload.size : undefined;
     }
 
     const outDir = path.join(config.program.storage.dbDirectory, 'discovery-peers');
     try {
-      res.json(await discoveryP2p.fetch(addressing, outDir));
+      res.json(await discoveryP2p.fetch(addressing, outDir, maxBytes));
     } catch (err) {
       winston.warn(`discovery P2P fetch failed for admin '${req.user?.username}' (${label}): ${err.message}`);
       throw new WebError(`fetch failed: ${err.message}`, 500);
--- a/src/api/admin.js
+++ b/src/api/admin.js
@@ -1097,7 +1097,11 @@ export function setup(mstream) {
 
     const outDir = path.join(config.program.storage.dbDirectory, 'discovery-peers');
     try {
-      res.json(await discoveryP2p.fetch(addressing, outDir, maxBytes));
+      // No hold: this raw route delivers a file, not a shelf membership —
+      // nothing here would ever forget() the blob, so rooting it in the
+      // sidecar store would pin those bytes forever. Un-held store copies
+      // age out with GC; the exported file is the deliverable.
+      res.json(await discoveryP2p.fetch(addressing, outDir, { maxBytes }));
     } catch (err) {
       winston.warn(`discovery P2P fetch failed for admin '${req.user?.username}' (${label}): ${err.message}`);
       throw new WebError(`fetch failed: ${err.message}`, 500);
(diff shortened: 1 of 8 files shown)


IrosTheBeggar:
## Overnight follow-up: holds/GC fix + full smoke run

**New commit `d9b06e8f`** — while smoking the fetch changes I confirmed a third latent bug flagged during the audit: **fetched blobs had no GC root in the sidecar store** (mark phase roots only tags; downloads create none; live-set rebuilt every cycle — verified against iroh-blobs 0.103.0 source and pinned by a new `tests/gc_semantics.rs` in the sidecar repo). Every held snapshot evaporated from the store within 15 min while the holds beacon kept advertising it — swarm fetches from us dialed in and failed. Fix pair:
- sidecar `fa9a124` (same PR 2): `hold` fetch param → persistent `held-<hash>` tag, plus a temp tag across the transfer so a sweep can't race a slow download; `forget` reclaims by hash as before.
- mStream: `fetch()` takes `{ maxBytes, hold }`; fetchPeer sends `hold:true`; invalid-snapshot rejects now `forget()` too; and **`reseedShelf()`** re-imports every shelf file at stack start + beacons the hold-set — heals pre-fix installs (works on v1.0.3 too, publish always tagged) and un-silences the beacon on servers with a shelf but no own snapshot.

## Verification matrix

| What | Result |
|---|---|
| Unit suite | 607/607 (two runs) |
| Discovery integration, patched **v1.0.4** dev sidecar | **52/52** |
| Discovery integration, pinned **v1.0.3** binary (CI semantics — post-check path) | **52/52** |
| Sidecar `cargo test` (16 unit + gc_semantics + stdin-EOF) | green |
| ESLint on all touched files | clean |

New integration suites: **announced-size lie** (honest exact-size fetch lands; under-announced blob refused, held copy survives, no liar bytes on disk), **boot reseed serves held snapshots** (origin gone; a third party fetches the held snapshot *from the server* after a cold boot and sees it in the holds beacon), **auto-fetch model gate** (one reconcile pass: matching model fetched, alien skipped, manual fetch bypasses).

## Standalone smokes (patched sidecar, loopback only)

- **300 MB blob announced as 20 KB** (the DoS scenario): single-origin branch aborted in **9 ms at 32 KB** transferred; swarm branch in **375 ms at 64 KB**; victim store growth ≈ 0. Uncapped control: full 300 MB in 18.2 s, byte-verified. 6/6 checks.
- **Param compat**: pinned v1.0.3 **ignores** `maxBytes`/`hold` and fetches fine (documented degrade — the JS post-check is the bound there, proven by the 52/52 v1.0.3 leg); patched v1.0.4 **enforces** (`transfer exceeded maxBytes (16384 > 10) — aborted`). 3/3 both legs.

One earlier v1.0.3 suite run had a single `zero-touch auto-publish` failure — server boot timed out at 90 s while I had the unit suite running concurrently; a clean re-run passed 52/52, so it was CPU-contention flake, not a regression.

Issue #128, opened 2018-09-21
request: show filename if id3tag is empty


aroundmyroom:
Title: undefined
Artist: undefined
Album: undefined
if possible: when Id3tag is empty the info is stored this way in the db, but if possible if title and artist is undefined show the filename in the play field
maybe with an * in front so that you know that mStream did actually notice it had no id3tag (show difference between indexed and non indexed files)

![image](https://user-images.githubusercontent.com/250108/45862744-6a78f200-bd73-11e8-9907-4f3f33b700d9.png)

IrosTheBeggar:
This is fixed now

aroundmyroom:
ok, that seems to me a rescan of de db is needed. Lets do that and test ;)

aroundmyroom:
db indexed, music playing as wanted .. no info in de id3tag field en the playbar gives me the name of the file

![image](https://user-images.githubusercontent.com/250108/45938564-634c2100-bfcb-11e8-9473-1d1f20012721.png)

great!

PR #929, opened 2026-08-29
feat(discovery): portable sonic seeds — read a vector out, take one in

IrosTheBeggar:
## Why

Auto-DJ's sonic mode seeds from **filepaths**, which name a row in *this* server's database. A client with several servers therefore can't ask them all for "something like what's playing" — the path means nothing anywhere else.

These two additions let the seed travel as what it actually is: a vector. Together they're what the app needs for a cross-server Auto DJ session.

## What's here

**`POST /api/v1/discovery/local/embeddings`** — the embedding for 1–8 paths, base64 float32 little-endian, with the model block. Batched because the rolling anchor seeds from the last few picks; one round trip per pick beats eight. Same failure semantics as its neighbours: uniform 404 for an unresolvable path, `notAnalyzed` for one the worker hasn't reached yet.

**`random-songs`: `similarToVector` + `similarToModelId`** — the same seed as a raw vector instead of paths. Nothing downstream of the sonic pool cared where a vector came from, so the change is confined to building it: the pool, the waterfall, the LRU cache and the reported cosine all work unmodified.

## Design notes

**A model-id mismatch is a hard 400**, deliberately unlike the federation peer route's soft `modelMismatch`. That route answers a *ranking*, where an empty result is truthful. This one answers "what do I play next", and comparing across model spaces returns confident nonsense — tracks with no real relationship to the seed. Callers are expected to read the model id first and skip a server that disagrees, so arriving here at all means a bug worth surfacing.

**The decode half of the federation route is now shared** as `decodeSeedVector`: dim, non-finite and zero-vector checks, the aligned copy a `Buffer`'s `byteOffset` makes necessary, and the defensive re-normalization without which a caller-averaged seed would inflate every similarity it produced. Two routes accepting foreign vectors through one front door beats a copy that drifts.

> [!IMPORTANT]
> **Note for callers.** The filepath form removes its seed hashes from the pool so the DJ never answers "what's next" with the song already playing. A vector carries no hashes, so that protection doesn't apply — `ignoreList` becomes load-bearing rather than merely helpful. Commented in both repos.

## Testing

**Two servers, disjoint libraries** (ports 3366/3377, synthetic tracks). A vector read from one and handed to the other returns *that server's own* tracks with a real cosine:

```
Change A: model=effnet-discogs v1 dim=1280 analyzed=true b64=6828 chars
Change B: pick "Alpha Track" | cosine 0.7974 | poolSize 3
          across 8 calls: Alpha×4, Gamma×3, Beta×1
Guard   : foreign model  -> 400 "Sonic seed is from model 'some-other-model', this library is indexed with 'effnet-discogs'"
Guard   : wrong length   -> 400 "embedding must be 1280 float32 values (little-endian), got 64 bytes"
Guard   : both forms     -> 400 conflict between exclusive peers
Regress : filepath seed still works -> "Gamma Track" | cosine 0.7929
```

**Unit tests** — 7 new, covering `decodeSeedVector`: well-formed decode, scaled input normalized to unit length, wrong length, non-finite, zero vector, and repeated decodes exercising the byte-alignment copy.

**Full suite** — `2509 tests, 2472 pass, 15 fail, 22 skipped`. Master baseline on the same machine is `2502 / 2461 / 15 / 22`: the same 15 failures, in `admin-mdns` (untracked WIP) and DLNA/discovery-seed timeouts. They're flaky rather than fixed — the failing names differ between runs, and `dlna-surface.test.mjs` passes 20/20 when run alone both with and without this branch.

### DIFF
--- a/src/api/discovery.js
+++ b/src/api/discovery.js
@@ -144,11 +144,87 @@ export function resolveVisible(uid, filter, canonHash, { withGenres = true } = {
   return row;
 }
 
+// Decode a client-supplied embedding into a unit vector this index can be
+// queried with. Extracted from the federation peer route so the two callers
+// that accept a foreign vector — federation/discovery/similar and the Auto-DJ
+// picker's vector seed — validate it identically; a divergence here would be
+// one route quietly accepting a vector the other rejects.
+//
+// Deliberately does NOT check modelId: the two callers differ on what a
+// mismatch means (federation answers a soft `modelMismatch`, the picker
+// refuses), so that policy stays with them.
+export function decodeSeedVector(index, base64) {
+  const raw = Buffer.from(base64, 'base64');
+  if (raw.length !== index.dim * 4) {
+    throw new WebError(`embedding must be ${index.dim} float32 values (little-endian), got ${raw.length} bytes`, 400);
+  }
+  // Aligned copy — a Buffer's byteOffset isn't guaranteed 4-byte aligned.
+  const ab = new ArrayBuffer(raw.length);
+  new Uint8Array(ab).set(raw);
+  const q = new Float32Array(ab);
+
+  let sumSq = 0;
+  for (let i = 0; i < q.length; i++) {
+    if (!Number.isFinite(q[i])) { throw new WebError('embedding contains non-finite values', 400); }
+    sumSq += q[i] * q[i];
+  }
+  const norm = Math.sqrt(sumSq);
+  if (norm === 0) { throw new WebError('embedding is a zero vector', 400); }
+  // The index vectors are L2-normalized at write time; "similarity" only
+  // means cosine if the query is normalized too. Callers should send unit
+  // vectors, but a scaled one costs nothing to fix here.
+  for (let i = 0; i < q.length; i++) { q[i] /= norm; }
+  return q;
+}
+
 function modelBlock(index) {
   return { id: index.modelId, version: index.modelVersion };
 }
 
 export function setup(mstream) {
+
+  // The embedding itself, for a client that needs to carry a seed somewhere
+  // this server can't reach — Auto-DJ spanning several servers averages the
+  // vectors of what is playing and hands the result to each of them, which
+  // only works if it can read them out in the first place.
+  //
+  // Batched because the rolling anchor seeds from the last few picks; one
+  // round trip per pick beats eight. Capped at 8 to match similarTo.
+  //
+  // Same failure semantics as the rest of this file: a uniform 404 for a path
+  // that can't be resolved, and notAnalyzed for one that resolves but has no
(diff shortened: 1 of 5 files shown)


IrosTheBeggar:
Review + rebase (2026-09-05). Rebased onto current master (36 commits behind, no conflicts — nothing on master touched these files), and pushed one extra commit with route-level tests; the PR had only pinned `decodeSeedVector` and exercised the two new behaviours by hand.

**The code is sound.** The extraction of `decodeSeedVector` is a straight lift with the model-id policy correctly left to each caller; the vector form of `random-songs` is confined to building the pool (cache key content-addressed on the base64, so the LRU behaves exactly as for the filepath form); the Joi pairing (`oxor` + the two `with`s + the custom guard) rejects every bad combination I could think of, and the tests now pin all of them. `--verify`'d locally: 7 unit + 28 sonic integration tests green on the rebased head (`test/integration/random-sonic.test.mjs` gained two describe blocks: reading a seed out, taking one in).

**One design point worth a second look — `ignoreList` is not the guard the description says it is.** The IMPORTANT note says that for a vector seed "`ignoreList` becomes load-bearing" for keeping the DJ off the song already playing. But `ignoreList` is a list of server-issued numeric track ids (not paths), and it is a *soft* cooldown: `finalisePick` falls back to the full candidate set as soon as the list covers every candidate. So with a vector seed and a small pool at high strictness — where the playing track is the most similar candidate of all — the DJ can and will re-serve the playing song. The filepath form never can, because the seeds' hashes are removed from `allowed` before the pool is built. Two ways to close it; neither blocks this PR:

- server: an optional `excludePaths` (≤ 8 vpaths on *this* server, resolved like seeds, unresolvable ones skipped silently, hashes dropped from `allowed`) would give the vector form the same hard guarantee — ~15 lines next to the seed-hash loop. Happy to add it if you want it in this PR.
- client: refuse an answer whose pick is what's playing / a recent anchor and take the next-best server. That is what I've done in the mobile PR (#128 there), which had a second problem here: it sent one shared `ignoreList` — server A's ids — to every server, and replaced it with the winner's; ids are per-server, so on any other server that list excluded unrelated tracks and the cooldown never followed the DJ. Now per-server.

**Version floor.** 6.25.0 shipped on 2026-08-30 without this PR, so the routes land in the next release (6.26.0 presumably). The mobile PR gated the feature on 6.25.0 — a 6.25.0 server would have been offered a session it silently sits out of; fixed there to 6.26.0.

Nits, no action needed: `Buffer.from(bytes)` in the embeddings route copies a view that `toString('base64')` would already have read from, harmless; the local discovery routes aren't in `docs/openapi.yaml`, but none of the existing ones are either, so consistent.

PR #969, opened 2026-09-09
feat(stats): Stats API v2 reads — summary, top, timeseries, history, tracks, periods

IrosTheBeggar:
## Summary

The read side of **Stats API v2**, on top of the write side that landed in #968 (S2 in the app repo's `PLAY_HISTORY_PLAN.md`). Six reads over the V70 listening log, every one scoped to the caller: their own plays, their visible libraries (`ignoreVPaths` narrows further), and **their timezone** — period and day boundaries follow the `tz` the client sends, never the server's clock, which is the wrapped-era bug class this replaces.

## The routes

| Route | Returns |
|---|---|
| `GET /api/v1/stats/summary` | the period's numbers as data: events, counted plays, unique tracks/artists/albums, listened time, skip and completion rates, discoveries (first counted play ever falls in the period), library coverage, sessions derived from 30-minute gaps, day streaks, the top day, peak hour and weekday, and the local/peer split |
| `GET /api/v1/stats/top?entity=tracks\|artists\|albums\|genres&metric=plays\|time` | ranked rows with share of the period; a track row carries the library's metadata object when it still resolves, otherwise the snapshot recorded at ingest (always for a federated peer's track, which also carries `peerId`); artists and albums merge by name across origins |
| `GET /api/v1/stats/timeseries?bucket=hour\|day\|week\|month\|hourOfDay\|weekday` | the per-user hourly rollup re-bucketed into the caller's zone |
| `GET /api/v1/stats/history` | the log newest first, cursor-paginated, filterable by track hash and origin |
| `POST /api/v1/stats/tracks` | per-track counters for a batch of paths or hashes (Song Info, badges) |
| `GET /api/v1/stats/periods` | the preset periods that overlap the log's bounds |

All take either `period=week|month|quarter|half|year|all` + `offset` or an explicit `from`/`to`, plus `tz`, `origin=all|local|peers`, `ignoreVPaths`.

## How

- `src/stats/queries.js` holds the SQL. Hashes resolve to tracks through two index-driven probes (`idx_tracks_audio_hash`, `idx_tracks_hash`) rather than a COALESCE join the planner can't seek. Time-series facts come from `user_hour_stats` (per-user totals, not narrowed by library or origin — a rollup has neither — and the part of the history that outlives retention); everything track-shaped comes from `play_events`.
- Period arithmetic is the `src/stats/time.js` half that #968 carried in advance: local midnights in the caller's IANA zone, Monday-first weeks, whole-period offsets, exact re-bucketing of UTC hours for any whole-hour zone (half-hour zones get half-hour error, documented).
- The shared route module `src/api/stats.js` gains a `requireAccount` guard on every route: a federation key or a guest token builds a user with no id, and stats are never read for one.
- OpenAPI: the six reads plus two shared schemas under the Stats tag.

## Verification

- `test/integration/stats-api.test.mjs`: boots a server, seeds tracks and plays through the store primitive, and checks the summary's counts and derived facts, top by both metrics with the federated snapshot path, timezone-correct day bucketing (a 22:30Z play moving to the next Berlin day), cursor pagination, per-track counters by path and hash, the periods listing, and the 400s. 9/9.
- The #968 ingest suite still passes on this branch (4/4); `eslint` clean; the OpenAPI file parses.
- The full review pass is in the app repo's `PLAY_HISTORY_PLAN.md` §3.4.

## Not in this PR

Still nothing advertises the surface in `features` — the `scrobble-by-filepath` shim, the retention sweep and the `stats` flag are the next change (S3), so clients never see a partial API.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -494,6 +494,21 @@ components:
             type: string
           description: Libraries (vpaths) to exclude from the query.
 
+    StatsPeriod:
+      type: object
+      description: The resolved range a stats read covered.
+      properties:
+        label: { type: string }
+        from: { type: string, format: date-time }
+        to: { type: string, format: date-time, description: "Exclusive." }
+        tz: { type: string }
+
+    StatsOriginSlice:
+      type: object
+      properties:
+        plays: { type: integer }
+        listenedMs: { type: integer }
+
     SearchArtistOrAlbumItem:
       type: object
       description: |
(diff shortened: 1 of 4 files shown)


IrosTheBeggar:
Added from the smoke round: `history?track=` now accepts the canonical `audio-hash`, the file `hash` (resolved through the library), or a peer's reported hash — a client holding only the file hash used to get an empty page. Covered in the integration test.

PR #978, opened 2026-09-11
stats: either-hash counters, peer snapshot enrichment, admin listening-history card, Stats under Navigation

IrosTheBeggar:
The optional items left after #977, plus one webapp change.

## Either-hash lookups on the per-track counters (`POST /api/v1/stats/tracks`)
A client holding the **file** hash of a track whose counters sit under the **audio** hash got nothing back. Every requested hash is now resolved through the tracks table in one batched query and looked up by its canonical key; each item is reported under the hash the client sent, with `canonicalHash` alongside when they differ. A hash the library does not know (a peer's key) is looked up as sent, so peer-keyed counters still resolve.

## Peer snapshot enrichment (`src/stats/enrich.js`)
A play of a peer's track stores whatever snapshot the app had at hand, often thin. After the ingest commit the server now asks the peer for the tracks' real metadata (`POST /api/v1/db/metadata/batch` on the peer, already on the federation allowlist, 4 s deadline, one call per peer per batch), completes the snapshot (the peer's strings win; duration and art fill in), and re-keys the play and its counters onto the audio hash through the shared hash migration, so the identical file turning up in this library merges the plays. A row is settled once (`enrichedAt`), even when the peer does not know the track; an unanswered peer leaves its rows for the daily pass, which runs on the retention sweep's cadence, bounded per run. It never fails an ingest. Tests reach it through `MSTREAM_TEST_FED_ENRICH_ENDPOINT`, the same pattern as the Last.fm fake.

## Admin retention UI
- New admin routes: `GET /api/v1/admin/stats` (the live `stats` config block, plays on file / oldest / accounts / federated rows still thin, the retention floor and last sweep, the enrichment counters), `POST …/retention`, `POST …/thresholds` (saved to the config file and applied live: the sweep and the counted rule read config per run), `POST …/sweep` (prune now), `POST …/enrich` (ask the peers now; per-pass numbers plus lifetime totals).
- Admin › Database gains a **Listening History** card with those facts, two edit modals (retention months; seconds / percent for when a play counts), and Prune now / Ask peers for metadata / Rebuild hourly totals buttons.

## Webapp
- The **Stats** sidenav entry moves from the Music section to the Navigation section, above Admin Panel.

## Verification
- New tests: `test/unit/stats-enrich.test.mjs` (real schema, fake peer: completion, re-key with counters following, unknown-to-peer settling, unanswered → backfill, enqueue filters, chunking), `test/integration/stats-enrichment.test.mjs` (fake peer over HTTP end to end incl. the 500 → admin pass recovery), `test/integration/stats-admin.test.mjs` (config file round-trips, live sweep and thresholds, validation, the admin gate), plus the either-hash case in `stats-api`.
- Full `npm test` run on the branch (result in the comments below), eslint and redocly clean.
- Browser smoke on a server booted from this branch with the seeded 120-day log (port 3161): the sidenav reads Navigation → Stats → Admin Panel → Log Out; the admin card shows 1,127 plays since May 13, the retention modal saved 12 months to disk and re-rendered the floor, Prune now and Ask peers both toasted their results, and the card surfaces the enrichment error honestly when federation is off ("federation endpoint is not running").

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -2681,7 +2681,7 @@ paths:
               description: At least one of the two lists.
               properties:
                 filePaths: { type: array, maxItems: 500, items: { type: string }, description: "Virtual paths (`<vpath>/<rel>`)." }
-                hashes: { type: array, maxItems: 500, items: { type: string }, description: "Canonical track hashes." }
+                hashes: { type: array, maxItems: 500, items: { type: string }, description: "Track hashes — the canonical (audio) hash or the file hash; either resolves to the same counters. A hash this library does not know is looked up as sent (a peer's key)." }
       responses:
         "200":
           description: One item per track that has counters; unknown tracks are omitted.
@@ -2695,7 +2695,8 @@ paths:
                     items:
                       type: object
                       properties:
-                        hash: { type: string }
+                        hash: { type: string, description: "The key as the client sent it." }
+                        canonicalHash: { type: string, description: "Present when the requested hash resolved to a different canonical key (a file hash sent where the counters sit under the audio hash)." }
                         filePath: { type: string, description: "Present for rows requested by path." }
                         plays: { type: integer }
                         skips: { type: integer }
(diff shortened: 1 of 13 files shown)


IrosTheBeggar:
Full `npm test` on the branch (commit 1426865f): 2450 tests, 2444 passed, 0 failed, 6 skipped. eslint and redocly clean.

IrosTheBeggar:
## Release smoke (2026-09-11)

**Real federation, not the fake.** Two full servers booted from this branch with federation on, paired over iroh (A minted a ticket, B added it as a peer). B ingested a thin play of A's track and completed it through the real dial: title/artist/duration from A, snapshot hash = A's audio hash, the row re-keyed and its counters moved (`stats/tracks` answers under the audio hash only), `admin/stats` shows enriched 1 / rekeyed 1 / thin 0. Ingest answered in 4 ms — never waiting on the peer. Then A was stopped: a second peer play was accepted in 6 ms, the row stayed thin, `lastError` = "peer-a: no answer within 4000ms", and the admin pass reported queued 1 / failed 1 after 4007 ms. Both rows render (the thin one from its snapshot, the completed one from A's metadata). B's config had no `stats` block: the admin readout showed the defaults.

**Either-hash on real library hashes** (branch server, seeded 120-day log): a track's file hash and audio hash return identical counters; the file-hash item carries `canonicalHash`; an unknown hash is omitted.

**Admin card in the browser:** retention modal → saved 12 months to disk, floor re-rendered; thresholds modal → "20 seconds, or 40%" saved and shown; Prune now → "0 plays pruned"; Ask peers → per-pass numbers; Rebuild hourly totals → "190 hours recomputed". No new console errors (the 400 on `admin/file-explorer/win-drives` and the 403s on `federation/peers` in public mode predate this PR).

**Sidenav:** Navigation → Stats → Admin Panel → Log Out; the Stats entry's reveal logic is unchanged and independent of the admin link's.

Local `npm test`: 2450 tests, 0 failures. Full-ci matrix: see the checks.

Issue #1004, opened 2026-09-19
Discovery: album and artist collection copies (job scope for federation-copy)


IrosTheBeggar:
## Context

The discovery plug-in work on `feat/discovery-plugins-2` ships **"Add to your collection" for a single song**: the `federation-copy` plug-in copies a paired peer's file into the caller's collection destination as a job (owned check by hash and tags, never overwrites, row inserted so it plays at once).

The design (cards 02, 03 and 07 in `docs/designs/discover-modal/`) also shows three album- and artist-level rows on the peer modal's Federation view. They are deliberately **hidden** in the webapp until the server can run them:

- **Add album to your collection**
- **Add artist to your collection** (every album, one job, album by album)
- **Add what you're missing** (diff the peer's discography against the local library by album identity, copy only the gaps)

## Why it is not a small change

A plug-in job today is "one plug-in acting on one recommendation":

- `discovery_plugin_jobs` (V74) has no options column, and the recommendation schema strips unknown keys, so a job cannot say "album" or "artist".
- Live jobs are unique per `(plugin, rec_key)`, so a song copy and an album copy of the same recommendation would dedupe against each other.

## What needs building

1. **A scope on the job.** Either a small migration adding an `options` JSON column, or a `scope` column (`song | album | artist | artist-missing`), accepted in the job create body and validated per plug-in. The live-job uniqueness has to include it (key suffix or a wider partial index).
2. **`federation-copy` run for the wider scopes.** List the peer's album songs / artist albums through the existing read proxy (`db/album-songs`, `db/artists-albums`), then copy song by song with the single-song path that exists: owned skip, never overwrite, per-song layout rendering (an album whose tags agree lands in one folder). Progress as "4 of 11 songs · 38 MB of 98 MB"; a cancel keeps what finished; the result lists copied, skipped and failed songs.
3. **"Add what you're missing".** Album identity match against the local library (album key / album artist), copying only albums the library lacks.
4. **Webapp.** Unhide the three rows (same job row vocabulary as the song row), and show the multi-song progress in the downloads tray.
5. **Peer-side policy (open question from the design review).** Copies are indistinguishable from playback on the peer; the pairing key's daily quota is the only cap. An artist copy can be hundreds of MB, so decide whether a per-key "allow copies" switch is wanted before this ships.

## Acceptance

- An album copy lands every song the library lacks under the caller's collection destination, skips owned ones, never overwrites, and survives a cancel with the finished songs kept.
[...]

PR #1001, opened 2026-09-17
feat(webapp): recommendation modal for Discover rows

IrosTheBeggar:
## Summary

Clicking a Discover recommendation now opens a modal instead of an inline row menu.

- **Network rows** ("From the network") open a window of plug-in sections: **Listen** (Deezer and iTunes previews, fetched only on a button press), **Federate** (admins only: compose a federation request to the anonymous peer, inline, with sent / paired states), **Find elsewhere** (canonical MusicBrainz chips plus search links).
- **Peer rows** ("From your peers") open on a **Federation** view built around the paired server: Play now / Queue next / Add to queue for the song; View / Play / Add to queue for the album; View artist; and **More like this on that peer**, which narrows the panel to one server under a removable chip. A **Plug-Ins** tab shows the same sections a network row gets.
- The header shows format from the peer's metadata route, online / last seen from the peers listing, album song count and length from the peer, and how many of the artist's albums you already have.
- **Hide rule:** a section exists only when the plug-in behind it exists and is configured; the Federate row disappears on a 403; nothing is ever shown locked. Phone widths render as a bottom sheet.
- The inline row menus and their "…" button are retired; peer rows keep the hover "+" for one-click queueing. Library rows are unchanged.

Server side: `POST /api/v1/discovery/federation/similar` takes an optional `peerId` to ask one paired server only (404 for an unknown or opted-out peer). The admin flag comes from the layered `/api/` endpoint; ping keeps its frozen flat contract (the server-info drift-lock test rejected an earlier attempt to put it there).

Also in this PR, as a separate commit: the design cards in `docs/designs/discover-modal/` (standalone HTML with `@dsCard` markers, plus the review board) and the `.design-sync/` generator that pushes them to Claude Design.

**Not built yet, pending API work** (listed on card 08): collection copies and their destination picker (design revision 2.2, card 10), the transfer allowance line, the accepts-requests hint, Auto DJ from a peer-owned seed.

## Test plan

- [x] `test/unit/discovery-plugins.test.mjs`, `test/integration/discovery-plugins-api.test.mjs`, `test/integration/api-server-info.test.mjs`, `test/integration/federation-discovery-aggregate.test.mjs` (new `peerId` case; `/api/` carries `user.admin`): 41 / 41
- [x] Full `npm test`: 2725 pass; the 1 failure was the drift-lock test before the admin flag moved off ping (green after), the 3 cancelled were `stats-forwarding` booting under smoke-rig load (3 / 3 standalone)
[...]

PR #1005, opened 2026-09-19
feat(discovery): plug-in jobs — collection copies, Get it (yt-dlp), Keep… and the webapp job UI (V74)

IrosTheBeggar:
## What this is

Discovery could tell you about a song you don't have; the only thing you could do with it was copy its title. This PR lets a user **act** on a recommendation: copy it from a paired server into their own library, download it, keep it, and watch all of that happen from the Discover panel.

It is the integration branch for the plug-in work since #997: the V74 job model, two acquire plug-ins, the retention and Keep… story for downloads, and the webapp UI on top.

## What's in it

**Schema V74 and the job runner** (`c551873d`)
- `discovery_plugin_jobs` (queued → running → done | failed | cancelled, one live job per plug-in and recommendation), a generic per-user `user_settings` store (secrets never read back), and `users.allow_discovery_jobs`.
- An in-process runner with a global cap and a per-plug-in cap, cancellation, and re-queue of interrupted work at boot.
- Jobs API: start, list, read, cancel. Owner-only, admins see all, foreign ids answer the same 404 as missing ones.

**Registry contracts**
- Per-user plug-in settings (`userSettings` with a Joi schema per key, `validateSetting`, `describeSettings`) behind generic `GET/PUT /api/v1/discovery/plugins/:name/settings`.
- Availability probes: a plug-in that cannot work (no yt-dlp, no ffmpeg) is left out of the user listing and answers 404; admins see it with the reason.

**Collection destination** (`src/discovery-plugins/destination.js`)
- One per account, shared by every acquire plug-in: a library the user may upload to, a base folder, and a layout rendered by the torrent path-template engine (now with an opt-in `{{PEER}}` variable; torrent templates still cannot use it). Default = the library's admin Path Template, else `{{ARTIST}}/{{ALBUM}}`.
- `GET/PUT /api/v1/discovery/collection/destination`.

**`federation-copy` plug-in — "Add to your collection"** (`3c25d980`)
- Copies a paired peer's file through the same stream proxy playback uses, with progress and cancel. Skips what the library already has (by hash, then by artist + album + title), never overwrites, inserts the row so it plays at once.

**`youtube` plug-in — "Get it"** (`c056a1ec`), **off by default**
- Searches with yt-dlp, scores uploads against the recommendation (shared match scorer, Topic-channel bonus, covers / karaoke / live dropped, minimum score 0.62), downloads the best one into a **Discover downloads** scratch library (created on first use, a subfolder per user), tags it from the recommendation.
- `src/util/yt-dlp.js` is a new helper; the existing Youtube DL route (`src/api/ytdl.js`) was rewritten on it with the same routes and behaviour. Process trees are killed on cancel (`taskkill /T` on Windows, the process group on POSIX).

**Keep… and retention** (`4d569519`)
- `POST /api/v1/discovery/plugin-jobs/:id/keep` moves a finished download into the collection destination (or a one-off one). Ratings and play counts follow by themselves (they key on the track hash); playlist entries on the old path are rewritten; a file in use answers 409.
- A retention pass removes downloads nobody kept after `discoveryJobs.downloads.retentionDays` (file mtime is the clock; 0 = never), marks their jobs, removes day-old partials and prunes old job rows. Runs 90 s after boot, then every 6 h; `POST /api/v1/admin/discovery-jobs/sweep` runs it on demand.

**What a client needs to draw jobs** (`54b6271e`)
- The plug-in listing carries `jobs.allowed` for the caller; `POST …/plugin-jobs/lookup` returns the caller's newest job per plug-in for a recommendation; `POST …/plugin-jobs/clear` drops settled rows (never live jobs or a download still waiting to be kept).
- Starting a job answers **409** when the live job already covering the recommendation belongs to another account. It used to hand that account's row back.

**Webapp** (`56127da2`, `214e37d0`, `c76ab241`)
- The recommendation window from #1001, plus: an "Add to your collection" row and a destination bar in the Federation view; the destination picker (library, base folder with Browse…, layout with variable chips and live validation, a preview of the final path); a **Get it** section whose rows are their jobs (queued → downloading → done with Play · Queue · Keep… → kept / failed with Retry / cancelled / expired); and a **downloads strip** under the Discover panel for jobs that outlive the window.
- `webapp/alpha/discover-jobs.js` is the pure half (row states, the layout preview, strip contents, queue patching). Its unit test holds the layout preview to the server's engine value for value, including error codes, so the two cannot drift.
- Nothing is shown locked: rows, the bar and Keep… are absent when the plug-in is not listed, the account may not start jobs, uploads are off, or there is no library to write to.
- Polling runs only while a job is live and the tab is visible, and backs off while the server does not answer.

**Also here:** the design cards for the window (`docs/designs/discover-modal/`, revision 2.2) and the design-sync notes; OpenAPI for every new route.

## Relation to #1001

This branch **contains #1001** (the recommendation window). GitHub would not retarget #1001 onto this branch because the branch already holds its commits, so it stays open against master. Merging #1001 first shrinks this diff; merging this one makes #1001 redundant.

## Defaults and gates

| Setting | Default | Note |
| --- | --- | --- |
| `discoveryPlugins.federation-copy.enabled` | `true` | Needs a paired peer, upload rights and a writable library; otherwise its row does not exist |
| `discoveryPlugins.youtube.enabled` | **`false`** | Needs yt-dlp and ffmpeg; listed only when the probe passes |
| `discoveryJobs.enabledFor` | `all` | `whitelist` + `users.allow_discovery_jobs`, the torrent pattern; flipped live from the admin API |
| `discoveryJobs.maxConcurrent` | 2 | Each plug-in also caps itself |
| `discoveryJobs.downloads.retentionDays` | 30 | 0 = never expire |

Copies and Keep… are uploads by another road: the server's `noUpload` and the account's `allow_upload` both apply. A download into the scratch library is not an upload and is gated by the jobs gate only.

## Testing

- New suites: `test/db/db-v74-discovery-jobs`, `test/db/discovery-plugin-jobs`, `test/unit/discovery-federation-copy`, `test/unit/discovery-youtube`, `test/unit/webapp-discover-jobs`, `test/integration/discovery-plugin-jobs`, `test/integration/discovery-federation-copy` (two real servers paired over iroh, copying into a per-run library), `test/integration/discovery-youtube` (a scripted yt-dlp stand-in: search, ranking, download, Keep…, sweep, cancel mid-download, the no-binary case).
- Run locally on the final commit: discovery unit + DB suites 68/68, plug-in integration suites 46/46, OpenAPI validates.
- The last full local run was before the final three commits: 2798 passed, 0 failed, 14 cancelled in `federation-e2e` from local load (a test rig was running beside it; that suite is 16/16 standalone). I have not repeated the full run since, so the `full-ci` matrix on this PR is the real check.
- The UI was driven end to end in a browser against a two-server rig: copy from a peer, download, Keep… (one-off and remembered), failure, Retry, Cancel, Clear finished, the hide rules (whitelist gate, uploads off), phone width, Esc layering. No console errors from the new code.
- Not tested: a real YouTube download in CI (by design), macOS and Linux process-tree kill outside CI.

## Where to look hard

- `src/discovery-plugins/downloads.js keepDownload` and `retention.js`: they move and delete files. Paths come from the jobs store and are re-resolved through `getVPathInfo`; the sweep only walks the scratch library.
- `src/util/yt-dlp.js startDownload`: the fallback that finds the output file after a non-zero exit uses a directory snapshot, so a failed download can never adopt an earlier file.
- The live-job unique index is server-wide on `(plugin, rec_key)`. For user-scoped plug-ins that means two accounts cannot fetch the same song at the same moment (the second gets the 409 above). Widening it is folded into #1004.

## Follow-ups

- #1004: album and artist collection copies, "Add what you're missing", and a per-user job scope.
- Remaining plug-ins (webhook, ListenBrainz, Lidarr, Soulseek via slskd), a Connected services page for per-user tokens, an admin card for the plug-in and jobs settings, translations for the new strings, user-facing docs.


IrosTheBeggar:
Superseded by #1008, which carries this branch together with the other two discovery plug-in PRs, current master, and the fixes from a smoke round. The write-up here still describes this part in detail; the branch is kept.

PR #1006, opened 2026-09-19
feat(admin): Discovery Plugins view — plug-ins, jobs, Discover downloads (with a size cap) and activity

IrosTheBeggar:
## What this is

A new view in the admin panel, **Discovery Plugins**, under Config directly below Discovery. #1005 gave users things to do with a Discover recommendation (copy it from a paired server, download it, keep it); until now an operator could only steer that from the config file. This is the panel for it, plus the one piece of server behaviour it surfaced as missing: a **size cap** on Discover downloads.

**Stacked on #1005** (base `feat/discovery-plugins-2`), so the diff shows only this work. Merge #1005 first; GitHub retargets this to master when that branch goes.

Design cards: `docs/designs/discovery-plugins-admin/` (nine cards and a review board). Decided before building: under Discovery · one card with four tabs · no "what leaves the server" line on the rows · plug-in defaults unchanged (previews on, YouTube off) · a size cap on the downloads folder.

## The view

One card with the Discovery view's boxed status header ("5 of 6 plug-ins on, 1 cannot run · jobs open to all users · 2 jobs live") and four tabs.

- **Plug-ins** — every registered plug-in as a row with a live switch, grouped by the section of the recommendation window it feeds, in the window's order (Listen, Get it, Send to, Find elsewhere). A plug-in that is on but cannot run says why in its probe's own words ("yt-dlp not found (yt-dlp)"), with `[settings]` and `[check again]`; users never see such a row. `[settings]` opens one generic modal over the keys the server declares editable: YouTube's binary (with **Test**, a dry run of the probe against the values in the form), format, per-file size limit and results scored; the iTunes storefront.
- **Jobs** — who may start jobs: all users or a whitelist, applied on click, with each account's upload right beside its tick (copies and Keep… need it), plus the runner's two limits. A server with no users gets the reason Whitelist is unavailable rather than a dead control.
- **Downloads** — what is waiting in Discover downloads, how big against the cap, how old, per user; the retention clock and the cap; the full state; Sweep now with its result.
- **Activity** — every account's jobs, newest first, in the player's own state words (the admin page loads `webapp/alpha/discover-jobs.js`), with cancel and the reason behind a failure: the error verbatim, plus one plain sentence for the causes we recognise.

Everything applies at once and toasts; nothing needs a restart, so there is no APPLY button. The view polls every 3 s only while a job is live and only while it is on screen. Tables stack into labelled rows where they do not fit (Activity below 1400 px, the others on phones).

## Server

- `GET /api/v1/admin/discovery-plugins/status` — one answer for the view: every registered plug-in (on or off, `available` / `reason` / probe `detail`, its editable keys and their values, and `connectedUsers` for one that keeps a per-user secret), the job gate and load, and the downloads numbers.
- `POST /api/v1/admin/config/discovery-plugins` takes `settings` beside `enabled`: only the keys a plug-in declares as `adminSettings`, checked with the schema the config file is held to, applied live, re-probed. A refusal reads `"<field>: <why>"` so the modal can put it under the right input.
- `POST /api/v1/admin/discovery-plugins/probe` — "check again" replaces the cached answer; with `settings` it is a dry run that saves nothing.
- **The youtube probe now runs yt-dlp** (`--version`). A path that exists but cannot be executed used to pass the probe and then fail every job with `spawn EFTYPE`; it is now reported on the plug-in's row. The probe's ffmpeg wait timer is unref'd (it held test processes open for 15 s on machines with yt-dlp installed).
- **Size cap:** `discoveryJobs.downloads.maxSizeMb`, 5120 by default, 0 = none. Retention is a clock, not a quota, and a busy server could fill a disk inside it. At the cap a new download fails with `Discover downloads is full (5.1 GB of 5.0 GB). Keep or remove some downloads, or ask the admin to raise the limit.` Checked before a download starts, so it can be overshot by one file (`youtube.maxFilesizeMb` bounds that); half-written files count. Set through `downloadsMaxSizeMb` on the jobs setter, beside the new `downloadsRetentionDays`.
- The downloads numbers are read from the folder (size and mtime, the retention pass's own clock: downloaded tracks store no size). The pass and the usage summary share one walk and one partial-file rule, and the pass remembers its last run for the "next pass" tile.
- `GET /api/v1/admin/users` carries `allowDiscoveryJobs`; the admin's all-accounts jobs list names each owner (null for a deleted account, and for the shared anonymous account of a server with no users).
- OpenAPI now documents all six admin plug-in routes. The switch, the jobs setter and the per-user flag never were.

## Also in this PR

- `.design-sync/build-cards.mjs` bundles several card sets (`cardSets`). An upload reconciles deletes, so without this the admin cards would have replaced the Discover-window cards in the Claude Design project. The admin set prefixes its tokens and classes (`--adm-*`, `adm-*`) so both stylesheets can share one bundle.
- The two acquire plug-ins' descriptions are rewritten for an operator (they are what the row shows).

## Testing

- New: `test/integration/discovery-plugins-admin.test.mjs` (11 tests: status, connected users, settings and their refusals, probe, downloads numbers, the size cap refusing and releasing a download, the users flag and owner names; a second server **without** the yt-dlp test hook for the missing-binary and exists-but-cannot-run cases, since the hook outranks the configured binary) and `test/unit/discovery-plugins-admin.test.mjs` (6).
- Run locally on the final commit: discovery unit + DB suites 74/74, plug-in integration suites 48/48, OpenAPI validates. I have not run the full suite locally; the `full-ci` matrix is the check for that.
- The view was driven in a browser against the two-server rig: switching a plug-in off and on (and watching it leave the user listing), the settings modal with Test and Save, the whitelist table, the downloads tiles with a real download, the full state with a filler file and a 1 MB cap, a refused download and its "why", a live job with progress and cancel, desktop (1500 px), mid (850 px, stacked Activity) and phone (375 px) widths with no sideways scroll.
- Not covered: the view has no automated test (the admin webapp has none); strings exist in English only.

## To decide before merging: an executable path in the admin API

The YouTube plug-in's `binary` setting, and the Test button's dry run (which runs the path it is given with `--version` and saves nothing), are the first place the admin API can name an executable for the server to run. Until this PR every executable path (`transcode.ffmpegDirectory`, and this same `binary`) could only be set in the config file.

It stays inside the trust the admin API already carries: an admin can add any folder on the host as a library and upload into it, and choose any backup destination. It is still the most direct primitive so far, and on a server with no users every caller is the admin unless `lockAdmin` is set.

The design card draws the editable field with Test, so that is what is built. Two alternatives, each a small change:

- **Read-only in the panel.** Show the configured path and the probe's answer, keep `[check again]`, leave the path to the config file: drop `binary` from the plug-in's `adminSettings` and the dry run has nothing executable left to try.
- **A fixed drop folder.** Look for yt-dlp in a server-owned folder as well as on PATH (the way `bin/ffmpeg` works) and have the row name that folder, so choosing what runs needs access to the host's disk rather than the admin API.

## Still assumed

Every control applies at once with a toast, with no APPLY button on the access radios, unlike Torrent's "Enabled For". That was one of the six design calls and the only one not answered; it is a small change either way.


IrosTheBeggar:
Superseded by #1008, which carries this branch together with the other two discovery plug-in PRs, current master, and the fixes from a smoke round. The write-up here still describes this part in detail; the branch is kept.

